PR #3, opened 2023-10-25
feat: create nextjs app and template home page

### LINKED ISSUE #2: Setup turborepo
For initial release, this application will contain a NextJS App Router frontend. We'll include Sanity and any other apps we need later.

Note that due to the issues we've been facing with React Bootstrap in terms of SSR, we'll be switching to [Tailwind](https://tailwindcss.com/).

samderanova:
This PR creates a Tailwind NextJS frontend with a template home page, some Radix UI components, and GitHub deployment workflows.

Resolves #2


tyleryy:
Forgive me if I am out of line or wrong, but I think we should talk about what stack we should use before we set up the NextJS project. Specifically, should we use TailwindCSS or SCSS? I believe you select which one to use at the beginning of the next js project creation, so it would be useful to make a hard decision beforehand.

samderanova:
Yeah this is part of the reason why I created this PR instead of directly pushing it to main.

alexanderl19 (approved):

PR #5, opened 2023-11-05
Setup basic landing page

samderanova:
- Customized PostCSS configuration to allow syntax similar to Sass (see `Button.module.css`)
- Created `favicon.ico`
- Imported necessary font and font colors
- Created custom Button component
    - While the hover does work, I could not get the transition to work properly. The background color for it is actually a background image as it uses the `radial-gradient` property, so I used `transition: background-image 1s ease`, but this does not seem to have any effect.
- Added basic landing page with a button that links to the form to subscribe to our mailing list


samderanova:
The Vercel settings for the project were configured incorrectly. The deploy preview can be found [here](https://irvinehacks-site-2024-alfuay2jy-hackatuci.vercel.app/).

apps/site/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,31 @@
+import styles from "./Button.module.css";
+
+interface ButtonProps {
+	text: string;
+	href?: string;
+	isLink: boolean;
+}
+
+const Button: React.FC<ButtonProps> = ({ text, href, isLink }) => {
+	if (isLink) {
```
taesungh: Suggestion: isn't `href` self-sufficient to describe `isLink`?

apps/site/src/components/Button/Button.tsx
```diff
@@ -0,0 +1,31 @@
...
+				rel="noopener noreferrer"
+				className={styles.button + " font-body"}
+			>
+				{text}
+			</a>
+		);
+	}
+	return (
+		<input
+			type="submit"
```
taesungh: Note: think using an actual `button` element would be more appropriate, we can revisit this.

apps/site/src/app/globals.css
```diff
@@ -1,27 +1,30 @@
...
-  color: rgb(var(--foreground-rgb));
-  background: linear-gradient(
-      to bottom,
-      transparent,
-      rgb(var(--background-end-rgb))
-    )
-    rgb(var(--background-start-rgb));
+h1 {
+	font-size: calc(30px + 2vw);
+	font-weight: bold;
```
taesungh: Suggestion: Since our typeface has multiple intermediate weights, it might be better to use the actual weight number, i.e. 700 for bold.

apps/site/src/app/globals.css
```diff
@@ -1,27 +1,30 @@
...
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
 
 :root {
-  --foreground-rgb: 0, 0, 0;
-  --background-start-rgb: 214, 219, 220;
-  --background-end-rgb: 255, 255, 255;
+	--color-cream: #fffce2;
+	--color-brown: #432810;
```
taesungh, line 10: Note: not too familiar with Tailwind, but we can look into more native was to specify color themes.

apps/site/src/app/(home)/sections/Landing/Landing.tsx
```diff
@@ -0,0 +1,19 @@
+import Button from "@/components/Button/Button";
+import styles from "./Landing.module.css";
+
+const Landing = () => {
+	return (
+		<section className={styles.landingSection}>
+			<div className={styles.landingGroup}>
+				<h1 className="font-display">IrvineHacks 2024</h1>
+				<p className="font-display">January 26-28</p>
```
taesungh, line 9: Nitpick: although we don't have an official style guide, we tend to prefer using n-dashes (e.g. 26–28, `&ndash;`) for ranges.

taesungh (commented):
Looking alright from preview, small notes.

apps/site/src/assets/backgrounds/landing-background.png
taesungh, line 1: Note: this graphic may be more appropriate as a compressed JPEG, especially since it's used as a background image (no next/image optimization).

samderanova: Figma doesn't support exporting as JPEG.

taesungh (approved):
Works for now, thanks for the quick work under crunch.

PR #8, opened 2023-11-07
Feature/landing page fireflies

tyleryy:
### Overview

Added Fireflies using Three.js/React Three Fiber.

https://drei.pmnd.rs/?path=/docs/staging-sparkles--docs
https://github.com/pmndrs/drei
https://docs.pmnd.rs/react-three-fiber/getting-started/introduction
https://github.com/pmndrs/react-three-next


samderanova:
We should work on getting necessary content on the site first, so I won't be addressing this PR until afterwards. Also, please create an issue before opening up a PR so that we know what you're working on.

tyleryy:
I agree we should work on the content first, and I plan to do so. I just wanted to do this really quick for the landing page. I imagine that our landing page will be up for awhile, so I wanted it to look good for any passing by hackers. 

# Old

Screen Shot 2023-11-06 at 7 34 23 PM

# New

https://github.com/HackAtUCI/irvinehacks-site-2024/assets/66880934/c5e8cca7-d6fd-4d32-ae6e-ee1700267484

samderanova:
Ah that's true, I see your point. I'll take a look when I can then 👍

samderanova (approved):
Looks good to me, thank you for working on this. I think we should be able to use React Three instead of Framer for the fog though but we can discuss that in a separate issue.

samderanova (changes requested):
It turns out that this takes up quite a bit of memory. There's a native approach in Sass that we can try implementing first. See [Blinkers.scss](https://github.com/HackAtUCI/ZotHacks-Site-2022/blob/bdb412f54a8edf2873c05088c58c0792af112866/src/app/components/blinkers/Blinkers.scss) from an earlier version of ZotHacks.

tyleryy:
I did a Lighthouse scan on the site. I believe the performance is actually ok. I'm a bit new to web optimization and performance, so I may be wrong. Also, on my own machine, it took about 800 MB of memory, which is not awful.

Screen Shot 2023-11-10 at 3 08 57 PM
Screen Shot 2023-11-10 at 3 09 18 PM
Screen Shot 2023-11-10 at 3 09 41 PM
Screen Shot 2023-11-10 at 3 10 24 PM

apps/site/src/app/(home)/sections/Landing/Landing.tsx
```diff
@@ -16,4 +39,68 @@ const Landing = () => {
 	);
 };
 
+const Fireflies = () => {
```
samderanova, line 42: Nitpick: since this takes up a lot of lines, we should extract this to its own component.

apps/site/src/components/canvas/Examples.tsx
```diff
@@ -0,0 +1,105 @@
+"use client";
```
samderanova, line 1: Do we use any of these examples? If not, we should remove them.

apps/site/src/components/dom/Layout.tsx
```diff
@@ -0,0 +1,40 @@
+"use client";
+
+import { ReactNode, useRef } from "react";
+import dynamic from "next/dynamic";
+const Scene = dynamic(() => import("@/components/canvas/Scene"), {
+	ssr: false,
+});
```
samderanova, line 7: Suggestion: I don't think you need to import `Scene` dynamically (and disable SSR) here because both are client components. I tried importing it normally and it still works.

apps/site/src/components/dom/Layout.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const Layout = ({ children }: { children: ReactNode }) => {
+	const ref = useRef<any>();
```
samderanova, line 10: Nitpick: Since the ref is for a `div` you should say `useRef<HTMLDivElement>(null)` instead.

samderanova (changes requested):
Some small suggestions/nitpicks:

apps/site/src/components/dom/Layout.tsx
tyleryy: I made the change, but let's put a pin in this just in case. I am unsure about how Three.js / React Three Fiber interacts with SSR, which I believe this dynamic import prevents. Just a learning from using Framer Motion, which had odd behavior with SSR.

apps/site/src/components/canvas/View.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+// export const Common = ({ color }: any) => (
+// 	<Suspense fallback={null}>
+// 		{color && <color attach="background" args={[color]} />}
+// 		<ambientLight intensity={0.5} />
+// 		<pointLight position={[20, 30, 10]} intensity={1} />
+// 		<pointLight position={[-10, -10, -10]} color="blue" />
+// 		<PerspectiveCamera makeDefault fov={40} position={[0, 0, 6]} />
+// 	</Suspense>
+// );
```
samderanova, line 15: Question: is there a reason why this is commented out?

samderanova (commented):
One more small question:

apps/site/src/components/canvas/View.tsx
tyleryy: I left it there for reference. I'll take it out.

samderanova (approved):

PR #17, opened 2023-11-12
Set up about section

samderanova:
This creates the about section for the new IrvineHacks site. One issue is that unlike in the Figma, there is no clear transition between the landing and this new section.


apps/site/src/app/(home)/sections/About/About.tsx
```diff
@@ -0,0 +1,67 @@
...
+					<h2>
+						<span className={styles.statistic}>500+</span> Hackers
+					</h2>
+					<h2>
+						<span className={styles.statistic}>&#36;10,000</span> in
+						Prizes
+					</h2>
+				</div>
+				<h3 className="text-xl lg:text-2xl mb-5">
+					Create &#43; Connect &#43; Inspire
```
taesungh, line 40: Question: is there a particular reason the plus sign needs to be escaped?

apps/site/src/app/(home)/sections/About/About.tsx
```diff
@@ -0,0 +1,67 @@
...
+				/>
+				<div className="text-2xl lg:text-3xl mb-5">
+					<h2>
+						<span className={styles.statistic}>36</span> Hours
+					</h2>
+					<h2>
+						<span className={styles.statistic}>500+</span> Hackers
+					</h2>
+					<h2>
+						<span className={styles.statistic}>&#36;10,000</span> in
```
taesungh, line 35: Question: is there a particular reason the dollar sign needs to be escaped?

apps/site/src/app/(home)/sections/About/About.tsx
```diff
@@ -0,0 +1,67 @@
+import Image from "next/image";
+import Lantern from "@/lib/components/Lantern/Lantern";
+import koiJump from "@/assets/images/koi-jump.png";
+import styles from "./About.module.css";
+
+const About = () => {
+	return (
+		<>
+			<section className="container md mx-auto text-center my-14 w-3/4 sm:w-1/2 xl:w-2/5 relative">
```
taesungh, line 9: Thought: personally not a fan of this proportion based container sizing: the base 3/4 seems to not provide enough space on narrower views. There may be some bias with being used to the bucketed fixed widths we previously used with Bootstrap.

apps/site/src/app/(home)/sections/Landing/Landing.tsx
```diff
@@ -1,26 +1,35 @@
...
 import Fireflies from "../../components/Fireflies";
 
+import landingBackground from "@/assets/backgrounds/landing-background.jpg";
+
 const Landing = () => {
 	return (
-		<section className={styles.landingSection}>
+		<section
+			style={{ backgroundImage: `url(${landingBackground.src})` }}
+			className="bg-auto xl:bg-cover bg-no-repeat bg-center"
```
taesungh, line 16: Issue: the background image does not scale down enough on shorter views.

taesungh (commented):
Nice start, one issue with the background art and a suggestion about asset compression along with other smaller notes.

apps/site/src/app/(home)/sections/About/About.tsx
samderanova: I thought this and the other symbols had to be escaped, like how we did with the quotes. I'll unescape them.

apps/site/src/app/(home)/sections/About/About.tsx
samderanova: I can switch over to the fixed width classes then.

samderanova: While Tailwind has [classes](https://tailwindcss.com/docs/width) that are fixed width, I don't see any that are large enough for what we want. I did replace `w-3/4` with `w-96` though.

apps/site/src/app/(home)/sections/Landing/Landing.tsx
samderanova: Changed it so `bg-cover` is now used for all viewport sizes.

apps/site/src/app/(home)/sections/About/About.tsx
```diff
@@ -0,0 +1,67 @@
+import Image from "next/image";
+import Lantern from "@/lib/components/Lantern/Lantern";
+import koiJump from "@/assets/images/koi-jump.png";
+import styles from "./About.module.css";
+
+const About = () => {
+	return (
+		<>
+			<section className="container md mx-auto text-center my-14 w-96 sm:w-1/2 xl:w-2/5 relative">
```
taesungh, line 9: Issue: `w-96` is too wide for smaller phones (24rem > 320px). The Bootstrap approach I mentioned before relies on max-widths for each breakpoint.

taesungh (approved):
Thanks for adding.
Note for later: vertical padding between sections could be improved.

PR #20, opened 2023-11-13
Set up environment variables

### LINKED ISSUE #19: Set up dev and prod environment variables
As we gradually finish and merge different parts of our application, we'll want to hide them until we are ready for launch. This includes:
- **the rest of the home page:** one variable should control whether to show just the landing or everything on the home page
- **the application page:** another variable should control whether to show the application page or hide it

samderanova:
Resolves #19 .

I've only set up the one for the landing page since that's what we're working on right now, but I'll create a separate issue for the application page when we start working on it.

The `process.env.MAINTENANCE_MODE_HOME` will be added to the production only builds of Vercel, which means that you'll still be able to see everything on local development environments and preview builds, but not production.

### DIFF
--- a/apps/site/src/app/(home)/page.tsx
+++ b/apps/site/src/app/(home)/page.tsx
@@ -1,5 +1,7 @@
 import { Landing } from "./sections";
 
 export default function Home() {
-	return <Landing />;
+	// Show landing section only if still in maintenance,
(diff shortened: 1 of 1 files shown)


waalbert (approved):
LGTM

PR #33, opened 2023-11-24
Add minimal Python API deployable with Next/Vercel

taesungh:
- Create `apps/api` to hold API files which are copied to `apps/site` during Vercel build to be deployed as a Serverless Function
- Include FastAPI/Uvicorn handler and basic VS Code setup for Python
- Specify `api/index.py` as Vercel Serverless Function in `vercel.json`


apps/api/README.md
```diff
@@ -0,0 +1,19 @@
...
+python src/dev.py
+```
+
+which will start a Uvicorn server with auto-reload.
+
+## Deployment Configuration
+
+For deployment, the following environment variables need to be set:
+
+-   `PYTHONPATH=src/api` to properly import Python modules
```
samderanova, line 19: Question: I remember last year, you created a `.env` file with this key value pair in it. Is there a reason you're not doing that this time around?

apps/site/copy-api.sh
```diff
@@ -0,0 +1,4 @@
+# Copy API files frow apps/api
```
samderanova, line 1: Typo: "from"

samderanova (changes requested):
Thanks for getting this done! A few things:

apps/api/README.md
taesungh: Good catch, looks like I missed committing it because the name was in `.gitignore`.
The reason to have it locally is mainly for the VS Code Python extension.

samderanova (approved):
There is a conflict with between Next and Vercel where accessing `/api` results in an infinite redirect but other routes (i.e. `/api/docs`) seem to work. We can resolve this in a separate issue.

Issue #30, opened 2023-11-23
Restyle landing page


samderanova:
This issue is devoted to transitioning from the static background image for the landing section to the layered images from the Figma.

PR #37, opened 2023-11-25
fix: restore missing transparent section of background art

### LINKED ISSUE #36: Hotfix: Landing JPG vs PNG
Downloading the landing section background as a JPG removes the transparent background of the bottom, meaning that now the bottom of the background image is white instead of transparent. We should switch back to a PNG and increase the animation delay time to account for the longer time it will take to load it.

image

samderanova:
Resolves #36 by replacing the landing background JPG with its corresponding PNG version.


taesungh (commented):
Not a huge fan of the increased asset size. Could we explore a hybrid masking approach using a lossy-compressed background art with a separate alpha mask?

apps/site/src/assets/backgrounds/landing-background.jpg
samderanova, line 1: Question: Do you change the color of the bottom part in order for the mask to work?

samderanova (commented):
Thanks for adding the mask! One quick question:

apps/site/src/assets/backgrounds/landing-background.jpg
taesungh: No that was just a visual effect to make the water look like it's glowing from the lanterns.
However, I'm noticing a separate issue where when the aspect ratio of the view is wider than the graphic, the bottom (and top) are getting cropped to prevent a smooth transition.

taesungh: To elaborate, some parts of the water near the bottom are semitransparent, so a background color needed to be chosen anyway during the JPEG conversion. A pure white looked a little off together with the mask, so I went with the dark orange.

taesungh (approved):
Works for now. Let's not wallow in this too much since we'll eventually have the more fluid design from #30.

PR #38, opened 2023-11-26
fix: regenerate pnpm-lock.yaml

waalbert:
Due to merge conflicts in pnpm-lock.yaml, the deploy builds fail. It has been regenerated in an attempt to fix this.


samderanova (commented):
Thanks for the quick fix. Please do a *regular* merge.

samderanova (approved):

PR #41, opened 2023-11-28
Configure API to work on Vercel from Next SSR

taesungh:
## Changes
- Set up axios instance to use correct public Vercel URL for API calls during server-side rendering
  - Inject user's client-side cookies during API request
- Create demo example to illustrate server-side API call with axios and a native POST request from the client
- Update Uvicorn `host` to resolve IPv6 issues
- Configure Next rewrite which acts like a Vercel rewrite
  - This allows for local client-side testing of API endpoints and makes the API paths publicly accessible at `/api/`

## Testing
Clone the branch locally and visit `/demo` while also having the local API server running with Uvicorn
- Observe the dynamic response text when providing a cookie with the name `username`
- Try entering a number in the input field and submitting the form

Visit `/demo` in the deployed preview
- Observe the same behavior as in the development environment

### DIFF
--- a/apps/site/next.config.js
+++ b/apps/site/next.config.js
@@ -1,4 +1,25 @@
-/** @type {import('next').NextConfig} */
-const nextConfig = {}
+// All server-side API requests are handled by lib/utils/api.ts
+// but this rewrite still exists for locally testing native POST requests
+const LOCAL_API_URL = "http://localhost:8000";
 
-module.exports = nextConfig
+// When deployed on Vercel, this path acts like a rewrite in vercel.json
+// Next.js would normally be unable to find the API endpoint
+// but Vercel somehow steps in and makes the Serverless Function visible
+const VERCEL_API_PATH = "/api/";
+
+/** @type {import('next').NextConfig} */
+const nextConfig = {
+	rewrites: async () => {
+		return [
+			{
+				source: "/api/:path*",
+				destination:
+					process.env.NODE_ENV === "development"
+						? `${LOCAL_API_URL}/:path*`
+						: VERCEL_API_PATH,
+			},
+		];
+	},
+};
+
+module.exports = nextConfig;
--- a/apps/site/src/app/demo/page.js
+++ b/apps/site/src/app/demo/page.js
@@ -1,0 +1,1 @@
+export { default as default } from "./Demo";
--- a/apps/api/requirements.txt
+++ b/apps/api/requirements.txt
@@ -1,1 +1,2 @@
 fastapi==0.104.1
+python-multipart==0.0.5
--- a/apps/api/src/app.py
+++ b/apps/api/src/app.py
@@ -1,6 +1,10 @@
 from fastapi import FastAPI
 
+from routers import demo
+
 app = FastAPI()
+
+app.include_router(demo.router, prefix="/demo", tags=["demo"])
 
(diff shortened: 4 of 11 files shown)


apps/site/next.config.js
```diff
@@ -1,4 +1,25 @@
...
-const nextConfig = {}
+const nextConfig = {
+	rewrites: async () => {
+		return [
+			{
+				source: "/api/:path*",
+				destination:
+					process.env.NODE_ENV === "development"
+						? `${LOCAL_API_URL}/:path*`
+						: VERCEL_API_PATH,
```
samderanova, line 19: Question: Do we need to append `/:path*` to the end of `VERCEL_API_PATH` as well or is that something we'll save for another issue?

samderanova (commented):
Couple questions:

apps/site/next.config.js
taesungh: I don't fully understand what's going on under the hood, but it appears that Vercel treats the Next rewrite as a [Vercel rewrite](https://vercel.com/docs/edge-network/rewrites), so `/api/` references the Serverless Function defined in `api/index.py`. Including `/:path*` would actually break the rewrite and make other API subpaths inaccessible. The value in `next.config.js` could even be `/api/index.py` and Vercel somehow interprets that correctly.

The rewrite was previously expressed in `vercel.json` but removed since it appeared to be redundant, and a rewrite in `next.config.js` was needed anyway for proxying to localhost and could be included with comments unlike in `vercel.json`.

apps/site/src/app/demo/page.js
```diff
@@ -0,0 +1 @@
+export { default as default } from "./Demo";
```
samderanova, line 1: Nitpick: Shouldn't this be a `.ts` file?

samderanova: Suggestion [non-blocking]: It might be good to obscure this route in prod so it can only be accessible via previews.

samderanova (changes requested):

apps/site/src/app/demo/page.js
taesungh: Regarding file type, the file was just for a reexport and not any typed code, so I didn't bother, but I suppose it's up to preference.
Regarding hiding the route, good point, but I think we should deploy once to production to verify the behavior still works when the actual domain name (instead of *.vercel.app) is applied before hiding it.

samderanova (commented):
@waalbert was dealing with an Axios error related to IPv6 though (ECONNREFUSED ::1:8000) until he removed the `::1` entry from `/etc/hosts`. I'm pretty sure this isn't the proper way to resolve the error, but you did already change the Uvicorn host to `localhost` so I'm not too sure what the issue is here. To make matters even more peculiar, I'm not able to replicate the same error on my machine, perhaps due to an operating system/networking difference.

taesungh:
Interesting, on my machine, Node v16 resolves localhost to IPv4 whereas Node v17, v18, and v19 resolve IPv6 first, but interestingly Node v20 resolves IPv4 first. None of the versions seem to respect the order of the hosts file.

samderanova:
Ok that makes more sense. Currently, I'm on Node 20.

samderanova (approved):

PR #50, opened 2023-11-30
Setup/footer addition

Bl20052005:
added footer, moved and changed  .env file, changed tailwind to add /lib


.env
```diff
@@ -1 +0,0 @@
-PYTHONPATH=apps/api/src
```
samderanova, line 1: Question: Is there a reason why you moved the .env? This .env should remain where it is because it's important to the backend.

apps/site/src/lib/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+const Footer = () => {
+
+	const imageArray = [
+	[hack, "Hack Logo"],
+	[mail, "Email"],
+	[facebook, "Facebook"],
+	[instagram, "Instagram"],
+	[linkedin, "Linkedin"],
+	[youtube, "Youtube"]]
```
samderanova, line 19: Suggestion: It would be more readable if you used a list of hashmaps/objects here instead of nested arrays for enhanced readability. See [`Footer.tsx`](https://github.com/HackAtUCI/zothacks-site-2023/blob/main/apps/site/src/components/Footer/Footer.tsx) for an example. Additionally, we usually place constants like these outside of the component.

apps/site/src/lib/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+	return (
+	<footer className={`${styles.footer} flex flex-col items-center`}>
+		<div className='flex flex-row footer-logos items-center gap-10 mt-12'>
+			{imageArray.map((item, index) => {
+				return(
+					<a>
```
samderanova, line 26: Sorry, I realized that I didn't even provide you with the links to our social media icons. Here are the links:
- Hack icon: https://hack.ics.uci.edu
- Mail icon: mailto:hack@uci.edu
- Facebook: https://www.facebook.com/groups/HackAtUCI/
- Instagram: https://www.instagram.com/hackatuci/
- LinkedIn & YouTube: we actually decided not to include these, so please remove them.

samderanova: As such, please update the `<a>` tags accordingly by providing the links as values to the `href` attribute.

apps/site/src/lib/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,47 @@
...
+						/>
+					</a>
+				)
+
+			})}
+		</div>
+		<div className=" flex flex-row gap-1 items-center mt-2">
+			<h2>Made with</h2>
+			<svg className=" fill-white" xmlns="http://www.w3.org/2000/svg" height="16" width="16" viewBox="0 0 512 512"><path d="M47.6 300.4L228.3 469.1c7.5 7 17.4 10.9 27.7 10.9s20.2-3.9 27.7-10.9L464.4 300.4c30.4-28.3 47.6-68 47.6-109.5v-5.8c0-69.9-50.5-129.5-119.4-141C347 36.5 300.6 51.4 268 84L256 96 244 84c-32.6-32.6-79-47.5-124.6-39.9C50.5 55.6 0 115.2 0 185.1v5.8c0 41.5 17.2 81.2 47.6 109.5z"/></svg>
+			<h2>in Irvine, CA • Learn more about the <a><span style={{"color" : "#FBA80A"}}>Hack @ UCI Team</span></a></h2>
```
samderanova, line 40: Similarly, please update the `a` tag here to have an `href` of `"https://hack.ics.uci.edu"`.

apps/site/tailwind.config.ts
```diff
@@ -19,7 +20,7 @@ const config: Config = {
 				sans: ["Mulish"],
 			},
 		},
-	},
+	},
```
samderanova, line 23: Nitpick: please remove the extra space here.

apps/site/src/lib/components/Footer/Footer.tsx
samderanova: Additionally, the deployment fails because of each element should have a `key` prop. You can read more about it [here](https://react.dev/learn/rendering-lists).

samderanova (changes requested):
Great start on this so far! Please address my comments below.

Additionally, you'll want to install the Prettier extension on VSCode if you haven't already and ensure that it automatically formats on save. Please also make sure that you are using tabs to indent that are 4 spaces wide.

samderanova (changes requested):
A few more changes to address:

Bl20052005:
probably in a separate one then as I just did the merge

PR #60, opened 2023-12-04
Setup navbar

IanWearsHat:
- All links are temporary, clicking anteater logo goes back to home

- Expanded button style in `\lib\components\Button\Button.module.css` to include alternate login button style
  - Added optional prop for `\lib\components\Button\Button.tsx` to enable alt login style

- Scrolling adds a 50% opacity black background to navbar, but only for desktop
  - Mobile menu gets a background only when active

- Underline effect and spinning logo same as last year's site 🫡


Bl20052005:
I've noticed that on smaller screenwidths, the black bar appears and disappears suddenly and then the rest of the menu scrolls through, is this intentional or is this just kind of hard to fix?

tyleryy (approved):
Looks good! Really clean implementation. 

Some potential design changes: 
- Maybe try making the login button on mobile view bigger than the other navbar elements in the dropdown menu. 
- Maybe enable the black background for mobile. I feel the it flashing in when the dropdown menu is clicked looks slightly janky
- Maybe increase the glow effect for the text on hover

IanWearsHat:
> I've noticed that on smaller screenwidths, the black bar appears and disappears suddenly and then the rest of the menu scrolls through, is this intentional or is this just kind of hard to fix?

I took @tyleryy 's advice and just enabled the background by default instead of making it transparent.

IanWearsHat (commented):
Implemented @tyleryy 's design changes.

Bl20052005:
looks good to me now

apps/site/src/lib/components/Navbar/Navbar.module.scss
```diff
@@ -0,0 +1,58 @@
...
+	background-repeat: no-repeat;
+	background-position-y: 35px;
+	background-size: 0% $underline-size;
+	transition: 
+		filter 0.1s ease,
+		text-shadow 0.1s ease,
+		background-size 0.25s ease-in-out;
+
+	&:hover,
+	&.active {
```
taesungh, line 53: Issue: hover styles should also be applied to the `:focus` state
Issue: this currently targets the `.active` class, not the `:active` state

apps/site/src/lib/components/Navbar/Navbar.module.scss
```diff
@@ -0,0 +1,58 @@
...
+
+	// needed for underline effect to work
+	padding-bottom: 0.75rem;
+
+	background-image: linear-gradient(white, white);
+	background-repeat: no-repeat;
+	background-position-y: 35px;
+	background-size: 0% $underline-size;
+	transition: 
+		filter 0.1s ease,
```
taesungh, line 48: Issue: transitions without initial values can lead to unexpected behavior. For example, on Safari, there is an initial flash effect with the brightness filter unstable.

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,87 @@
...
+		const mobileNavInitStateHandler = () => {
+			setListShown(false);
+			setTransitionApplied(false);
+		};
+
+		// hides list on breakpoint
+		window
+			.matchMedia("(min-width: 768px)")
+			.addEventListener("change", mobileNavInitStateHandler);
+	}, []);
```
taesungh, line 31: Thought: this seems more complicated than it needs to be. There should be a way to hide the list with CSS.

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+	useEffect(() => {
+		const scrollHandler = () =>
+			window.scrollY !== 0 ? setHasScrolled(true) : setHasScrolled(false);
```
taesungh, line 35: Thought: using a binary scroll endpoint leads to a bit of a jarring effect when near the top of the page.
Would it be possible to smooth this out, perhaps with a transition or a dynamic opacity?

apps/site/src/lib/components/Button/Button.tsx
```diff
@@ -7,16 +7,17 @@ interface ButtonProps {
 	text: string;
 	className?: string;
 	href?: ComponentProps<typeof Link>["href"];
+  alt?: boolean;
```
samderanova, line 10: Suggestion: It might be better semantically to separate the two styles as "light" and "dark" versions, so we would replace `alt` with something like `useLightVersion` or something like that.

apps/site/src/lib/components/Navbar/Navbar.module.scss
```diff
@@ -0,0 +1,58 @@
...
+	font-size: 24px;
+	text-shadow: 0px 0px 20px #fffce2;
+
+	// needed for underline effect to work
+	padding-bottom: 0.75rem;
+
+	background-image: linear-gradient(white, white);
+	background-repeat: no-repeat;
+	background-position-y: 35px;
+	background-size: 0% $underline-size;
```
taesungh, line 46: Suggestion [non-blocking]: given the lower weight of the sans-serif font, it may be more better to use a thinner underline effect.

taesungh (dismissed):
Looking mostly alright. Some minor notes about styling and small issues with state usage and interactive accessibility.

apps/site/src/lib/components/HackLogo/HackLogo.tsx
```diff
@@ -0,0 +1,19 @@
+import hackAnteater from "@/assets/logos/hackuci-anteater.svg";
+import hackGear from "@/assets/logos/hackuci-gear.svg";
+import Image from "next/image";
+import styles from "./HackLogo.module.scss";
```
samderanova, line 4: Nitpick [repeating]: We usually order imports by (from top to bottom) first party libraries (ex. Next/React related imports), third party libraries, TS(X) file imports, images, and then SCSS files.

apps/site/src/lib/components/Navbar/Navbar.module.scss
```diff
@@ -0,0 +1,58 @@
...
+	@media (max-width: 768px) {
+		transform: translateY(-110%);
+	}
+}
+
+.showList {
+	@media (max-width: 768px) {
+		transform: translateY(0px);
+	}
+}
```
samderanova, line 32: Nitpick: For styles as simple as these, please see if it's possible to use Tailwind classes instead. See the Tailwind docs for [translating](https://tailwindcss.com/docs/translate), for example.

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,87 @@
...
+	useEffect(() => {
+		const mobileNavInitStateHandler = () => {
+			setListShown(false);
+			setTransitionApplied(false);
+		};
+
+		// hides list on breakpoint
+		window
+			.matchMedia("(min-width: 768px)")
+			.addEventListener("change", mobileNavInitStateHandler);
```
samderanova, line 30: Nitpick [repeating]: For [`useEffect`](https://react.dev/reference/react/useEffect), you'll want to return a function that removes the event listener when this component is unmounted.

samderanova: Suggestion: As Taesung noted, it might be easier to use some media queries and `display: hidden/block` instead of the event listener.

samderanova (changes requested):
Great start so far! A few things to note:

apps/site/src/lib/components/Navbar/Navbar.tsx
IanWearsHat: When the window gets smaller and hits the md breakpoint, the transition is applied and animates the list hiding, which isn't supposed to happen. I think that's why I added that event listener so that on the md breakpoint, the list doesn't animate and only does so when the button is pressed again. I don't exactly know how to fix this other than the event listener for a window size change.

apps/site/src/lib/components/Navbar/Navbar.module.scss
```diff
@@ -0,0 +1,37 @@
...
+		text-shadow 0.1s ease,
+		background-size 0.25s ease;
+
+	&:hover,
+	&:focus,
+	&:active {
+		text-shadow:
+			0px 0px 20px #ffa700,
+			0px 0px 20px #ffda7b;
+		filter: brightness(3.05);
```
taesungh, line 34: Issue: The filter is still causing some flashing glitches on webkit. Given this is being applied to just text and it's shadow, could we replace the filter effect with a brighter text color and/or text-shadow color?

apps/site/src/lib/components/Navbar/Navbar.module.scss
```diff
@@ -0,0 +1,37 @@
...
+		filter 0.1s ease,
+		text-shadow 0.1s ease,
+		background-size 0.25s ease;
+
+	&:hover,
+	&:focus,
+	&:active {
+		text-shadow:
+			0px 0px 20px #ffa700,
+			0px 0px 20px #ffda7b;
```
taesungh, line 33: Question: these two shadows have the same offset and size but neither are transparent colors, so only the first one appears to show up. What was the intent here?

apps/site/src/lib/components/HackLogo/HackLogo.module.scss
```diff
@@ -0,0 +1,28 @@
...
+
+	@keyframes spin {
+		from {
+			transform: rotate(0deg);
+		}
+		to {
+			transform: rotate(360deg);
+		}
+	}
+}
```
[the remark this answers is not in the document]
samderanova: I believe this is asking for a new line at the end of this file.

apps/site/src/lib/components/Navbar/Navbar.module.scss
IanWearsHat: The intention was to make the shadows brighter while keeping the size.

taesungh: Got it, I must have been thinking of something else, the colors do indeed mix.

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,78 @@
...
+						"transition-transform duration-500 ease-in-out md:transition-none md:translate-y-0 md:opacity-100 " +
+						"[&>*]:mb-5 [&>*]:md:mb-0 font-display gap-10 p-5 pt-3 bg-black bg-opacity-50 md:bg-opacity-0 md:p-0 md:flex md:items-center"
+					}
+					onTransitionEnd={() => setHidden(!listShown)}
+				>
+					<NavLinkItem href="/">Home</NavLinkItem>
+					<NavLinkItem href="/">Sponsor</NavLinkItem>
+					<NavLinkItem href="/">Schedule</NavLinkItem>
+					<NavLinkItem href="/">Resources</NavLinkItem>
+					<NavLinkItem href="/">Stage</NavLinkItem>
```
samderanova, line 70: Issue: We'll want to obscure most of these routes since they aren't ready yet. For now, please comment them out.

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,78 @@
...
+					<Button text="Login" href="/" isLightVersion />
```
samderanova, line 71: Issue: This one should direct to `/api/saml/login`.

samderanova (changes requested):
Looking good, a few more things to address before we merge.

apps/site/src/lib/components/Navbar/Navbar.tsx
taesungh: Not quite: we'll have our own login page at `/login`.

samderanova: Ahh that's right, sorry about that. Post finals brain.

Yes, as Taesung said, we'll be making our own login page to redirect UCI students to `/api/saml/login` and email magic links for non-UCI students.

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,82 @@
...
+					{/* <NavLinkItem href="/">Home</NavLinkItem>
+					<NavLinkItem href="/">Sponsor</NavLinkItem>
+					<NavLinkItem href="/">Schedule</NavLinkItem>
+					<NavLinkItem href="/">Resources</NavLinkItem>
+					<NavLinkItem href="/">Stage</NavLinkItem> */}
+					<Button
+						text="Login"
+						href="/api/saml/login"
+						isLightVersion
+					/>
```
samderanova, line 75: Issue: Sorry, I was wrong here. The `href` should be `/login` instead of `/api/saml/login`, as mentioned in [this comment](https://github.com/HackAtUCI/irvinehacks-site-2024/pull/60#discussion_r1427410721).

Separate issue: This link shouldn't open up in a new tab because we want to show our page at `/login` first (which is being implemented).

samderanova (changes requested):

apps/site/src/lib/components/Navbar/Navbar.module.scss
```diff
@@ -0,0 +1,35 @@
...
+
+	// needed for underline effect to work
+	padding-bottom: 0.5rem;
+
+	background-image: linear-gradient(white, white);
+	background-repeat: no-repeat;
+	background-position-y: 34px;
+	background-size: 0% $underline-size;
+	transition:
+		filter 0.1s ease,
```
taesungh, line 22: Nitpick: since we're not using the `filter` property any more, we don't need to include it in the transition.

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -0,0 +1,78 @@
...
+	}, []);
+
+	return (
+		<NavMenu.Root
+			className={`${
+				hasScrolled ? "md:bg-opacity-50" : ""
+			} transition-colors duration-0 md:duration-700 ease-out w-full z-10 flex flex-col fixed bg-black bg-opacity-0 md:flex-row md:items-center`}
+		>
+			<NavMenu.List
+				className={"bg-black bg-opacity-50 md:bg-opacity-0 flex p-3"}
```
samderanova, line 35: Nitpick: Since this is just a string, you don't need the curly brackets.

samderanova (changes requested):

samderanova (approved):
Thanks for all of your work on this. Please *squash* merge when you're ready.

taesungh (approved):

PR #61, opened 2023-12-06
fix: more fireflies

tyleryy:
# Overview

- added more fireflies particles
- moved the Perspective Camera up a little to have fireflies condense more on the landing section of the Home page (where the boats and Lunar New Year stands are)


alexanderl19 (approved):

samderanova (approved):
Slightly concerned about the amount of resource usage this increase might take up but looks good to me otherwise. Please *squash* when merging.

Issue #47, opened 2023-11-29
Add countdown timer to schedule page


samderanova:
Like with the ZotHacks website, we should set up a countdown to the hacking start time that should switch to another countdown to the hacking end time.

We can reference the code [here](https://www.hover.dev/components/countdown).

PR #62, opened 2023-12-06
feat: animated countdown and schedule page

tyleryy:
# Overview

Created a schedule page route and added animated countdown component. For Issue #47.

# Details

- used component libray hover.dev (see references)
- imported framer motion as dependency
- countdown component flashes 0's on page load due to initial state. Placed Loader component over until countdown state is changed to not 0's
- created placeholder div for where the schedule component will go

# Future Changes
- need to change date to IrvineHacks date
- may need hide route with Maintenance Mode env var
- Idea: use env var for the IrvineHacks date to precompute the Countdown Timer date (to fix hydration error from ZotHacks Site)

# References

[countdown](https://www.hover.dev/components/countdown)
[loader](https://www.hover.dev/components/loaders)
[framer motion docs](https://www.framer.com/motion/)


pnpm-lock.yaml
```diff
@@ -19,7 +19,7 @@ importers:
         version: link:packages/tsconfig
```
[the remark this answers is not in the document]
tyleryy: I don't actually know. I tried deleting and reinstalling the package lock, but I am not sure if that did anything.

tyleryy:
Added Fireworks to fit with Lunar New Year Theme:

references:
-  https://github.com/crashmax-dev/fireworks-js/ 
-  https://fireworks.js.org/

apps/site/src/app/schedule/components/ShiftingCountdown/ShiftingCountdown.tsx
```diff
@@ -0,0 +1,151 @@
...
+const SECOND = 1000;
+const MINUTE = SECOND * 60;
+const HOUR = MINUTE * 60;
+const DAY = HOUR * 24;
+
+const ShiftingCountdown = () => {
+	const intervalRef = useRef<ReturnType<typeof setInterval> | null>(null);
+	const [showTimer, setShowTimer] = useState(false);
+	const [showFireworks, setShowFireworks] = useState(false);
+	const [timerLabel, setTimerLabel] = useState("Hacking Starts");
```
taesungh, line 27: Suggestion: derived states make the logic more difficult to track. I think for all of these, we could compute the resultant values from the remaining time rather than maintaining separate states for them.

tyleryy: I agree separate states are difficult to track. Though wouldn't I need state changes to perform a re-render in psuedo-real-time? How do you suggest I implement it?

taesungh: A re-render will happen whenever any state is updated, and you have the countdown's `remaining` state which is constantly updated.

apps/site/src/app/schedule/components/ShiftingCountdown/ShiftingCountdown.tsx
```diff
@@ -0,0 +1,151 @@
...
+		days: 0,
+		hours: 0,
+		minutes: 0,
+		seconds: 0,
+	});
+
+	useEffect(() => {
+		intervalRef.current = setInterval(handleCountdown, 1000);
+
+		return () => clearInterval(intervalRef.current || undefined);
```
taesungh, line 39: Question/suggestion: is there a particular reason the component needs to keep track of the interval ID? If not, we can manage it only within the scope of the setup function.

tyleryy: Yea the interval can be managed only in the scope off the UseEffect callback, but I believe it is good practice to assign a reference to an interval id to persist past component re-renders. The component I believe will not re-render, so it technically is not necessary.

taesungh: Nothing else in the component needs to be aware of the interval ID, so there's no point in maintaining the ref. Aside: every state change will trigger a re-render, so I'm not sure what you mean by the second sentence.

tyleryy: kk ill remove it

apps/site/src/app/schedule/components/ShiftingCountdown/ShiftingCountdown.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { useEffect, useRef, useState } from "react";
+import BarLoader from "../BarLoader";
+import clsx from "clsx";
+import { Fireworks } from "@fireworks-js/react";
+
+import styles from "./ShiftingCountdown.module.scss";
+
+// NOTE: Change this date to whatever date you want to countdown to :)
+const HACKING_STARTS = "2024-01-26T18:00:00.000";
+const HACKING_ENDS = "2024-01-28T24:00:00.000";
```
samderanova, line 13: Issue [non-blocking]: 24:00:00 isn't a valid time, but we can update both of these later.

samderanova (approved):

alexanderl19 (approved):

PR #63, opened 2023-12-09
Add preliminary Apply page and application form

Bl20052005:
initial pr for application setup, still some work to do


apps/site/src/app/apply/sections/Form/SchoolInformation.tsx
```diff
@@ -0,0 +1,80 @@
+import styles from "./Form.module.scss";
+import DropdownSelect from "../Components/dropdownSelect";
+
+//I'm not sure if we want to have different submission values
+//than display texts for the inputs, so I seperated both
+//for good measure
```
samderanova, line 6: This is fine! We'll need to process these in the backend, so we should keep them separate as you have it.

apps/site/src/app/apply/sections/Form/SchoolInformation.tsx
```diff
@@ -0,0 +1,80 @@
...
+            <div className="flex gap-5 w-full">
+                <div className="flex flex-col w-full">
+                    <label className={`${styles.label}`} htmlFor="major">What is your major? <span className="text-[#FF2222]">*</span></label>
+                    <input className={`${styles.input}`} type="text" name="major" required/>
+                </div>
+            </div>
+
+            <div className="flex gap-5 w-full">
+                <div className="flex flex-row items-center gap-5 max-[600px]:flex-col max-[600px]:items-center max-[600px]:gap-1 max-[600px]:w-full text-center">
+                    <p className={`text-lg mb-0 p-0`}>Is this your first Hacakthon? <span className="text-[#FF2222]">*</span></p>
```
samderanova, line 63: Typo: `hackathon`, not `"Hacakthon"

apps/site/src/app/apply/sections/Title/Title.tsx
```diff
@@ -0,0 +1,10 @@
+import styles from "./Title.module.scss";
+
+export default function Title() {
+    return(
+        <>
+            <h1 className={`${styles.title} font-display text-[#fffce2] text-8xl text-center max-[500px]:text-7xl`}>Apply</h1>
+            <h2 className="text-[#fffce2] text-4xl font-bold m-5 text-center max-[900px]:text-3xl max-[500px]:text-2xl">Applications close on January 30th, 11:59PM PST</h2>
```
samderanova, line 7: Issue: Applications close on January 14th, 2024.

apps/site/src/app/layout.tsx
```diff
@@ -2,6 +2,7 @@ import type { Metadata } from "next";
 import water from "@/assets/backgrounds/water.jpg";
 import Footer from "@/lib/components/Footer/Footer";
 import "./globals.css";
+import Apply from "./apply/page";
```
samderanova, line 5: Nitpick: we have a certain order of importing things. See [this reply](https://github.com/HackAtUCI/irvinehacks-site-2024/pull/60#discussion_r1417792469) from Ian's PR.

apps/site/src/assets/images/koi-swim-left.png
samderanova, line 1: Question: was it possible to export this as an SVG without it having an `image` tag inside of it?

apps/site/src/app/apply/sections/Form/BasicInformation.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+            <div className="flex gap-5 w-full max-[1000px]:flex-col max-[1000px]:items-center">
+                <div className="flex flex-col w-6/12 max-[1000px]:w-full">
+                    <label className={`${styles.label}`} htmlFor="password">Password <span className="text-[#FF2222]">*</span></label>
+                    <div className="w-full flex items-center gap-3">
+                        <input
+                        className={`${styles.input} w-full`} 
+                        type={passwordVisible ? "text" : "password"}
+                        name="password" 
+                        pattern="(?=.*\d)(?=.*[a-z])(?=.*[A-Z]).{8,}"
```
samderanova, line 78: Praise: great job with the regex!

samderanova (changes requested):
Thanks for getting started on this! I've left a lot of comments, so please let me know if you have any questions on them.

Additionally, I'm not sure what could be happening but the code doesn't seem to be formatted. Is Prettier formatting on save?

apps/site/src/app/apply/sections/Form/BasicInformation.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const [passwordVisible, setPasswordVisible] = useState(false);
+    const [confirmPasswordVisible, setConfirmPasswordVisible] = useState(false);
+
+    const handleChangePassword = () => {
+        setPasswordVisible(!passwordVisible)
+    }
+    
+    const handleChangeConfirmPassword = () => {
+        setConfirmPasswordVisible(!confirmPasswordVisible)
+    }
```
taesungh, line 46: Note: we won't need any of this as users will be required to log in before being able to view the application.

apps/site/src/app/apply/sections/Form/SchoolInformation.tsx
```diff
@@ -0,0 +1,80 @@
...
+                    <div className="flex gap-2 items-center">
+                        <input type="radio" id="hack-yes" name="hack-check" value="Yes" defaultChecked={true} required/>
```
tyleryy, line 65: Should we have it default to "Yes" or just be blank? I feel the latter is better. @samderanova

apps/site/src/app/apply/sections/Title/Title.tsx
```diff
@@ -0,0 +1,10 @@
+import styles from "./Title.module.scss";
+
+export default function Title() {
+    return(
+        <>
+            <h1 className={`${styles.title} font-display text-[#fffce2] text-8xl text-center max-[500px]:text-7xl`}>Apply</h1>
```
tyleryy, line 6: I feel "Apply" as a title of the page looks off. Maybe something like "IrvineHacks Application Form"?

apps/site/src/app/apply/sections/Form/SchoolInformation.tsx
```diff
@@ -0,0 +1,80 @@
...
+];
+
+export default function SchoolInformation() {
+    return(
+        <div className="flex flex-col gap-5 w-11/12">
+            <p className="text-4xl m-0 font-bold text-center max-[700px]:text-3xl">Education</p>
+
+            <div className="flex gap-5 w-full max-[1000px]:flex-col max-[1000px]:items-center">
+                <div className="flex flex-col w-6/12 max-[1000px]:w-full">
+                    <DropdownSelect
```
tyleryy, line 36: For both dropdowns, I feel like having default values look off.

apps/site/src/app/apply/sections/Form/AgeInformation.tsx
```diff
@@ -0,0 +1,24 @@
...
+            <div className="flex flex-col gap-5">
+                <p className="m-0 text-lg">Because of limitations imposed by UCI, we are legally not allowed to host minors (those under 18) for IrvineHacks 2024. By answering yes, you affirm that you are and will be 18 years or older by January 26, 2024.</p>
+                <p className="text-[#FF2222] m-0 text-lg">We will be checking ID. If you are a minor, you will be turned away at the door.</p>
+            </div>
+
+            <div className="flex flex-col gap-1 w-full items-center">
+                <p className="text-xl font-bold m-0">Will you be 18 years or older by January 26th, 2024? <span className="text-[#FF2222]">*</span></p>
+                <div className="flex gap-5">
+                    <div className="flex gap-2 items-center">
+                        <input type="radio" id="minor-yes" name="minor-check" value="Yes" defaultChecked={true} required/>
```
tyleryy, line 13: repeating: i don't think we should have defaults.

tyleryy (changes requested):

tyleryy:
Offnote: I feel a component library might have helped here. But it is already built so moot point.

apps/site/src/app/apply/sections/Form/BasicInformation.tsx
Bl20052005: huh, so would we even need a password at all?

apps/site/src/app/apply/sections/Title/Title.tsx
Bl20052005: the "Apply" was on the figma page, and "IrvineHacks Application Form" is a tiny bit long in my opinion but I don't really know; I can change if it you think it would look best

apps/site/src/app/apply/sections/Form/BasicInformation.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    return(
+        <div className="flex flex-col gap-5 w-11/12">
+            <p className="text-4xl m-0 font-bold text-center max-[700px]:text-3xl">Basic Information</p>
+
+            <div className="flex gap-5 w-full max-[1000px]:flex-col max-[1000px]:items-center">
+                <div className="flex flex-col w-6/12 max-[1000px]:w-full">
+                    <label className={`${styles.label}`} htmlFor="first-name">First Name <span className="text-[#FF2222]">*</span></label>
+                    <input className={`${styles.input}`} type="text" name="first-name" required/>
```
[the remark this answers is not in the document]
Bl20052005: Is it fine if I put all IDs the same as the names? I was also a bit worried about duplicate IDs but I suppose that's not going to happen as long as the radios / selects all have differing preemptive identifiers passed in

apps/site/src/app/apply/sections/Title/Title.tsx
tyleryy: cc: @samderanova

apps/site/src/app/apply/sections/Components/DropdownSelect.tsx
```diff
@@ -0,0 +1,65 @@
...
+	labelText: string;
+	values: Array<{ value: string; text: string }>;
+}
+
+interface OtherProps {
+	value: string;
+	name: string;
+}
+
+const OtherPopup = (props: OtherProps) => {
```
tyleryy, line 19: Decompose these props

tyleryy, line 19: repeating: decompose props

apps/site/src/app/apply/sections/Components/DropdownSelect.tsx
```diff
@@ -0,0 +1,65 @@
...
+					id={`${props.name}-other-input`}
+					className="border-b-2 p-1 h-6 border-black w-6/12"
+					required
+				/>
+			</div>
+		);
+	}
+};
+
+export default function DropdownSelect(props: SelectProps) {
```
tyleryy, line 41: repeating: decompose these props

apps/site/src/app/apply/sections/Form/ProfileInformation.tsx
```diff
@@ -0,0 +1,67 @@
...
+						className={`${styles.input}`}
+						type="url"
+						name="portfolio"
+						id="portfolio"
+						placeholder="http://"
+					/>
+				</div>
+			</div>
+
+			<div className="flex gap-5 w-full">
```
tyleryy, line 38: Probably make these two text boxes for paragraph answers a component. That way we can pass a prop in for the question text instead of hardcoding it and also increase the number of paragraph questions we want in the app.

apps/site/src/app/apply/sections/Form/ProfileInformation.tsx
```diff
@@ -0,0 +1,67 @@
...
+import styles from "./Form.module.scss";
+
+export default function ProfileInformation() {
+	return (
+		<div className="flex flex-col gap-5 w-11/12">
+			<p className="text-4xl m-0 font-bold text-center max-[700px]:text-3xl">
+				Profile Information
+			</p>
+
+			<div className="flex gap-5 w-full max-[1000px]:flex-col max-[1000px]:items-center">
```
tyleryy, line 11: Perhaps it could be useful to make these textboxes components. They look the same in structure.

apps/site/src/app/apply/sections/Form/SchoolInformation.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+export default function SchoolInformation() {
+	return (
+		<div className="flex flex-col gap-5 w-11/12">
+			<p className="text-4xl m-0 font-bold text-center max-[700px]:text-3xl">
+				Education
+			</p>
+
+			<div className="flex gap-5 w-full max-[1000px]:flex-col max-[1000px]:items-center">
+				<div className="flex flex-col w-6/12 max-[1000px]:w-full">
```
tyleryy, line 35: repeating: maybe make the dropdown menus into components

apps/site/src/app/apply/sections/Form/SchoolInformation.tsx
```diff
@@ -0,0 +1,106 @@
...
+				</div>
+			</div>
+
+			<div className="flex gap-5 w-full">
+				<div className="flex flex-row items-center gap-5 max-[600px]:flex-col max-[600px]:items-center max-[600px]:gap-1 max-[600px]:w-full text-center">
+					<p className={`text-lg mb-0 p-0`}>
+						Is this your first Hackathon?{" "}
+						<span className="text-[#FF2222]">*</span>
+					</p>
+					<div className="flex gap-2 items-center">
```
tyleryy, line 78: repeating: radio component

apps/site/src/app/apply/sections/Form/BasicInformation.tsx
```diff
@@ -0,0 +1,165 @@
...
+	};
+
+	return (
+		<div className="flex flex-col gap-5 w-11/12">
+			<p className="text-4xl m-0 font-bold text-center max-[700px]:text-3xl">
+				Basic Information
+			</p>
+
+			<div className="flex gap-5 w-full max-[1000px]:flex-col max-[1000px]:items-center">
+				<div className="flex flex-col w-6/12 max-[1000px]:w-full">
```
tyleryy, line 54: repeating: textfield component

apps/site/src/app/apply/sections/Form/BasicInformation.tsx
```diff
@@ -0,0 +1,165 @@
...
+						className={`${styles.input}`}
+						type="text"
+						name="last-name"
+						id="last-name"
+						required
+					/>
+				</div>
+			</div>
+
+			<div className="flex gap-5 w-full max-[1000px]:flex-col max-[1000px]:items-center">
```
tyleryy, line 80: repeating: another textfield component or with a prop to make it variable width size, which you can achieve since Tailwind is like inline styles

apps/site/src/app/apply/sections/Form/BasicInformation.tsx
```diff
@@ -0,0 +1,165 @@
...
+						/>
+						<PasswordEyeProps
+							visible={confirmPasswordVisible}
+							handler={handleChangeConfirmPassword}
+						/>
+					</div>
+				</div>
+			</div>
+
+			<div className="flex gap-5 w-full max-[1000px]:flex-col max-[1000px]:items-center">
```
tyleryy, line 145: repeating: maybe radio component with scalable radio options

tyleryy (changes requested):
- [x] Blocking: we could create components for each of the repeatable parts of the form (dropdowns, textfields, paragraph blocks, etc.). It would make the code more DRY and less boilerplatey
- [ ] suggestion: I feel the color theme for the app form is not that good. I personally don't like the white background color. Maybe a different color theme would be better to fit our "Night on Lunar New Year" cc: Graphics
- [ ] suggestion: Experiment with making the Submit Application button pop out more (maybe make it glow more or smth cause I feel it is lackluster)
- [ ] suggestion: I still think "Apply" looks incredibly jarring. Like it kinda scares my eyes away from it. cc: Graphics

Why didn't we use a component library here? The form is pretty much unstyled anyway, so using prebuilt components wouldn't be a downside.
Example: https://mui.com/material-ui/react-radio-button/

apps/site/src/app/apply/sections/Title/Title.tsx
samderanova: We should stick with Apply. I personally don't see an issue with it, especially since we used it last year.

apps/site/src/app/apply/sections/Form/SchoolInformation.tsx
samderanova: I agree, we should leave it blank.

samderanova:
I should have suggested the use of Radix in this case.

> * [ ]  Blocking: we could create components for each of the repeatable parts of the form (dropdowns, textfields, paragraph blocks, etc.). It would make the code more DRY and less boilerplatey
> * [ ]  suggestion: I feel the color theme for the app form is not that good. I personally don't like the white background color. Maybe a different color theme would be better to fit our "Night on Lunar New Year" cc: Graphics
> * [ ]  suggestion: Experiment with making the Submit Application button pop out more (maybe make it glow more or smth cause I feel it is lackluster)
> * [ ]  suggestion: I still think "Apply" looks incredibly jarring. Like it kinda scares my eyes away from it. cc: Graphics
> 
> Why didn't we use a component library here? The form is pretty much unstyled anyway, so using prebuilt components wouldn't be a downside. Example: https://mui.com/material-ui/react-radio-button/

I should have suggested the use of Radix here. I'm slightly concerned about installing Material just so we can use radio buttons.

We can transition to using components in a separate issue. For now, @Bl20052005 , please resolve the issues Tyler brought up, including extracting out the components so that you don't repeat yourself too much, and then re-request reviews from us.

tyleryy:
> I should have suggested the use of Radix here. I'm slightly concerned about installing Material just so we can use radio buttons.

Material has other components as well textfields, dropdowns, etc. Tho I agree adding another component library is too much. I think they allow you to only import certain components, not sure if it is worth looking into tho.

A possible alternative or improvement is to use [shadcn](https://ui.shadcn.com/). You use a cli tool to download only the components you need, so it is lightweight. Plus it is built on Radix and Tailwind which we use already

apps/site/src/app/apply/sections/Components/PasswordEye.tsx
```diff
@@ -0,0 +1,14 @@
+import { Eye, EyeOff } from "lucide-react";
+
+interface PasswordProps {
+	visible: boolean;
+	handler: React.MouseEventHandler<SVGElement>;
+}
+
+export default function PasswordEyeProps({ visible, handler }: PasswordProps) {
```
tyleryy, line 8: I don't think this should be named with "Props" since it is a component

tyleryy (changes requested):

apps/site/src/app/apply/sections/Components/PasswordEye.tsx
tyleryy: Or actually I think @taesungh wants the password section remove all together, if I understand his comment correctly. So, you probably don't need to do this.

samderanova:
> Or actually I think @taesungh wants the password section remove all together, if I understand his comment correctly. So, you probably don't need to do this.

Yeah @Bl20052005, as both @taesungh and @tyleryy have said, we don't need the username or password section because they'll be logging in before accessing the application

apps/site/src/app/apply/sections/Components/DropdownSelect.tsx
```diff
@@ -0,0 +1,70 @@
...
+	containerClass,
+}: SelectProps) {
+	const [value, setValue] = useState("");
+
+	return (
+		<div className={containerClass}>
+			<label className={`${labelStyle}`} htmlFor={name}>
+				{labelText} <RequiredAsterisk />
+			</label>
+			<select
```
samderanova, line 54: Issue: This `select` field needs to be accessible, meaning clicking the label should focus the user's cursor on the `select` field. This can be done by setting the `htmlFor` of the `label` to the `id` of the `select` field, not the `name`.

samderanova (commented):
Nice work with the fixes! As I mentioned, in my comments, I'll take care of the logic for automatically focusing the cursor on the input fields tied to the `Other` radios.

 A few more things to address:

apps/site/src/app/apply/sections/Components/SimpleRadio.tsx
```diff
@@ -0,0 +1,62 @@
...
+	containerClassInputLabels,
+	containerClassValues,
+	labelClass,
+	isRequired,
+}: TextfieldProps) {
+	return (
+		<div className={containerClassTotal}>
+			<p className={titleClass}>
+				{`${title} `}
+				{isRequired ? <RequiredAsterisk /> : <></>}
```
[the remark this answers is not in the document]
Bl20052005: I looked through the docs and it seems that ?? only works with null / undefined, which is a bit problematic because isRequired is a boolean, so I'm not entirely sure how I would implement this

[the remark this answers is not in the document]
samderanova: My mistake, I wasn't thinking straight. Sure, that works.

samderanova (approved):
Ok looking good for now, there are still some things we should address, such as using a component-based UI frameworks ([shadcn](https://ui.shadcn.com/) for example), but we can address these in later issues as this PR is getting quite large.

apps/site/src/app/apply/sections/Components/SimpleRadio.tsx
```diff
@@ -0,0 +1,59 @@
...
+	containerClassInputLabels,
+	containerClassValues,
+	labelClass,
+	isRequired,
+}: TextfieldProps) {
+	return (
+		<div className={containerClassTotal}>
+			<p className={titleClass}>
+				{`${title} `}
+				{!isRequired || <RequiredAsterisk />}
```
taesungh, line 34: Nitpick: using negation with disjunction mangles the interpretability a little. A direct conjunction could be easier to understatnd
```suggestion
				{isRequired && <RequiredAsterisk />}
```

apps/site/src/app/apply/sections/Components/SimpleRadio.tsx
```diff
@@ -0,0 +1,59 @@
...
+		<div className={containerClassTotal}>
+			<p className={titleClass}>
+				{`${title} `}
+				{isRequired && <RequiredAsterisk />}
+			</p>
+			<div className={containerClassValues}>
+				{values.map((value, i) => {
+					return (
+						<div
+							key={`${name}-${i}`}
```
taesungh, line 40: Suggestion: if all of the values are unique, we don't need to rely on the index for the key.

apps/site/src/app/apply/sections/Form/ProfileInformation.tsx
```diff
@@ -0,0 +1,55 @@
...
+				<TextInput
+					name="linkedin"
+					labelClass={styles.label}
+					labelText="LinkedIn"
+					inputClass={styles.input}
+					containerClass="flex flex-col w-6/12 max-[1000px]:w-full"
+					isRequired={false}
+					type="url"
+					placeholder="http://"
```
taesungh, line 22: Suggestion [repeating]: if we are using a placeholder, we should suggest https instead of plain http.

apps/site/src/app/apply/sections/Form/AgeInformation.tsx
```diff
@@ -0,0 +1,46 @@
+import RequiredAsterisk from "../Components/RequiredAsterisk";
+import SimpleRadio from "../Components/SimpleRadio";
```
taesungh, line 2: Chore: relative imports are usually more cumbersome to manage. We can use the alias for the src directory and use absolute imports.

apps/site/src/app/apply/sections/Components/SimpleRadio.tsx
Bl20052005: I was thinking that since technically nothing restricts the values from being repeated (such as having the same id), I personally don't see why using index for key is necessarily a bad idea, but I'd be interested in more elaboration

apps/site/src/app/apply/sections/Form/AgeInformation.tsx
Bl20052005: I'm not entirely sure how to do this or what this is referring to to be honest, so I might not be able to implement this

taesungh: We can specify imports from the top-level down using `@` to reference our `src` directory instead of using `../` to traverse relative paths. Refer to [About.tsx](https://github.com/HackAtUCI/irvinehacks-site-2024/blob/main/apps/site/src/app/(home)/sections/About/About.tsx) for an example.
```tsx
import { thing } from "@/lib/components"
```

Bl20052005: should we also do this to files that are in the same directory (./)?

samderanova: No, for the files that are in the same directory, just use `./`.

Bl20052005: got it, then this should be set

apps/site/src/app/apply/sections/Components/SimpleRadio.tsx
samderanova: See the React documentation on [rendering lists](https://react.dev/learn/rendering-lists).

> You might be tempted to use an item’s index in the array as its key. In fact, that’s what React will use if you don’t specify a key at all. But the order in which you render items will change over time if an item is inserted, deleted, or if the array gets reordered. Index as a key often leads to subtle and confusing bugs.

Bl20052005: hm, then it sounds like a more randomly-generated key would be preferrable (like the crypto.randomUUID() the doc mentioed)? But it would be a bit difficult to have them not change on render.

I don't really know what's the best solution here to be honest, I guess in the context that we're using this component we can definitely use the name / id of values though since I'm not sure how the random key would work (and it seems that indices are kind of just a fallback option that we should use when we have nothing else)

apps/site/src/app/apply/sections/Components/RadioSelect.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+	return (
+		<div className={containerClass}>
+			<p className="m-0 text-lg mb-4">
+				{labelText} <span className="text-[#FF2222]">*</span>
+			</p>
+			<div className="w-10/12 flex flex-col gap-2">
+				{values.map((item, i) => {
+					const keyandId = `${IdentifierId}-${i}`;
+					if (item.value == "other") {
```
samderanova, line 60: Nitpick: We generally want to use strict equality (i.e. `===` instead of `==`) to prevent any unwanted behavior.

apps/site/src/app/apply/sections/Components/SimpleRadio.tsx
```diff
@@ -0,0 +1,59 @@
...
+	containerClassTotal,
+	containerClassInputLabels,
+	containerClassValues,
+	labelClass,
+	isRequired,
+}: TextfieldProps) {
+	return (
+		<div className={containerClassTotal}>
+			<p className={titleClass}>
+				{`${title} `}
```
samderanova, line 33: Nitpick: Since there's no other TS code inside the `{}`, you can simplify line 33 to `{title}`. Same for the `labelText` in line 27 of `TextInput.tsx`.

apps/site/src/app/apply/sections/Components/SimpleRadio.tsx
samderanova: Random keys also are not a good idea. From the same section of the docs:

> Similarly, do not generate keys on the fly, e.g. with key={Math.random()}. This will cause keys to never match up between renders, leading to all your components and DOM being recreated every time. Not only is this slow, but it will also lose any user input inside the list items. Instead, use a stable ID based on the data.

You've set all the names to be unique and they should be that way, or else we wouldn't be able to process the form properly, so you can just remove the index part (the `-${i}`) from the key.

samderanova (changes requested):

apps/site/src/app/apply/sections/Components/SimpleRadio.tsx
Bl20052005: there is a space after the title

apps/site/src/app/apply/sections/Components/RadioSelect.tsx
taesungh: Hmm, did eslint not pick this up?

apps/site/src/app/apply/sections/Components/TextInput.tsx
```diff
@@ -0,0 +1,39 @@
...
+	inputClass,
+	containerClass,
+	placeholder,
+	type,
+	isRequired,
+}: TextProps) {
+	return (
+		<div className={containerClass}>
+			<label className={labelClass} htmlFor={name}>
+				{`${labelText} `} {isRequired && <RequiredAsterisk />}
```
taesungh, line 27: Thought: given most of the fields on the application are required, having the asterisk everywhere seems a bit excessive. Although we don't have an official style guide, here's one such suggestion from Material Design
> If most fields are required, indicate optional fields by displaying the word “optional” in parentheses next to the label text ([Material Design: Text fields - Guidelines](https://m3.material.io/components/text-fields/guidelines))

apps/site/src/app/apply/sections/Form/ProfileInformation.tsx
```diff
@@ -0,0 +1,54 @@
...
+				labelText="Why is collaboration important to being a programmer or technologist, and what does it mean to you? (150 words)"
+				inputClass={`bg-[#E1E1E1] p-3 h-48 resize-none rounded-xl`}
+				containerClass="flex flex-col w-full"
+				isRequired={true}
+			/>
+
+			<Textfield
+				name="job-answer"
+				labelClass={`${styles.label} mt-7`}
+				labelText="If you could do any job, what would it be? (ex. Youtuber, Body Builder, etc.) (100 words)"
```
rileysw, line 47: suggestion: If you could have any job in the world, what would it be?

apps/site/src/app/apply/sections/Components/TextInput.tsx
Bl20052005: according to this, it seems like we either want to move the asterisk to the placeholder, or have a subscript of small text that labels it as required, but all of this would still include the asterisk on required attributes

do you think it would be a better idea to specify which ones are optional instead by placing (optional) in the placeholder text?

And it seems like having a red asterisk isn't necessarily a terrible design choice, as I've seen many other forms (particularly, google forms) use this convention on every field notating a required attribute

rileysw (changes requested):
Sorry, a lot of these comments should've been reviewed among log before sending over to tech. Thanks Bo for the awesome job so far!

tyleryy:
Question: should we be using sanity for this page? Or are we hardcoding it? That way Log can make changes in realtime / on the fly.

samderanova:
We're just hard coding it since the questions have already been finalized.

apps/site/src/app/apply/sections/Form/SchoolInformation.tsx
```diff
@@ -0,0 +1,113 @@
...
+	{ value: "second-year-undergrad", text: "Second Year Undergraduate" },
+	{ value: "third-year-undergrad", text: "Third Year Undergraduate" },
+	{ value: "fourth-year-undergrad", text: "Fourth Year Undergraduate" },
+	{ value: "fifth-year-undergrad", text: "Fifth+ Year Undergraduate" },
+	{ value: "graduate", text: "Graduate" },
+];
+const universityOptions = [
+	{ value: "UC Irvine", text: "UC Irvine" },
+	{ value: "Cal Poly Pomona", text: "Cal Poly Pomona" },
+	{ value: "Cal State Long Beach", text: "Cal State Long Beach" },
```
rileysw, line 19: CSF before CSLB

apps/site/src/app/apply/sections/Form/ProfileInformation.tsx
```diff
@@ -44,7 +44,7 @@ export default function ProfileInformation() {
 			<Textfield
 				name="job-answer"
 				labelClass={`${styles.label} mt-7`}
-				labelText="If you could do any job, what would it be? (ex. Youtuber, Body Builder, etc.) (100 words)"
+				labelText="If you could have any job in the world, what would it be? (ex. Youtuber, Body Builder, etc.) (100 words)"
```
taesungh, line 47: Nitpick: T in YouTuber should be capitalized, bodybuilder is more commonly written as one word

rileysw (approved):
Two more typo changes, otherwise looks good from log!

apps/site/src/app/apply/sections/Components/SimpleRadio.tsx
taesungh: Given the purpose of this (and similar) components is for display options of form elements, it wouldn't make sense for the consumer to provide duplicate values. Using the value as the key would also report such a mistake which could go unnoticed if the index were combined.

apps/site/src/app/apply/sections/Components/TextInput.tsx
taesungh: Yes, I was suggesting the second as I quoted in the design guidelines. For Google Forms, I would say they are designed to need to accommodate custom forms that the user would construct and as such can't make as flexible design choices. We can revisit this and discuss with graphics as well.

taesungh (approved):
Let's call it here, we can fix smaller lingering issues separately.

samderanova (approved):
Agreed, we can fix any lingering issues later. Will be squash merging.

PR #64, opened 2023-12-10
Add workflow to run checks for the API and site

taesungh:
As we did before with last year's [HackUCI-Site](https://github.com/HackAtUCI/HackUCI-Site) repository, we can use a GitHub Actions workflow to run checks on pull requests and pushes to `main`.

The checks run for this pull request will initially fail because there is a type error in `saml.py` which will demonstrate that this runner works.

## Changes
- Include Python dev dependencies for checks runner
- Configure Python linting tools and provide compatible VS Code settings
- Update production deployment to run only when checks succeed

### DIFF
--- a/apps/api/setup.cfg
+++ b/apps/api/setup.cfg
@@ -1,0 +1,3 @@
+[flake8]
+max-line-length = 88
+extend-ignore = E203
--- a/.env
+++ b/.env
@@ -1,1 +1,2 @@
+# For VS Code Python extension
 PYTHONPATH=apps/api/src
--- a/.github/workflows/deploy-prod.yml
+++ b/.github/workflows/deploy-prod.yml
@@ -1,7 +1,11 @@
 name: Deploy Production
 
 on:
-    push:
+    workflow_run:
+        workflows:
+            - Run Checks
+        types:
+            - completed
         branches:
             - main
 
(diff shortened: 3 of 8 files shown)


taesungh:
Update: the relevant SAML is not spotted during pull request since those lines were not changed but would be caught if we merged and pushed to `main`. Will try adding some files with errors on purpose to demonstrate the concept.

apps/api/setup.cfg
```diff
@@ -0,0 +1,3 @@
+[flake8]
+max-line-length = 88
```
samderanova, line 2: Question: I'm pretty sure I asked this already, but why 88 instead of [79](https://peps.python.org/pep-0008/#maximum-line-length) or 80? It does say we can increase it up to 99 though.

samderanova (dismissed):
Looking good, thanks for working on this. One small question:

apps/api/setup.cfg
taesungh: This is in accordance with the [line length in Black's code style](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#labels-line-length).

samderanova:
Sounds good, please squash merge.

taesungh:
The plan is to have #65 on standby and then merge this PR to verify the checks workflow fails because of a lingering type issue. Once that is verified, we can merge #65 to resolve the issue.

samderanova (approved):

PR #72, opened 2023-12-12
Integrate MongoDB Database

samderanova:
# Overview
One of the challenges we faced last year was the inability to test database functions locally. This pull request resolves this issue by, as @taesungh suggested, using [Docker](https://docs.docker.com/get-docker/) to create an instance of MongoDB. Specifically, there is now a [Docker Compose](https://docs.docker.com/compose/) file that launches three containers: one that runs the backend, one that runs a MongoDB instance, and one that runs a user interface for the instance.

# Docker
For those of you who are unfamiliar with Docker, [Docker](https://www.docker.com/get-started/) is a tool that allows developers to deploy applications in sandbox environments known as containers. See our [Docker starter pack](https://github.com/HackAtUCI/starter-pack-docker) for an example on how Docker (and Docker Compose) can be used.

Note that Vercel [does not support](https://vercel.com/guides/does-vercel-support-docker-deployments) Docker deployments, so the database functions will be run on the cloud database using a non-production collection, not on your local machine. Thus, the Docker containers should only be run in your local environments.

# Setup
1. Firstly, please install [Docker](https://docs.docker.com/get-docker/). The process varies greatly depending on the operating system, so please refer to this article.
2. Next, open Docker Desktop.
3. Open a terminal or your editor and navigate to the project. Running `pnpm dev` should be all you need to get started!
    - You can access the aforementioned web interface for the database at `http://localhost:8081`. 

One of the cons of this approach is that hot reload is no longer possible because the Python files have been copied over to the Docker container. This means that stopping the container and rerunning `pnpm dev` is the best option.

### DIFF
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -1,11 +1,14 @@
 FROM python:3.9
 
-COPY requirements.txt requirements.txt
-
 RUN apt-get update && apt-get install -y libxml2-dev libxmlsec1-dev libxmlsec1-openssl
+
+WORKDIR /app
+
+COPY requirements.txt .
+
 RUN pip install --no-cache-dir -r requirements.txt "uvicorn[standard]"
 
-COPY src/ src/
+COPY src/ .
 
 ENV PYTHONPATH=src/
 
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -1,9 +1,12 @@
 FROM python:3.9
 
-COPY . .
+COPY requirements.txt requirements.txt
 
-RUN apt-get update && apt-get upgrade -y
-RUN apt-get install -y pkg-config libxml2-dev libxmlsec1-dev libxmlsec1-openssl
-RUN pip install -r requirements.txt -r requirements-dev.txt
+RUN apt-get update && apt-get install -y libxml2-dev libxmlsec1-dev libxmlsec1-openssl
+RUN pip install --no-cache-dir -r requirements.txt "uvicorn[standard]"
+
+EXPOSE 8000
+
+COPY . .
 
 CMD ["python3", "src/dev.py"]
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -7,6 +7,6 @@ RUN pip install --no-cache-dir -r requirements.txt "uvicorn[standard]"
 
 EXPOSE 8000
 
-COPY . .
+COPY src/ src/
 
 CMD ["python3", "src/dev.py"]
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -9,4 +9,6 @@ EXPOSE 8000
 
 COPY src/ src/
 
-CMD ["python3", "src/dev.py"]
+ENV PYTHONPATH=src/
+
+CMD ["uvicorn", "app:app", "--host", "0.0.0.0"]
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -1,0 +1,9 @@
+FROM python:3.9
+
+COPY . .
+
+RUN apt-get update && apt-get upgrade -y
+RUN apt-get install -y pkg-config libxml2-dev libxmlsec1-dev libxmlsec1-openssl
+RUN pip install -r requirements.txt -r requirements-dev.txt
+
+CMD ["python3", "src/dev.py"]
--- a/apps/api/Dockerfile
+++ b/apps/api/Dockerfile
@@ -5,8 +5,6 @@ COPY requirements.txt requirements.txt
 RUN apt-get update && apt-get install -y libxml2-dev libxmlsec1-dev libxmlsec1-openssl
 RUN pip install --no-cache-dir -r requirements.txt "uvicorn[standard]"
 
-EXPOSE 8000
-
 COPY src/ src/
 
 ENV PYTHONPATH=src/
--- a/apps/api/docker-compose.yml
+++ b/apps/api/docker-compose.yml
@@ -1,0 +1,31 @@
+# Use root/example as user/password credentials
+version: "3.1"
+
+services:
+    mongo:
+        image: mongo
+        restart: always
+        ports:
+            - "27017:27017"
+        environment:
+            MONGO_INITDB_ROOT_USERNAME: root
+            MONGO_INITDB_ROOT_PASSWORD: example
+
+    mongo-express:
+        image: mongo-express
+        restart: always
+        ports:
+            - "8081:8081"
+        environment:
+            ME_CONFIG_MONGODB_ADMINUSERNAME: root
+            ME_CONFIG_MONGODB_ADMINPASSWORD: example
+
+            ME_CONFIG_MONGODB_URL: mongodb://root:example@mongo:27017/
+
+    fastapi-dev:
+        network_mode: host
+        build: .
+        ports:
+            - "8000:8000"
+        environment:
+            MONGODB_URI: mongodb://root:example@127.0.0.1:27017
--- a/apps/api/docker-compose.yml
+++ b/apps/api/docker-compose.yml
@@ -5,8 +5,6 @@ services:
     mongo:
         image: mongo
         restart: always
-        ports:
-            - "27017:27017"
         environment:
             MONGO_INITDB_ROOT_USERNAME: root
             MONGO_INITDB_ROOT_PASSWORD: example
@@ -19,13 +17,11 @@ services:
         environment:
             ME_CONFIG_MONGODB_ADMINUSERNAME: root
             ME_CONFIG_MONGODB_ADMINPASSWORD: example
-
             ME_CONFIG_MONGODB_URL: mongodb://root:example@mongo:27017/
 
     fastapi-dev:
-        network_mode: host
         build: .
         ports:
             - "8000:8000"
         environment:
-            MONGODB_URI: mongodb://root:example@127.0.0.1:27017
+            MONGODB_URI: mongodb://root:example@mongo:27017
--- a/apps/api/docker-compose.yml
+++ b/apps/api/docker-compose.yml
@@ -8,6 +8,8 @@ services:
         environment:
             MONGO_INITDB_ROOT_USERNAME: root
             MONGO_INITDB_ROOT_PASSWORD: example
+        volumes:
+            - mongodb_data_volume:/data/db
 
     mongo-express:
         image: mongo-express
@@ -25,3 +27,6 @@ services:
             - "8000:8000"
         environment:
             MONGODB_URI: mongodb://root:example@mongo:27017
+
+volumes:
+    mongodb_data_volume:
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -2,7 +2,7 @@
 	"name": "api",
 	"private": true,
 	"scripts": {
-		"dev": "docker compose up --no-deps --build",
+		"dev": "./run-docker.sh",
 		"test": "pytest",
 		"lint": "flake8 src tests && mypy src tests",
 		"format:write": "black src tests",
--- a/apps/api/package.json
+++ b/apps/api/package.json
@@ -2,7 +2,7 @@
 	"name": "api",
 	"private": true,
 	"scripts": {
-		"dev": "python src/dev.py",
+		"dev": "docker compose up --no-deps --build",
 		"test": "pytest",
 		"lint": "flake8 src tests && mypy src tests",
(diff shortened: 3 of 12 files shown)


apps/api/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.9
+
+COPY . .
+
+RUN apt-get update && apt-get upgrade -y
+RUN apt-get install -y pkg-config libxml2-dev libxmlsec1-dev libxmlsec1-openssl
```
taesungh, line 6: Issue: the separation here could cause issues with Docker.
> Using `apt-get` update alone in a `RUN` statement causes caching issues and subsequent `apt-get install` instructions to fail. Always combine RUN apt-get update with apt-get install in the same RUN statement. ([Best Practices for Dockerfile Instructions: apt-get](https://docs.docker.com/develop/develop-images/instructions/#apt-get))

Also don't think we need to run `apt-get upgrade` or install `pkg-config`?

```suggestion
RUN apt-get update && apt-get install -y libxml2-dev libxmlsec1-dev libxmlsec1-openssl
```

apps/api/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.9
+
+COPY . .
```
taesungh, line 3: Suggestion: Copying source files as the first step defeats the purpose of the build cache. We can better use build cache by copying source code towards the end after installing relevant dependencies. See suggestion below on the line for `pip install ...`

apps/api/Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+
+RUN apt-get update && apt-get upgrade -y
+RUN apt-get install -y pkg-config libxml2-dev libxmlsec1-dev libxmlsec1-openssl
+RUN pip install -r requirements.txt -r requirements-dev.txt
```
taesungh, line 7: Suggestion: we can use the `--no-cache-dir` option to optimize the cache
> The --no-cache-dir option tells pip to not save the downloaded packages locally, as that is only if pip was going to be run again to install the same packages, but that's not the case when working with containers. ([FastAPI in Containers: Dockerfile](https://fastapi.tiangolo.com/deployment/docker/#dockerfile))

We also don't need all of the dev dependencies in the container as most are just for linting.
```suggestion
WORKDIR /code/
COPY requirements.txt .
RUN pip install --no-cache-dir -r requirements.txt "uvicorn[standard]"
COPY src/ src/
```

apps/api/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+
+services:
+    mongo:
+        image: mongo
+        restart: always
+        ports:
+            - "27017:27017"
+        environment:
+            MONGO_INITDB_ROOT_USERNAME: root
+            MONGO_INITDB_ROOT_PASSWORD: example
```
taesungh, line 12: Chore: could we look into using a named volume for persistent database storage?
Something like
```yaml
volumes:
  - type: volume
    source: mongodb_data_volume
    target: /data/db
```

apps/api/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+
+            ME_CONFIG_MONGODB_URL: mongodb://root:example@mongo:27017/
+
+    fastapi-dev:
+        network_mode: host
+        build: .
+        ports:
+            - "8000:8000"
+        environment:
+            MONGODB_URI: mongodb://root:example@127.0.0.1:27017
```
taesungh, line 31: Chore: we can reference the container directly as you did within `mongo-express` instead of using 127.0.0.1
```suggestion
            MONGODB_URI: mongodb://root:example@mongo:27017
```

apps/api/package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "api",
 	"private": true,
 	"scripts": {
-		"dev": "python src/dev.py",
+		"dev": "docker compose up --no-deps --build",
```
taesungh, line 5: Issue: if the concept is to run this with turbo dev, then `docker compose up` wouldn't quite work since the containers wouldn't ever be stopped (e.g. with `docker compose down`).

apps/api/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+
+    mongo-express:
+        image: mongo-express
+        restart: always
+        ports:
+            - "8081:8081"
+        environment:
+            ME_CONFIG_MONGODB_ADMINUSERNAME: root
+            ME_CONFIG_MONGODB_ADMINPASSWORD: example
+
```
taesungh, line 22: Nitpick: stray newline?

apps/api/src/routers/demo.py
```diff
@@ -15,3 +17,18 @@ async def me(username: Annotated[Union[str, None], Cookie()] = None) -> str:
 async def square(value: Annotated[int, Form()]) -> int:
     """Calculate the square of the value."""
     return value * value
+
+
+@router.post("/user")
+async def get_user(name: Annotated[str, Form()]) -> list[dict[str, Any]]:
+    results = await retrieve(Collection.USERS, {"name": name}, ["name", "ucinetid"])
+    for i in range(len(results)):
+        results[i]["_id"] = str(results[i]["_id"])
```
taesungh, line 26: Nitpick [non-blocking since just for demo]: we could use direct object iteration instead of range-based iteration.
```suggestion
    for result in results:
        result["_id"] = str(result["_id"])
```

apps/api/src/routers/demo.py
```diff
@@ -15,3 +17,18 @@ async def me(username: Annotated[Union[str, None], Cookie()] = None) -> str:
 async def square(value: Annotated[int, Form()]) -> int:
     """Calculate the square of the value."""
     return value * value
+
+
+@router.post("/user")
+async def get_user(name: Annotated[str, Form()]) -> list[dict[str, Any]]:
```
taesungh, line 23: Nitpick [non-blocking since just for demo]: return type could be more specific than `Any`, i.e. `str`.

apps/api/src/routers/demo.py
```diff
@@ -15,3 +17,18 @@ async def me(username: Annotated[Union[str, None], Cookie()] = None) -> str:
 async def square(value: Annotated[int, Form()]) -> int:
     """Calculate the square of the value."""
     return value * value
+
+
+@router.post("/user")
```
taesungh, line 22: Thought: since this is just fetching data, I think it would be more appropriate to specify the endpoint as a GET method?

taesungh (changes requested):
Thanks for getting to this. Some improvements to be made on the Dockerfile and minor notes about the demo API which aren't a huge deal since it is just for demonstration purposes. One lingering thought about how we'll actually have developers invoke `docker compose`.

apps/api/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
+# Use root/example as user/password credentials
+version: "3.1"
+
+services:
+    mongo:
+        image: mongo
+        restart: always
+        ports:
+            - "27017:27017"
```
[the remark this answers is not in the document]
samderanova: That's true, I was just thinking that if someone wanted to access the database via another method, ex. `mongosh`, having this port exposed would provide that option.

apps/api/src/routers/demo.py
samderanova: Shouldn't we keep it as `Any` since we could be storing different data types? Also, if we change it to `str` that results in some type errors because the return type for `retrieve` is `list[dict[str, object]]`.

apps/api/Dockerfile
samderanova: Ah, I was looking at the pip `xmlsec` package, which had `pkg-config` as one of the necessary libraries. I'll remove it.

apps/api/package.json
samderanova: Resolved this by running a script instead:

```
docker compose up
docker compose down
```

apps/api/run-docker.sh
```diff
@@ -0,0 +1,2 @@
+docker compose up
```
taesungh, line 1: Issue: I think we still need the `--build` option to rebuild the images when files change.

apps/api/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM python:3.9
+
+COPY requirements.txt requirements.txt
+
+RUN apt-get update && apt-get install -y libxml2-dev libxmlsec1-dev libxmlsec1-openssl
+RUN pip install --no-cache-dir -r requirements.txt "uvicorn[standard]"
+
+EXPOSE 8000
+
+COPY . .
```
taesungh, line 10: Suggestion: We don't need to copy everything to the image, only the `src` folder.
```suggestion
COPY src/ src/
```

apps/api/src/dev.py
```diff
@@ -3,7 +3,7 @@
 if __name__ == "__main__":
     uvicorn.run(
         "app:app",
-        host="localhost",
+        host="0.0.0.0",
```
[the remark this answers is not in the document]
samderanova: Will proceed with the `uvicorn` approach so we don't need to create such a flag/environment variable

apps/api/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM python:3.9
+
+COPY requirements.txt requirements.txt
+
+RUN apt-get update && apt-get install -y libxml2-dev libxmlsec1-dev libxmlsec1-openssl
+RUN pip install --no-cache-dir -r requirements.txt "uvicorn[standard]"
+
+EXPOSE 8000
```
taesungh, line 8: Question: do we need to expose the port here? I thought the `ports` part of the compose file would handle it.

apps/api/package.json
```diff
@@ -2,7 +2,7 @@
 	"name": "api",
 	"private": true,
 	"scripts": {
-		"dev": "python src/dev.py",
+		"dev": "./run-docker.sh",
```
taesungh, line 5: Thought: do we need an entire separate executable file for this, or could we write the statements together here?
```shell
docker compose up --build && docker compose down
```
Could see it being cleaner both ways, unsure which is better.

apps/api/src/services/mongodb_handler.py
```diff
@@ -0,0 +1,119 @@
...
+from pydantic import BaseModel, Field
+from pymongo.results import InsertOneResult, UpdateResult
+
+log = getLogger(__name__)
+
+STAGING_ENV = os.getenv("DEPLOYMENT") == "STAGING"
+
+MONGODB_URI = os.getenv("MONGODB_URI")
+MONGODB_CLIENT = AsyncIOMotorClient(MONGODB_URI)
+MONGODB_CLIENT.get_io_loop = asyncio.get_event_loop
```
taesungh, line 19: Thought: do we know if this workaround is still needed when running on Vercel?

apps/api/Dockerfile
samderanova: No, I was using this during a debugging process, since I was trying to run just this container by itself. I'll remove this.

apps/api/package.json
samderanova: If we have it this way though, I don't think `docker compose down` will run when pressing `ctrl + c`. I tried it on my own machine:

`./run-docker.sh`
image

vs. 

chained commands
image

taesungh: Got it, thanks for discovering how to get both up and down working with the file.

apps/api/src/services/mongodb_handler.py
samderanova: I removed it and tested the deployment. It still works, so we no longer need it.

apps/api/src/services/mongodb_handler.py
```diff
@@ -0,0 +1,123 @@
...
+class BaseRecord(BaseModel):
+    uid: str = Field(alias="_id")
+
+    def dict(self, *args: Any, **kwargs: Any) -> dict[str, Any]:
+        if "by_alias" in kwargs:
+            return BaseModel.model_dump(self, *args, **kwargs)
+        return BaseModel.model_dump(self, by_alias=True, *args, **kwargs)
+
+    class Config:
+        populate_by_name = True
```
taesungh, line 40: Runtime warning
> PydanticDeprecatedSince20: Support for class-based `config` is deprecated, use ConfigDict instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.5/migration/

taesungh (approved):
Nice work, thanks for including the demo and the Docker files which will be really helpful.

PR #73, opened 2023-12-12
Set up basic SendGrid Handler

samderanova:
This pull request includes a function that sends emails using [SendGrid](https://github.com/sendgrid/sendgrid-python).

Required environment variables:
- `SENDGRID_API_KEY`: needed to use the SendGrid API. This has already been added to Vercel.


apps/api/src/services/sendgrid_handler.py
```diff
@@ -0,0 +1,64 @@
...
+                    )
+                )
+                email_message.add_personalization(p)
+
+        email_message.from_email = sender_email
+        email_message.template_id = template_id
+
+        async with aiosendgrid.AsyncSendGridClient(api_key=SENDGRID_API_KEY) as client:
+            response = await client.send_mail_v3(body=email_message.get())
+            print(response.status_code)
```
tyleryy, line 60: nitpick: should this be print or log debug/info?

tyleryy (approved):

taesungh (commented):
Could we also import the unit tests from [`test_sendgrid_handler.py`](https://github.com/HackAtUCI/HackUCI-Site/blob/main/tests/test_sendgrid_handler.py)?
We'll also need the [`email_handler.py`](https://github.com/HackAtUCI/HackUCI-Site/blob/main/src/api/utils/email_handler.py) at some point, not sure if you want to do that now or separately.

samderanova:
> Could we also import the unit tests from [`test_sendgrid_handler.py`](https://github.com/HackAtUCI/HackUCI-Site/blob/main/tests/test_sendgrid_handler.py)? We'll also need the [`email_handler.py`](https://github.com/HackAtUCI/HackUCI-Site/blob/main/src/api/utils/email_handler.py) at some point, not sure if you want to do that now or separately.

Currently working on both 👍

apps/api/src/utils/email_handler.py
```diff
@@ -0,0 +1,43 @@
...
+class ContactInfo(Protocol):
+    email: EmailStr
+    first_name: str
+    last_name: str
+
+
+class Template(str, Enum):
+    # TODO: provide actual template IDs
+    CONFIRMATION_EMAIL = "d-2026cde7bebd45ad85723443808c5817"
+    GUEST_TOKEN = "d-b19f08e584cb4c0f97b55f567ee10afc"
```
taesungh, line 20: Note for later: confirming these still need to be changed?

apps/api/src/services/sendgrid_handler.py
samderanova: Either one is fine, it would probably be better if it was `log.debug` though.

apps/api/src/services/sendgrid_handler.py
```diff
@@ -0,0 +1,64 @@
...
+    try:
+        email_message = Mail()
+
+        if send_to_multiple:
+            if not isinstance(receiver_data, list):
+                raise TypeError(
+                    f"Expected {list} for receiver_data but got {type(receiver_data)}"
+                )
+            else:
+                for r in receiver_data:
```
taesungh, line 36: Issue: since we declared the parameter as type `Union[PersonalizationData, Iterable[PersonalizationData]]` but then assert instance of `list`, mypy coerces the type to `list[Any]` and the types of arguments to the `Email` constructor aren't checked.
While keeping `Iterable` is the most flexible approach, we may need to switch to `list` or use a typing overload to specify two versions based on `send_to_multiple`.

samderanova: Would we somehow have to adjust the stub so that it detects `send_to_multiple`?

apps/api/src/services/sendgrid_handler.py
```diff
@@ -0,0 +1,72 @@
...
+
+PersonalizationData = TypedDict(
+    "PersonalizationData",
+    {
+        "email": str,
+        "first_name": Optional[str],
+        "last_name": Optional[str],
+        "passphrase": Optional[str],
+    },
+)
```
taesungh, line 23: Chore: This is using an older syntax which has limited support. We can use the class-based syntax
We can also use the `NotRequired` qualifier instead of optional to indicate it does not need to be provided.
```python
class PersonalizationData(TypedDict):
    email: str
    first_name: NotRequired[str]
    ...
```

samderanova: Shoot, I was looking for something like this. I'll implement this.

apps/api/src/utils/PersonalizationData.py
taesungh, line 1: Thought: personally don't see any value in having a separate file for this as it displaces the type from the primary consumer. Think about in React where we leave component props interface in the same file.

samderanova: I see, that makes sense. In that case, I'll move it back to inside `sendgrid_handler.py`.

apps/api/src/services/sendgrid_handler.py
taesungh: Interestingly Pylance is smart enough and correctly narrows the generic to `list[PersonalizationData]`.
Honestly, I don't think we actually use the `send_to_multiple` option, so maybe we could get rid of it?

samderanova: Don't we use `send_to_multiple` to check whether it's one email or multiple?

apps/api/src/services/sendgrid_handler.py
```diff
@@ -0,0 +1,70 @@
...
+log = getLogger(__name__)
+
+SENDGRID_API_KEY = os.getenv("SENDGRID_API_KEY")
+
+
+class PersonalizationData(TypedDict):
+    email: str
+    first_name: NotRequired[str]
+    last_name: NotRequired[str]
+    passphrase: NotRequired[str]
```
taesungh, line 21: Suggestion: Given how this is template specific data, I think we can move the `Template` enum from `email_handler` to here, make `PersonalizationData` contain just the `email`, make subtypes for each template, and then overload `send_email` to use a specific data type corresponding to the template.

apps/api/src/services/sendgrid_handler.py
taesungh: In practice (in `email_handler`), we send only single emails, never multiple, and that was fine for last year. Not sure if we'll need this functionality this year though.

samderanova: We might need it for sending out acceptances

apps/api/src/services/sendgrid_handler.py
samderanova: What exactly do you mean by "subtypes"?

taesungh: We can extend a base version of the personalization type as needed
```python
class PersonalizationData(TypedDict):
    email: str

class ConfirmationPersonalization(PersonalizationData):
    first_name: str
    last_name: str

class GuestTokenPersonalization(PersonalizationData):
    passphrase: str
```

taesungh: Added in 703c09e

apps/api/src/services/sendgrid_handler.py
taesungh: Kept the option while also overloading in 703c09e. Still not perfect in narrowing but will work for now.

taesungh (approved):
Thanks for including stubs on this. Let's **squash** during merge.

apps/api/src/services/sendgrid_handler.py
```diff
@@ -0,0 +1,110 @@
...
+from sendgrid.helpers.mail import Email, Mail, Personalization
+
+log = getLogger(__name__)
+
+SENDGRID_API_KEY = os.getenv("SENDGRID_API_KEY")
+
+
+class Template(str, Enum):
+    # TODO: provide actual template IDs
+    CONFIRMATION_EMAIL = "d-2026cde7bebd45ad85723443808c5817"
```
njhuey, line 19: Should these be stored in env variables or is it fine to leave it hard-coded for now?

njhuey (approved):
lgmt

samderanova:
Please *do not* merge this PR yet, as I am testing template sending using the Hack group account. Previously SendGrid is tied to a separate Hack account, but for some reason, sending emails from this account is not working correctly.

njhuey (approved):
lgmt

taesungh (approved):
Rip account but I guess it's only reasonable when we're out of time.

apps/api/src/services/sendgrid_handler.py
samderanova: We can just leave them as is for now. Creating the templates is a one time thing, so if we do need to make any changes, we can simply update the template on SendGrid.

PR #74, opened 2023-12-12
Import Google Drive functions

samderanova:
This pull request imports the functions used to upload a file to Google Drive from last year's website.

The following environment variables are required:
- `RESUMES_FOLDER_ID`: the ID of the Google Drive folder to upload to
- Either `SERVICE_ACCOUNT_FILE`  or `GOOGLE_SERVICE_ACCOUNT_CREDENTIALS`: We use a [Google service acccount](https://cloud.google.com/iam/docs/service-account-overview) in tandem with [`aiogoogle`](https://aiogoogle.readthedocs.io/en/latest/) to automatically upload resumes when submitting a form. The keys are JSON that can either be stored in a file, in which case the path of the file should be stored in `SERVICE_ACCOUNT_FILE`, or be stored directly in `GOOGLE_SERVICE_ACCOUNT_CREDENTIALS`. For local development, it is recommended to take the `SERVICE_ACCOUNT_FILE` approach.


apps/api/src/utils/resume_handler.py
```diff
@@ -0,0 +1,52 @@
...
+from typing import Protocol
+
+from aiogoogle import HTTPError
+from fastapi import UploadFile
+
+from services import gdrive_handler
+
+log = getLogger(__name__)
+
+RESUMES_FOLDER_ID = os.getenv("RESUMES_FOLDER_ID")
```
taesungh, line 13: Could we add a note to the API README that we need this environment variable (as well as `GOOGLE_SERVICE_ACCOUNT_CREDENTIALS`) for deployment?

taesungh (approved):
Thanks for adding the stubs. Trusting things from last year will work as is.

rileysw (approved):

PR #76, opened 2023-12-12
feat: application pydantic models

samderanova:
This pull requests adds the Pydantic models related to application data. A few changes were made to the code:
- According to the application data, there are no word limits, so I removed `max_length` argument of the `Field` instances.
- In `utils.py`, two parameters/attributes were deprecated/no longer existing, so I updated them to their respective current ones.

### DIFF
--- a/apps/api/src/models/ApplicationData.py
+++ b/apps/api/src/models/ApplicationData.py
@@ -1,0 +1,45 @@
+from datetime import datetime
+from enum import Enum
+from typing import Optional
+
+from pydantic import BaseModel, EmailStr, Field, HttpUrl
+
+from .utils import form_body
+
+
+class Decision(str, Enum):
+    ACCEPTED = "ACCEPTED"
+    WAITLISTED = "WAITLISTED"
+    REJECTED = "REJECTED"
+
+
+Review = tuple[datetime, str, Decision]
+
+
+@form_body
+class RawApplicationData(BaseModel):
+    first_name: str
+    last_name: str
+    email: EmailStr
+    gender: str
+    pronouns: list[str]
+    ethnicity: str
+    is_18_older: bool
+    university: str
+    education_level: str
+    major: str
+    is_first_hackathon: bool
+    portfolio_link: Optional[HttpUrl]
+    linkedin_link: Optional[HttpUrl]
+    collaboration_question: Optional[str] = Field()
+    any_job_question: str = Field()
+
+    class Config:
+        anystr_strip_whitespace = True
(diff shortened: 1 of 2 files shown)


apps/api/src/models/ApplicationData.py
```diff
@@ -0,0 +1,44 @@
...
+    major: str
+    is_first_hackathon: bool
+    portfolio_link: Optional[HttpUrl]
+    linkedin_link: Optional[HttpUrl]
+    collaboration_question: Optional[str] = Field()
+    any_job_question: str = Field()
+
+    class Config:
+        anystr_strip_whitespace = True
+        max_anystr_length = 254
```
taesungh, line 38: Deprecation warning
> Support for class-based `config` is deprecated, use ConfigDict instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.5/migration/

taesungh, line 38: I think the max length in the config will override any individual properties although i'm not sure if specifying `Field` affects that. See HackAtUCI/HackUCI-Site#30 for the original discussion.

apps/api/src/models/ApplicationData.py
```diff
@@ -0,0 +1,44 @@
...
+    pronouns: list[str]
+    ethnicity: str
+    is_18_older: bool
+    university: str
+    education_level: str
+    major: str
+    is_first_hackathon: bool
+    portfolio_link: Optional[HttpUrl]
+    linkedin_link: Optional[HttpUrl]
+    collaboration_question: Optional[str] = Field()
```
taesungh, line 33: Thought: will the frontend specify null or empty string for these?
Suggestion: as mentioned before, I think we should follow FastAPI's suggestion and annotate with `Union[T, None]` instead of using `Optional`.

apps/api/src/models/ApplicationData.py
samderanova: I did some testing on my own machine with `RawApplicationData` and it turns out that the properties in `Field` override those in `ConfigDict`. For example, if the `Field` has `max_length = 10` and `ConfigDict` has `str_max_length = 254`, Pydantic will raise an exception if the provided string is greater than 10 characters.

taesungh (approved):
We can go with these for now and then figure out if we need to adjust `None` vs empty string when actually connecting the application form submission.

waalbert (approved):
LGTM

PR #80, opened 2023-12-13
feat: shadcn setup and schedule page

tyleryy:
# Overview
Setup [shadcn](https://ui.shadcn.com/) and built configurable schedule page from shadcn components.
# Ideas
- different color theme
- real time section for current event or next event that is going to happen. Placed in between timer and schedule. Changes in real time with some notification


waalbert:
The preview doesn't work for me

tyleryy:
> The preview doesn't work for me

The preview should work now I just fixed it. This was because I created an extra app folder from the `shadcn` CLI setup and it was overriding our actual app folder in `/src`. The CLI created the extra app folder because traditionally in a Next.js project you are not supposed to put the app folder in a src folder, and it was placing in a global.css file for the component styles. 

I moved shadcn generated global.css styles to our own global styles and now the styles are applied. However, I styled the components without any styles, so it now has a weird color scheme. I'm moving the branch to draft for now.

tyleryy:
The commit with the restyling includes code that sets up Sanity. Though not technically best practices and the PR is debatably too large. I feel we lose more trying to manually remove the Sanity Setup code. So I think we should review this PR with the Sanity changes. Argument can be made that Sanity setup is part of the page set up.

Issue #68, opened 2023-12-11
Set up guest authentication


samderanova:
As with last year, in order to allow applicants from other schools to log into their accounts, we should set up emails containing magic links will be sent to them.

This is another feature that can be imported from [last year's website](https://github.com/HackAtUCI/HackUCI-Site/blob/main/src/api/auth/guest_auth.py).

taesungh:
Backend added in #92, pending frontend guest login page in #89.

taesungh:
Added temporary functional login forms in #98, considering guest authentication API as complete.

PR #84, opened 2023-12-15
Set up user identity and JWT for native users

### LINKED ISSUE #69: Set up JWTs
As with [last year's website](https://github.com/HackAtUCI/HackUCI-Site/blob/main/src/api/auth/user_identity.py), we can rely on [JSON Web Tokens](https://jwt.io/introduction) for maintaining authentication status (including expiration).

taesungh:
Include the user identity code from last year to provide JWTs when completing SAML authentication.
The full guest authentication flow will come separately with #68, but the user identity foundation is provided here, e.g. the `GuestUser` model.

Closes #69.

## Changes
- Import user and user identity from last year
  - Import `user_identity` auth, `user` router, and `user_record` util, and associated tests from last year's [HackAtUCI/HackUCI-Site](https://github.com/HackAtUCI/HackUCI-Site)
  - Install [python-jose[cryptography]](https://github.com/mpdavis/python-jose)
  - Add user router to app
- Move `JWT_SECRET` check to `index.py`
- Resolve type errors w/ `NativeUser`, `EmailStr`
  - Ignore call-arg error when calling super constructor
  - `EmailStr` no longer needs to be explicitly constructed
- Resolve pydantic deprecation warnings
  - Replace `dict` with `model_dump` and `parse_obj` with `model_validate`
- Fix uid scoping when local part contains dot
- Incorporate user identity with SAML authentication
  - Provide user identity cookie when authenticating with SAML ACS
  - Rename `hackuci_auth` to `irvinehacks_auth`
- Improve type annotations in user/user_identity
  - Use `Annotated` for `Cookie` and `Union` instead of `Optional`
- Fix model issue with `IdentityResponse` for user
  - [`Optional` fields are no longer not required in Pydantic V2 models](https://docs.pydantic.dev/2.0/migration/#required-optional-and-nullable-fields)
  - Add unit test for `/user/me` endpoint
- Include note in API README about `JWT_KEY`
- Rename `log_out` to `logout` for consistency
- Add Pydantic mypy plugin to fix submodel call-arg

## Testing
Staging subdomain was manually assigned to the preview deployment. Confirmed SAML flow works and user identity cookie is provided which is properly consumed by `/api/user/me` endpoint.

### DIFF
--- a/apps/api/src/auth/user_identity.py
+++ b/apps/api/src/auth/user_identity.py
@@ -1,0 +1,156 @@
+import os
+from datetime import datetime, timedelta, timezone
+from typing import Any, Optional
+
+from fastapi import Cookie, HTTPException, Response, status
+from fastapi.testclient import TestClient
+from jose import JWTError, jwt
+from pydantic import BaseModel, EmailStr
+
+JWT_ALGORITHM = "HS256"
+JWT_SECRET = os.getenv("JWT_SECRET", "")
+
+
+class User(BaseModel):
+    uid: str
+    email: EmailStr
+
+    def __str__(self) -> str:
+        return self.uid
+
+
+class NativeUser(User):
+    ucinetid: str
+    display_name: str
+    affiliations: list[str]
+
+    def __init__(
+        self, *, ucinetid: str, display_name: str, email: str, affiliations: list[str]
+    ):
+        uid = f"edu.uci.{ucinetid}"
+        super().__init__(
+            uid=uid,
+            ucinetid=ucinetid,
+            display_name=display_name,
+            email=email,
+            affiliations=affiliations,
+        )
+
+
+class GuestUser(User):
+    def __init__(self, *, email: str):
+        uid = scoped_uid(EmailStr(email))
+        super().__init__(uid=uid, email=email)
+
+
+class UserTestClient(TestClient):
+    """Provide a test client that sends requests on behalf of the given user."""
+
+    def __init__(self, user: User, *args: Any, **kwargs: Any):
+        kwargs["cookies"] = kwargs.get("cookies", dict())
+        kwargs["cookies"]["hackuci_auth"] = _generate_jwt_token(user)
+        super().__init__(*args, **kwargs)
+
+
+class JWTClaims(BaseModel):
+    iat: datetime
+    exp: datetime
+    sub: str
+    data: dict[str, Any]
+
+
+def scoped_uid(email: EmailStr) -> str:
+    """Provide a scoped unique identifier based on the email domain."""
+    if uci_email(email):
+        raise ValueError("UCI user should use NativeUser")
+    local, domain = email.split("@")
+    reversed_domains = ".".join(reversed(domain.split(".")))
+    return f"{reversed_domains}.{local}"
+
+
+def uci_email(email: EmailStr) -> bool:
+    """Check whether or not an email address is part of UCI or a subdomain thereof."""
+    return email.endswith("@uci.edu") or email.endswith(".uci.edu")
+
+
+def utc_now() -> datetime:
+    """Return current datetime with proper UTC timezone."""
+    return datetime.now(timezone.utc)
+
+
+def remove_user_identity(response: Response) -> Response:
+    """Remove authentication cookie."""
+    response.set_cookie("hackuci_auth", "", max_age=0)
+    return response
+
+
+def issue_user_identity(user: User, response: Response) -> Response:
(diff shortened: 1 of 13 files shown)


apps/api/src/auth/user_identity.py
```diff
@@ -0,0 +1,157 @@
...
+        self, *, ucinetid: str, display_name: str, email: str, affiliations: list[str]
+    ):
+        uid = f"edu.uci.{ucinetid}"
+        super().__init__(  # type: ignore[call-arg]
+            uid=uid,
+            ucinetid=ucinetid,
+            display_name=display_name,
+            email=email,
+            affiliations=affiliations,
+        )
```
samderanova, line 39: Question: Is the reason why you ignored here because the additional arguments are needed to pass into the `BaseModel` constructor (in order to generate this model properly)?

apps/api/src/routers/user.py
```diff
@@ -0,0 +1,54 @@
...
+async def login(email: EmailStr = Form()) -> RedirectResponse:
+    log.info("%s requested to log in", email)
+    if user_identity.uci_email(email):
+        # redirect user to UCI SSO login endpoint, changing to GET method
+        return RedirectResponse("/api/saml/login", status.HTTP_303_SEE_OTHER)
+    return RedirectResponse("/api/guest/login", status.HTTP_307_TEMPORARY_REDIRECT)
+
+
+@router.get("/logout")
+async def log_out() -> RedirectResponse:
```
samderanova, line 35: Nitpick: One word `logout` to be consistent with the route name?

samderanova (commented):
Good work on this! Thankfully, stubs exist for `python-jose` already.

samderanova (changes requested):
A few nitpicks addressed above:

apps/api/src/routers/user.py
taesungh: I initially chose two words to make the function name a verb, but sure, consistency might be better here.

apps/api/src/auth/user_identity.py
taesungh: mypy didn't complain about this last year, but as of mypy v1.1.1, possibly because of the added support for `dataclass_transform`, we'll get error messages about the extra keyword arguments
> error: Unexpected keyword argument `ucinetid` for `__init__` of `User`  [[call-arg](https://mypy.readthedocs.io/en/latest/_refs.html#code-call-arg)]

Pydantic will raise other errors if I omit the extra attributes and try to assign them separately. There probably is a better way to do this, but I couldn't find such.

taesungh: Although not perfect, it seems using the [mypy plugin for pydantic](https://docs.pydantic.dev/latest/integrations/mypy/) _without_ [`init_forbid_extra`](https://docs.pydantic.dev/latest/integrations/mypy/#generate-a-signature-for-model__init__) will add kwargs as `Any` to `super().__init__`
> Revealed type is "def (*, uid: builtins.str, email: builtins.str, **kwargs: Any)"

apps/api/src/routers/user.py
taesungh: Changed in 7491f2e.

apps/api/src/auth/user_identity.py
samderanova: Awesome, thanks for finding a workaround for this.

apps/api/src/utils/user_record.py
```diff
@@ -0,0 +1,20 @@
+from enum import Enum
+
+from pydantic import Field
+
+from services.mongodb_handler import BaseRecord
+
+
+class Role(str, Enum):
```
rileysw, line 8: Remind me again, how did we assign mentor/volunteer roles last year? Sponsors and judges should also be able to view the schedule.

rileysw (approved):

apps/api/src/utils/user_record.py
taesungh: I don't think we ended up using all of these roles last year, but the concept was mentors and volunteers could also log into the portal and see personalized information, e.g. about their shifts or other event information they might want to know. To actually assign roles, we would need to do it manually in the database or implement a user management interface in the admin portal.

For the schedule, we made it public last year (HackAtUCI/HackUCI-Site@55fa9a2), not sure if you want to handle that differently this year.

rileysw: Got it, yeah public schedule is fine.

samderanova (approved):

samderanova:
Please *squash* merge.

PR #85, opened 2023-12-15
feat: portal page

njhuey:
WIP commit. 
TODO: incorporate user data to alter what is shown based on user status.


apps/site/src/app/portal/Portal.tsx
```diff
@@ -1,21 +1,62 @@
...
+	useEffect(() => {
+		if (status === null) {
+			router.push("/apply");
+		}
+	}, [status, router]);
+
+	if (status === null) {
+		return null;
+	}
+
```
samderanova, line 39: Nitpick: We can use [`redirect`](https://nextjs.org/docs/app/api-reference/functions/redirect) instead of `useRouter` to maintain SSR.

apps/site/src/app/portal/Portal.tsx
```diff
@@ -1,21 +1,62 @@
...
-					{key} - {value}
-				</p>
-			))}
-		</div>
+		<>
+			<section className=" w-full flex items-center flex-col">
+				<div className="m-24">
+					<Head>
+						<title>Portal | IrvineHacks 2024</title>
+					</Head>
```
samderanova, line 46: Nitpick: Instead of having the `title` here, you can [ `export const metadata`](https://nextjs.org/docs/app/building-your-application/optimizing/metadata) in `page.tsx` for `portal` like how it's done in [`layout.tsx`](https://github.com/HackAtUCI/irvinehacks-site-2024/blob/main/apps/site/src/app/layout.tsx).

samderanova (changes requested):
Thanks for taking the time to work on this! A few small nitpicks to address:

apps/site/src/app/portal/Portal.tsx
njhuey: useEffect requires the component to be a client component. I will still switch over to using redirect

apps/site/src/app/portal/Portal.tsx
```diff
@@ -1,21 +1,62 @@
-import api from "@/lib/utils/api";
+"use client";
```
samderanova, line 1: Issue: I should've said this in my previous comment, but now that you're using `redirect`, you don't need "use client" or `useEffect`. You can directly check if `status` is `null` and redirect based off of that.

apps/site/src/app/portal/Portal.tsx
```diff
@@ -1,21 +1,62 @@
...
+
+	useEffect(() => {
+		if (status === null) {
+			redirect("/apply");
+		}
+	}, [status]);
+
+	if (status === null) {
+		return null;
+	}
```
samderanova, line 41: Issue: You shouldn't need this part since you already check it above.

samderanova (changes requested):

apps/site/src/app/portal/Portal.tsx
```diff
@@ -1,21 +1,50 @@
...
+	reviewed = "REVIEWED",
+	accepted = "ACCEPTED",
+	rejected = "REJECTED",
+	waitlisted = "WAITLISTED",
+	confirmed = "CONFIRMED",
 }
 
+export const metadata: Metadata = {
+	title: "Portal | IrvineHacks 2024",
+};
```
taesungh, line 21: Chore/Issue: the metadata needs to be exported from the `page.tsx` instead of the component. See [`app/login/page.tsx`](https://github.com/HackAtUCI/irvinehacks-site-2024/blob/main/apps/site/src/app/login/page.tsx) for an example.

apps/site/src/app/portal/Portal.tsx
```diff
@@ -1,21 +1,50 @@
...
+					<h1
+						className={`${styles.title} font-display sm:text-[3rem] text-[#fffce2] text-6xl text-center`}
+					>
+						Portal
+					</h1>
+				</div>
+				<div className="bg-white text-black max-w-4xl rounded-2xl p-6 flex flex-col mb-24 w-full">
+					<h2 className="text-4xl font-semibold">Status</h2>
+					<VerticalTimeline status={status} />
+					<Message status={status as PortalStatus} />
```
taesungh, line 44: Question: is there a particular reason the cast `as PortalStatus` was needed?

apps/site/src/app/portal/VerticalTimeline.tsx
```diff
@@ -0,0 +1,89 @@
+import Image from "next/image";
+
+import BorderCircle from "@/assets/icons/border-circle.svg";
+import CheckCircle from "@/assets/icons/check-circle-fill.svg";
+import XCircle from "@/assets/icons/x-circle-fill.svg";
+
+import { PortalStatus } from "./Portal";
+
+interface VerticalTimelineProps {
+	status: string;
```
taesungh, line 10: Chore: `status` could be typed as `PortalStatus`

taesungh (commented):
Nice work on this, thanks for figuring out how to work with the new `getUserIdentity` function. One issue about page metadata and smaller notes about type usage.

apps/site/src/app/portal/Portal.tsx
njhuey: I believe it is to avoid a typescript error: 
`Type 'string' is not assignable to type 'PortalStatus'.ts(2322)`

apps/site/src/app/portal/Message.tsx
```diff
@@ -0,0 +1,36 @@
...
+				continue developing your skills and passion for technology. We
+				would love to see you apply again next year!
+			</p>
+		),
+		[PortalStatus.waitlisted]: <></>,
+		[PortalStatus.accepted]: <></>,
+		[PortalStatus.confirmed]: <></>,
+	};
+
+	return <div className="">{messages[status]}</div>;
```
samderanova, line 33: Nitpick [non-blocking]: empty className?

apps/site/src/assets/icons/check-circle-fill.svg
samderanova, line 1: Suggestion for later: I suppose it wouldn't really matter too much at this moment, but we could look into using icons from `lucide-react` instead of these SVGs.

samderanova (approved):
Thanks for getting this done so quickly!

samderanova (approved):

apps/site/src/app/portal/Portal.tsx
taesungh: Got it, we can revisit this: think we could have identity itself type the status.

taesungh (approved):
Works out. Thanks for figuring out how to work with the new server-side identity. Please update the PR description before merging.

Issue #87, opened 2023-12-16
Stop assigning production domain to deployments


taesungh:
As we are pulling the final things together before release, we don't want to prematurely deploy our site to our production domain for people to see things are still being worked on/are in a non-functioning state.

taesungh:
We can revert the change once we are ready to officially release applications for the year.

PR #88, opened 2023-12-16
Temporarily disable production deployment flag

taesungh:
As part of #87
- Until we finish the application page and system, disable deployments to the production domain
- We will also need to manually assign the staging domain because Vercel still has not resolved our domain ownership issue


samderanova (approved):
Thanks for the quick change!

Issue #94, opened 2023-12-16
Frontend: Remove default values for multiselect in application form


taesungh:
On the application form, for **Current Education Level**, **School Name**, and **What is your major?**, the default value should be unselected instead of the first option.

PR #95, opened 2023-12-16
Remove default values from application dropdowns

rileysw:
Address #94


samderanova (approved):

taesungh (approved):
Nice and simple, thanks for adding.

PR #96, opened 2023-12-16
feat: UserContext implementation

samderanova:
Sets up the [UserContext](https://github.com/HackAtUCI/HackUCI-Site/blob/main/src/utils/userContext.tsx) from last year's website. Note that because the `Navbar` relies on this context, we need to use this context in `layout.tsx`, which essentially converts everything to a client component.


taesungh (commented):
As I mentioned over call, it might not be terrible to run the server-side API call from the components that need them since it's looking like it will be just the Navbar and the currently active page. There is a slight disadvantage with prop-drilling, but the levels of children should not be that excessive.

rileysw (approved):

samderanova:
Closing this PR as we have decided to do SSR API calls as Taesung mentioned.

PR #97, opened 2023-12-16
feat: enable application submission

### LINKED ISSUE #90: Enable application submission to backend route
Currently, submitting an application does not properly send it to the backend. The form's `action` should be set to `/api/user/apply`.

samderanova:
This does not entirely resolve #90 but takes a step towards doing so using native POST (instead of `axios` POST which we can use for error handling).


taesungh (commented):
This will work for now. However keep in mind that it would be better to have client-side error handling in case anything goes wrong during the form submission in which case we would need to use `event.preventDefault` and then send the POST request ourselves with axios or alike.

taesungh (approved):
Approving to unblock as submission should still work, can revisit the client-side error handling after we have the other pages working.

Issue #91, opened 2023-12-16
Determine how to pass user data through components


samderanova:
Last year, we used [`useContext` ](https://github.com/HackAtUCI/HackUCI-Site/blob/main/src/utils/userContext.tsx) in order to pass user data throughout components. However, since we are now using NextJS App Router, using `useContext` would force us to add `"use client"` at the top of the majority of our pages, eliminating the purpose of server side rendering.

Currently, prop drilling from `layout.tsx` seems to be the only reasonable alternative (cc. @taesungh). We could also have each page query the backend for the information, but this increases the number of requests to the backend by a large amount.

samderanova:
Testing out [this library](https://github.com/manvalls/server-only-context#readme) to see if it works.

samderanova:
Since we are short on time, we should rely on `useContext` as with last year.

samderanova:
As mentioned in #96, we'll be doing server side API calls instead of using `useContext` because using `useContext` would require us to add `"use client"`. We ideally want to leverage the benefits of SSR as much as possible.

PR #100, opened 2023-12-16
Setup/frontend user identity

samderanova:
Address #91. Set up a utility function to get the user's identity and use prop drilling like what we did with Sanity requests to send the identity as a prop to the navbar component since the navbar is a client component.

### DIFF
--- a/apps/site/src/lib/components/Button/Button.tsx
+++ b/apps/site/src/lib/components/Button/Button.tsx
@@ -17,7 +17,7 @@ const Button: React.FC<ButtonProps> = ({
 	href,
 	className,
 	isLightVersion,
-	usePrefetch
+	usePrefetch,
 }) => {
 	if (href) {
 		return (
--- a/apps/site/src/lib/components/Button/Button.tsx
+++ b/apps/site/src/lib/components/Button/Button.tsx
@@ -6,7 +6,7 @@ import clsx from "clsx";
 interface ButtonProps {
 	text: string;
 	className?: string;
-	href: string;
+	href?: ComponentProps<typeof Link>["href"];
 	isLightVersion?: boolean;
 	usePrefetch?: boolean;
 }
@@ -20,10 +20,11 @@ const Button: React.FC<ButtonProps> = ({
 }) => {
 	if (href) {
 		return (
-			<a
+			<Link
 				href={href}
 				target={isLightVersion ? "" : "_blank"}
 				rel="noopener noreferrer"
+				prefetch={usePrefetch}
 				className={clsx(
 					styles.button,
 					isLightVersion && styles.lightButton,
@@ -32,7 +33,7 @@ const Button: React.FC<ButtonProps> = ({
 				)}
 			>
 				{text}
-			</a>
+			</Link>
 		);
 	}
 	return (
--- a/apps/site/src/lib/components/Button/Button.tsx
+++ b/apps/site/src/lib/components/Button/Button.tsx
@@ -2,12 +2,14 @@ import type { ComponentProps } from "react";
 import Link from "next/link";
 import styles from "./Button.module.css";
 import clsx from "clsx";
+import { disableInstantTransitions } from "framer-motion";
 
 interface ButtonProps {
 	text: string;
 	className?: string;
 	href?: ComponentProps<typeof Link>["href"];
 	isLightVersion?: boolean;
+	usePrefetch?: boolean;
 }
 
 const Button: React.FC<ButtonProps> = ({
@@ -15,6 +17,7 @@ const Button: React.FC<ButtonProps> = ({
 	href,
 	className,
 	isLightVersion,
+	usePrefetch
 }) => {
 	if (href) {
 		return (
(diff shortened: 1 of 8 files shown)


apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -66,7 +71,19 @@ function Navbar() {
...
+					{!status && <NavLinkItem href="/apply">Apply</NavLinkItem>}
+					{status !== null && (
+						<NavLinkItem href="/portal">Portal</NavLinkItem>
+					)}
+					{uid === null ? (
+						<Button text="Login" href="/login" isLightVersion />
+					) : (
+						<Button
+							text="Logout"
+							href="/api/user/logout"
```
taesungh, line 83: Issue: `Button` uses next/link under the hood which will [prefetch the link](https://nextjs.org/docs/app/api-reference/components/link#prefetch) causing the browser to load the logout page.

taesungh: Could we also add a rewrite or separate handler for `/logout` so the end user won't see the API path on link hover?

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -66,7 +71,19 @@ function Navbar() {
 					<NavLinkItem href="/">Schedule</NavLinkItem>
 					<NavLinkItem href="/">Resources</NavLinkItem>
 					<NavLinkItem href="/">Stage</NavLinkItem> */}
-					<Button text="Login" href="/login" isLightVersion />
+					{!status && <NavLinkItem href="/apply">Apply</NavLinkItem>}
+					{status !== null && (
+						<NavLinkItem href="/portal">Portal</NavLinkItem>
+					)}
+					{uid === null ? (
```
taesungh, line 78: Nitpick [non-blocking]: would prefer naming this as `isLoggedIn` or alike so the intent is clear.

apps/site/src/lib/components/Navbar/NavbarParent.tsx
```diff
@@ -0,0 +1,13 @@
...
+export interface Identity {
+    uid: string | null;
+    role: string | null;
+    status: string | null;
+}
+
+export default async function NavbarParent() {
+    const identity = await getUserIdentity();
+
+    return <><Navbar identity={identity} /></>
```
taesungh, line 7: Nitpick: the fragment is unnecessary since there is only one child.

apps/site/src/lib/components/Button/Button.tsx
```diff
@@ -2,26 +2,30 @@ import type { ComponentProps } from "react";
...
 	isLightVersion?: boolean;
+	usePrefetch?: boolean;
 }
 
 const Button: React.FC<ButtonProps> = ({
 	text,
 	href,
 	className,
 	isLightVersion,
+	usePrefetch
```
taesungh, line 20: Note [out of scope]: we could extend the `InternalLinkProps` and pass `...rest` so we don't have to individually specify link props we need.

apps/site/src/lib/components/Button/Button.tsx
```diff
@@ -2,26 +2,30 @@ import type { ComponentProps } from "react";
 import Link from "next/link";
 import styles from "./Button.module.css";
 import clsx from "clsx";
+import { disableInstantTransitions } from "framer-motion";
```
taesungh, line 5: Chore: Stray import?

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -1,19 +1,25 @@
 "use client";
 
+import Image from "next/image";
+import React, { useState, useEffect } from "react";
```
taesungh, line 4: Chore: stray `React` import?

apps/site/src/lib/components/Navbar/Navbar.tsx
taesungh: Think the Next server-side redirect or rewrite could also fix the no-refresh issue.

apps/site/src/lib/utils/getUserIdentity.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export interface Identity {
+	uid: string | null;
+	role: string | null;
+	status: string | null;
+}
+
+export default async function getUserIdentity() {
+	const identity = await api.get<Identity>("/user/me");
+	return identity.data;
```
taesungh, line 11: Thought [non-blocking]: if the fetch fails, does the entire site render crash?

apps/site/src/lib/components/Navbar/Navbar.tsx
samderanova: Got it to work, but I had to sacrifice `next/link` in the process.

taesungh (approved):
Thanks for finding a way to make this work. Approving to unblock, we can revisit later if needed.

njhuey (approved):
lgmt

PR #102, opened 2023-12-16
Change "Stay Updated" button to "Apply"

rileysw:
Home page button says "Apply" and links to Apply page

### DIFF
--- a/apps/site/src/lib/components/Button/Button.tsx
+++ b/apps/site/src/lib/components/Button/Button.tsx
@@ -20,8 +20,6 @@ const Button: React.FC<ButtonProps> = ({
 		return (
 			<Link
 				href={href}
-				target={isLightVersion ? "" : "_blank"}
-				rel="noopener noreferrer"
 				className={clsx(
 					styles.button,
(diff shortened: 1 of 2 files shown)


apps/site/src/lib/components/Button/Button.tsx
```diff
@@ -20,8 +20,6 @@ const Button: React.FC<ButtonProps> = ({
 		return (
 			<Link
 				href={href}
-				target={isLightVersion ? "" : "_blank"}
```
rileysw, line 23: I removed this so a new tab doesn't open when clicking button link.

This affects other pages that use the same Button component. However, imo none of them should open new tabs. If we need that functionality for another button, we can implement a condition by modifying the interface.

The affected buttons are below:

Home Page - Apply button
Home Page - Mentor and Volunteer buttons
Application Page - Submission button
Navbar - Login button

IanWearsHat (approved):
LGTM

samderanova (approved):

samderanova:
Thanks for working on this! Please do a *regular* merge.

Issue #93, opened 2023-12-16
Add API endpoint to accept hacker application


samderanova:
The endpoint `/api/user/apply` needs to be set up in order to process applications properly. This should be a simple copy and paste from [last year's website](https://github.com/HackAtUCI/HackUCI-Site/blob/0b181fb750094373e2ab3f689d2691312bb88ee7/src/api/routers/user.py#L62).

taesungh:
See PR discussion about lingering submission issues, also tracking in #90.

samderanova:
As mentioned in #90, we need to add additional code in the backend to detect whether a user's application contains the value "Other" for one of the fields (meaning the user has disabled JS). If so, we should return some 400 status code.

PR #103, opened 2023-12-16
Added resume field in Application Page

IanWearsHat:
## Validation
- Form will prevent submission if:
  - File is not pdf
  - File is over 0.5 MB
## Feedback
- If any of the errors above occur, text will appear saying an error as occurred
- If successful, text will appear saying so
## Form Change
- Changed form to be multipart to allow a file to be sent

### DIFF
--- a/apps/site/src/app/apply/sections/Form/AgeInformation.tsx
+++ b/apps/site/src/app/apply/sections/Form/AgeInformation.tsx
@@ -30,7 +30,7 @@ export default function AgeInformation() {
 			</div>
 
 			<SimpleRadio
-				name="minor-check"
+				name="minor_check"
 				values={yesNoOptions}
 				title="Will you be 18 years or older by January 26th, 2024?"
 				titleClass="text-xl font-bold m-0 text-center"
--- a/apps/site/src/app/apply/sections/Form/Form.tsx
+++ b/apps/site/src/app/apply/sections/Form/Form.tsx
@@ -38,6 +38,9 @@ export default function Form() {
 			/>
 			<form
 				className={`${styles.form} text-[#000000] w-8/12 flex flex-col items-center py-12 gap-10 z-1 max-[800px]:w-9/12 max-[400px]:w-11/12`}
+				action="/api/bruh"
+				method="post"
+				encType="multipart/form-data"
 			>
 				<BasicInformation />
 				<SchoolInformation />
--- a/apps/site/src/app/apply/sections/Form/Form.tsx
+++ b/apps/site/src/app/apply/sections/Form/Form.tsx
@@ -39,6 +39,7 @@ export default function Form() {
 			<form
 				method="post"
 				className={`${styles.form} text-[#000000] w-8/12 flex flex-col items-center py-12 gap-10 z-1 max-[800px]:w-9/12 max-[400px]:w-11/12`}
+				action="/api/user/apply"
 				encType="multipart/form-data"
 			>
 				<BasicInformation />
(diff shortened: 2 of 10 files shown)


apps/site/src/app/apply/sections/Form/AgeInformation.tsx
```diff
@@ -30,7 +30,7 @@ export default function AgeInformation() {
 			</div>
 
 			<SimpleRadio
-				name="minor-check"
+				name="minor_check"
```
samderanova, line 33: Question: is there a reason why you changed these to underscores? If I remember correctly, @taesungh is working on writing the code to process the application, so you might want to sync with him before changing the names.

apps/site/src/app/apply/sections/Form/ResumeOutputFeedback.tsx
```diff
@@ -0,0 +1,20 @@
+interface outputFeedbackProps {
```
samderanova, line 1: Nitpick: We generally use TitleCase for interfaces.

apps/site/src/app/apply/sections/Form/ResumeOutputFeedback.tsx
```diff
@@ -0,0 +1,20 @@
...
+	errorMessage: string;
+	resumePath: string;
+}
+
+export default function OutputFeedBack(props: outputFeedbackProps) {
```
samderanova, line 6: Suggestion: you can extract the properties out from the props like so: `OutputFeedback({errorMessage, resumePath}: OutputFeedbackProps)`

samderanova (changes requested):
Great work on this! A few questions/nitpicks:

apps/site/src/app/apply/sections/Form/AgeInformation.tsx
taesungh: Yes, I will be updating this field name along with others as part of #93. I'm not sure if we should wait for my PR or if I should try to work with these new changes.

samderanova: If you haven't gotten too far with #93 yet (i.e. written the field names), we can merge this one first.

IanWearsHat: I was testing to see if I could send the form in FastAPI (just a simple print statement), but for some reason having a hyphen in the parameters of the post request I was sending raised a compiler error, so I changed them all to underscores. 

Sorry, I should have thought about other people working with these names.

samderanova: No, that makes sense. Python can't process hyphens in names so I think that's why you're receiving those errors.

samderanova (dismissed):

apps/site/src/app/apply/sections/Form/Form.tsx
```diff
@@ -38,11 +39,12 @@ export default function Form() {
 			<form
 				method="post"
 				className={`${styles.form} text-[#000000] w-8/12 flex flex-col items-center py-12 gap-10 z-1 max-[800px]:w-9/12 max-[400px]:w-11/12`}
-				action="/api/user/apply"
+				encType="multipart/form-data"
```
taesungh, line 42: Chore: I think we still need the action endpoint here?

apps/site/src/app/apply/sections/Form/AgeInformation.tsx
taesungh: No worries, this should probably have been noticed the first time around. I'm almost ready with the apply endpoint, but the file upload is also an important part I should test, so I can rebase after this is merged.

apps/site/src/app/apply/sections/Form/Form.tsx
IanWearsHat: Oops, I thought I left it on merge

samderanova (approved):

taesungh (approved):
Approving to unblock #93, we can revisit the actual design: currently not a fan of the "Successfully uploaded" text when it actually hasn't been uploaded.

PR #104, opened 2023-12-16
Added reply to parameter to email handler

IanWearsHat:


apps/api/src/utils/email_handler.py
```diff
@@ -6,6 +6,7 @@
 from services.sendgrid_handler import Template
 
 IH_SENDER = ("apply@irvinehacks.com", "IrvineHacks 2024 Applications")
+REPLY_TO_HACK_AT_UCI = ("hack@uci.edu", "Hack at UCI")
```
samderanova, line 9: Issue: The `reply_to` parameter needs to be added to the `send_email` function and the `reply_to` attribute of the Mail object in `send_email` needs to be set to the provided value.

samderanova: This will result in some type errors, but I can take care of those.

samderanova (changes requested):
Nice start on this. One important thing to address:

apps/api/src/utils/email_handler.py
samderanova: In the meantime, can you add `reply_to: Tuple[str, str]` as an argument to all of the overloads?

apps/api/src/services/sendgrid_handler.py
```diff
@@ -36,6 +36,7 @@ class GuestTokenPersonalization(PersonalizationData):
 async def send_email(
     template_id: Literal[Template.CONFIRMATION_EMAIL],
     sender_email: Tuple[str, str],
+    reply_to: Tuple[str, str],
```
samderanova, line 39: Issue [repeating]: We'll want to place `reply_to` specifically after `receiver_data` since that's how we order the arguments in our function calls.

samderanova (commented):
Looking good, one small issue:

samderanova (approved):

taesungh (approved):
Thanks for adding this. Nice work figuring out how to work with the typing overloads as well as the unit tests.

PR #107, opened 2023-12-17
Run Sanity deploy preview only when PR is labeled

### LINKED ISSUE #106: Limit preview deployments for Sanity
We recently hit Vercel's hobby tier limit for deployments per day. To reduce our usage, we could limit when a preview deployment is made for the Sanity project by requiring a certain label on the pull request.

taesungh:
Resolves #106 since we are limited by Vercel's hobby tier limit on deployments per day. This will effectively reduce our usage by half.
- Run the Sanity preview deployment workflow only when the pull request is labeled with 'sanity'


samderanova (approved):
Thanks for working on this. It'll definitely decrease the amount of deployments we make.

PR #110, opened 2023-12-17
Setup/application preface

IanWearsHat:
Preface is shown before actual application.

Clicking button will add query parameter `/apply?prefaceAccepted=true` which will allow the application component to render


apps/site/src/app/apply/sections/ApplyConfirmation/ConfirmationDetails.tsx
```diff
@@ -0,0 +1,32 @@
...
+				be over the age of 18 by February 3rd, 2023.
+			</p>
+			<p className={`${styles.policy} text-4xl`}>
+				<strong>
+					Applications are due on January 14th, 2024 at 11:59PM PST.
+				</strong>
+			</p>
+			<Button
+				text="Proceed to Application"
+				href="/apply?prefaceAccepted=true"
```
samderanova, line 28: Issue: We'll want to redirect applicants to `/login` instead of this route if they aren't logged in. You can check this with the `getUserIdentity` function we made in `apps/site/src/lib/utils/getUserIdentity.ts`.

apps/site/src/app/apply/sections/ApplyConfirmation/ConfirmationDetails.tsx
```diff
@@ -0,0 +1,32 @@
...
+			/>
+		</div>
+	);
```
samderanova, line 31: Issue: Please also advertise applications for mentors and volunteers, as we did [last year](https://github.com/HackAtUCI/HackUCI-Site/blob/main/src/views/apply/components/ApplicationPreface.tsx).

samderanova (changes requested):
Great start so far! A few things to address:

apps/site/src/app/apply/sections/ApplyConfirmation/ConfirmationDetails.tsx
IanWearsHat: Do you think the button should change text if not logged in?

apps/site/src/app/apply/page.tsx
```diff
@@ -3,44 +3,60 @@ import Image from "next/image";
...
+export default function Apply({
+	searchParams,
+}: {
+	searchParams?: {
+		prefaceAccepted?: string;
+	};
+}) {
+	const applyBody =
+		searchParams !== undefined &&
+		searchParams?.prefaceAccepted == "true" ? (
```
samderanova, line 23: Nitpick: We generally prefer using strict equality (`===`).

apps/site/src/app/apply/sections/ApplyConfirmation/ConfirmationDetails.tsx
```diff
@@ -0,0 +1,49 @@
...
+				States and will be over the age of 18 by January 26th, 2024.
+			</p>
+			<strong className="text-lg text-[#FF2222]">
+				Applications are due on January 14th, 2024 at 11:59PM PST.
+			</strong>
+			<Button
+				text={isLoggedIn ? "Proceed to Application" : "Log in to Apply"}
+				href={isLoggedIn ? "/apply?prefaceAccepted=true" : "/login"}
+			/>
+			<hr className="mt-5" />
```
samderanova, line 35: Issue: The `hr` doesn't appear (at least on my screen). You could set `width: 100%` and then give it a height of some small amount of pixels.

image

samderanova (changes requested):
A few more small things, but other than that, looks good to me!

samderanova (commented):
Nice work! Please *squash* merge.

samderanova (approved):

taesungh (approved):
Nicely done. The **Proceed to Application** should ideally have seamless navigation, but I think #120 will fix that.

Issue #111, opened 2023-12-17
Improve resume upload field


taesungh:
The resume field on the application form is currently required. Additionally, following up from #103, the validation text is a little deceptive by saying "successfully uploaded" when the file has in fact not been actually uploaded.
- Make the resume file input not required
- Update the validation text to provide a more reasonable message, e.g. "Selected <filename>"
  - It might also be helpful to change the icon once selected to a document icon

PR #113, opened 2023-12-17
Add `/user/apply` API endpoint to receive hacker applications

### LINKED ISSUE #93: Add API endpoint to accept hacker application

taesungh:
To accept hacker applications, the API needs to process the application form data. The `/user/apply` endpoint was imported mostly as is with minor improvements (see commit history for changes after import) to typing and one change about acquiring the applicant's email.

Lingering issues:
- Form fields with "Other" options are currently not received
  - As noted in #90, this can be handled with client-side submission, but ideally the API also takes care of it
- The resume upload is currently required (see #111)
  - When no file upload is included, the browser will still send an empty octet-stream in the native post request
  - Same note as above about API handling vs client-side submission

Closes #93.

## Changes
- Import last year's user apply endpoint/tests as is
  - Import `/apply` route in user router, user record models, and tests from last year's [HackAtUCI/HackUCI-Site](https://github.com/HackAtUCI/HackUCI-Site) repository
- Fix user apply types, tests, and models
  - [`Optional` values no longer default to `None` in Pydantic V2](https://docs.pydantic.dev/2.0/migration/#required-optional-and-nullable-fields)
  - Replace model `dict` with `model_dump`
  - Update sample fields in unit test based on changes to application
  - Fix other type errors
    - Specify type for `Applicant.Role`
    - Use `Status` Enum value in expected `Applicant` object
    - Ignore arg-type error when constructing sample application data
- Add `HttpUrl` type to resume upload
- Fix application form field name correspondence
  - Update application form field names to match between frontend and API
- Fix URL serialization from Pydantic to Mongo BSON
  - Include `field_serializer` in `ProcessedApplicationData` to convert `HttpUrl`s to `str`
  - Add unit test to confirm application data can be BSON-serialized
- Use user email from identity for confirmation
  - Last year's application form had a redundant email field which was removed this year
  - Use the email from the user identity for sending the application confirmation email
- Use `Annotated` syntax for /apply endpoint params

## Testing
1. Temporarily disable the resume upload and email handler by immediately calling `return`
2. Load the Docker environment for the API
3. Open the site with the application form on `/apply`
4. Complete the application with standard (non-other) values and a resume file
5. Observe the application is properly received by the API

### DIFF
--- a/apps/api/src/models/ApplicationData.py
+++ b/apps/api/src/models/ApplicationData.py
@@ -2,7 +2,7 @@ from datetime import datetime
 from enum import Enum
 from typing import Union
 
-from pydantic import BaseModel, ConfigDict, Field, HttpUrl
+from pydantic import BaseModel, ConfigDict, Field, HttpUrl, field_serializer
 
 from .utils import form_body
 
@@ -39,3 +39,9 @@ class ProcessedApplicationData(RawApplicationData):
     resume_url: Union[HttpUrl, None] = None
     submission_time: datetime
     reviews: list[Review] = []
+
+    @field_serializer("linkedin", "portfolio", "resume_url")
+    def url2str(self, val: Union[HttpUrl, None]) -> Union[str, None]:
+        if val is not None:
+            return str(val)
+        return val
--- a/apps/api/src/models/ApplicationData.py
+++ b/apps/api/src/models/ApplicationData.py
@@ -30,13 +30,13 @@ class RawApplicationData(BaseModel):
     education_level: str
     major: str
     is_first_hackathon: bool
-    portfolio_link: Union[HttpUrl, None]
-    linkedin_link: Union[HttpUrl, None]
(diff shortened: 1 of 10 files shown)


samderanova (approved):

samderanova:
I'm not sure if we can test successful application uploads because we aren't allowed to log in locally, unless the only thing you wanted to test was getting `401 Unauthorized`.

taesungh:
My bad, we should provide an internal fake login endpoint, tracking in #118.
Regarding the lingering submission issues, we can discuss more in #90 if we'll force client-side submission or be flexible on the API side.

Issue #89, opened 2023-12-16
Guest Login Page


samderanova:
Once non-UCI users enter their emails, they should be sent to a guest login page notifying them that they have been emailed a passphrase. See [`GuestLogin`](https://github.com/HackAtUCI/HackUCI-Site/blob/main/src/views/guest-login/GuestLogin.tsx) for details on how to implement this.

taesungh:
Note: backend was added in #92. We'll want to revert 7bd7405 when adding the frontend guest verification page.

rileysw:
Design is dependent on #86

taesungh:
@alexanderl19, unstyled functional form added in #98, design with working custom validation was provided in #114 which can be copied.

PR #114, opened 2023-12-17
Set up login form

### LINKED ISSUE #81: Create login pages
This page should exist on `/login`, consisting of a form that has only one input asking for the user's email.

~~The user must enter a `.edu` address. If the address is a `uci.edu` one, we should redirect them to `/api/saml/login`, which is where our UCI Shibboleth login is placed. Otherwise, we should send a POST request to `/api/guest/login` containing the body `{ email: [user's email] }` and redirect them to a guest login page telling them that a magic link has been emailed to them (see #68).~~

Edit: As [`LoginForm`](https://github.com/HackAtUCI/HackUCI-Site/blob/main/src/views/login/components/LoginForm.tsx) implies, if the email is valid, we can redirect to `/api/user/login`.

samderanova:
Resolves #81

- Add styles to the page according to the Figma
- Redirect to `/portal` if a user is already logged in
- Added client-side form validation


apps/site/src/app/login/components/LoginForm.tsx
```diff
@@ -1,31 +1,61 @@
...
+	const handleSubmit = (event: FormEvent<HTMLFormElement>): void => {
+		const form = event.currentTarget;
+		if (!form.checkValidity()) {
+			// prevent submission to display validation feedback
+			event.preventDefault();
+			setValidated(false);
+		} else {
+			setValidated(true);
+		}
+	};
```
taesungh, line 23: Suggestion: the `ValidatingForm` was designed to be a shared component between `LoginForm` and the guest `VerificationForm` since both will want to use custom form validation. Could we adapt `ValidatingForm` to fit both components?

apps/site/src/app/login/components/LoginForm.tsx
```diff
@@ -1,31 +1,61 @@
...
+		<form
+			noValidate
+			method="post"
+			action={LOGIN_PATH}
+			onSubmit={handleSubmit}
+			className="bg-white p-10 rounded-2xl mx-5 text-black"
+		>
+			<div className="flex flex-col mb-5">
+				<label htmlFor="email">
+					Email <RequiredAsterisk />
```
taesungh, line 35: Thought/suggestion: required asterisk feels redundant when there is only one input field.

apps/site/src/app/login/components/LoginForm.tsx
samderanova: Sure, but how would we pass the validated state down to the children?

apps/site/src/app/login/components/LoginForm.tsx
```diff
@@ -1,31 +1,61 @@
...
+				</small>
+				{validated === false ? (
+					<p className="feedback invalid text-red-500 peer">
+						Sorry, that email address is invalid.
+					</p>
+				) : null}
 			</div>
-			<button type="submit">Continue</button>
-		</ValidatingForm>
+			<Button text="Continue" />
```
taesungh, line 57: Chore: button is missing `type="submit"`?

samderanova: If `href` isn't provided, then the button is by default of `type="submit"` in Button.tsx.

apps/site/src/app/login/components/LoginForm.tsx
taesungh: Ahh, I'm a bit too used to React Bootstrap. Looks like the way they do it is by applying a `was-validated` class to the form element based on the prop ([Form.tsx:56](https://github.com/react-bootstrap/react-bootstrap/blob/master/src/Form.tsx#L56)) and then Bootstrap displays the validation accordingly.
```css
.was-validated :invalid ~ .invalid-feedback { display: block }
.invalid-feedback { display: none }
```
We don't necessarily have to do it the same way, but this seems rather straightforward. Otherwise, there might be some way with `useContext` but that seems a bit complicated.

apps/site/src/app/login/components/LoginForm.tsx
taesungh: Ahh got it. Unsure if that's a reasonable default, but that's out of the scope of this PR.

apps/site/src/app/login/components/LoginForm.tsx
samderanova: Implemented in `[5f33e63](https://github.com/HackAtUCI/irvinehacks-site-2024/pull/114/commits/5f33e636c09e5c74ac95d72aaf576342ff434bbf)`

apps/site/src/app/login/components/LoginForm.tsx
```diff
@@ -1,30 +1,35 @@
+import Button from "@/lib/components/Button/Button";
 import ValidatingForm from "@/lib/components/ValidatingForm/ValidatingForm";
 
-const EMAIL_REGEX = /^\w+([.-]?\w+)*@\w+([.-]?\w+)*(\.\w{2,3})+$/;
+const EMAIL_REGEX = /^\w+([\.\-]?\w+)*@\w+([\.\-]?\w+)*(\.\w{2,3})+$/;
```
taesungh, line 4: Note escaping `.` and `-` within the character class shouldn't be required but I understand your browser seemed to be enabling strict unicode mode or something.

apps/site/src/lib/components/ValidatingForm/ValidatingForm.module.scss
```diff
@@ -0,0 +1,9 @@
+.validated {
+	&:invalid p {
+		display: block;
+	}
+}
+
+.notYetValidated p {
+	display: none;
+}
```
taesungh, line 9: Nitpick [non-blocking]: bit weary of selecting on `p` elements instead of a more specific class name, but this will work for now. Coincidentally, this will already work out for the Guest Login page.

taesungh (dismissed):
Works for now, thanks for figuring out how to work with the `ValidatingForm` component, will be helpful when finishing #89 as well.

apps/site/src/lib/components/ValidatingForm/ValidatingForm.module.scss
samderanova: Implemented in [`bc51652`](https://github.com/HackAtUCI/irvinehacks-site-2024/pull/114/commits/bc516527b7a991b47d5d1b2bb1863191b61a020b).

taesungh: Okay, note for later that we'll need to make the corresponding change on the Guest Login `VerificationForm`.

taesungh (approved):
Works out, let's **squash** merge.

PR #115, opened 2023-12-17
update: SSO redirect to /portal instead of /

### LINKED ISSUE #112: Redirect to portal after SSO login
Currently the guest authentication redirects the user to `/portal` once logged in whereas the SAML authentication redirects the user to the home page (`/`).
- For consistency, update the SAML ACS redirect to `/portal`
  - The corresponding unit test will also need to be updated
- The actual redirection/routing based on status will be handled by the front end

samderanova:
Resolves #112


taesungh (approved):
Nice and easy.

PR #116, opened 2023-12-17
Resume upload field improvements

### LINKED ISSUE #111: Improve resume upload field

samderanova:
Resolves #111 .

- Made resume field optional
- More accurate validation text ("selected" instead of "uploaded")
- Changed icon to file icon instead of download icon


IanWearsHat (approved):
Looks good to me. Sorry again for blocking.

apps/site/src/app/apply/sections/Form/ResumeInformation.tsx
```diff
@@ -67,9 +66,9 @@ export default function ResumeInformation() {
 			</label>
 			<label
 				htmlFor="resume_upload"
-				className={`${styles.upload} cursor-pointer mb-3`}
+				className={`${styles.upload} cursor-pointer mb-3 p-5 rounded-xl bg-[#e1e1e1]`}
 			>
-				<Image src={uploadImage} width="100" alt="Upload resume icon" />
+				<FileText className="m-auto" width={50} height={50} />
```
taesungh, line 71: Note: the original suggestion I had was using two different icons, one for before selecting and one for after selecting a valid file. Might be cool to have a third icon for invalid (too large or wrong type) but not essential.

samderanova: Implemented the two icon suggestion in [f3ba74c](https://github.com/HackAtUCI/irvinehacks-site-2024/pull/116/commits/f3ba74c924ddace989d4630be4c1a4122f272411).

apps/site/src/app/apply/sections/Form/ResumeOutputFeedback.tsx
```diff
@@ -13,7 +13,7 @@ export default function OutputFeedBack({
 
 	return (
 		<span className="text-xl">
-			{resumePath ? "Successfully uploaded " + resumePath : ""}
+			{resumePath ? "Selected " + resumePath : ""}
 		</span>
```
taesungh, line 17: Suggestion: this text is even larger than the input field label. Could we make it a bit smaller and also perhaps change the text color or decoration (maybe italic or underline) of the file name so it is a little more clear to the user?

taesungh (commented):
I'll be adding the API handler for the empty octet stream since we aren't removing the file field on the client in #125 and didn't account for it in #128.

apps/site/src/app/apply/sections/Form/ResumeInformation.tsx
taesungh: Actually, I think we do need the tri-state: the green check icon would conflict with the invalid message shown in red below.

apps/site/src/app/apply/sections/Form/ResumeOutputFeedback.tsx
```diff
@@ -8,12 +8,12 @@ export default function OutputFeedBack({
...
 	if (errorMessage) {
-		return <span className="text-[#FF2222] text-xl">{errorMessage}</span>;
+		return <span className="text-[red] italic">{errorMessage}</span>;
 	}
 
 	return (
-		<span className="text-xl">
-			{resumePath ? "Successfully uploaded " + resumePath : ""}
+		<span className="italic">
+			{resumePath ? "Selected " + resumePath : ""}
```
taesungh, line 16: Note: I meant just the file name in italics, but this is also fine.

taesungh (approved):
Thanks for adding the different icons for each state.

apps/site/src/app/apply/sections/Form/ResumeInformation.tsx
```diff
@@ -67,9 +68,25 @@ export default function ResumeInformation() {
 			</label>
```
IanWearsHat, line 68: Issue: If we have made the resume optional, then we should remove the asterisk in the label

IanWearsHat (commented):
Just one minor text issue in the label

taesungh (approved):

PR #120, opened 2023-12-17
feat: restore next/link

### LINKED ISSUE #117: Change `a` tags in `Button.tsx` back to `Link`s
As a quick and dirty fix to get logging out working, the `Link` tags were changed to `a` tags in `Button.tsx`, as they were causing issues with page reloads. This issue is devoted to putting them back.

samderanova:
Resolves #117 .


apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -85,12 +87,15 @@ function Navbar({ identity }: NavbarProps) {
...
 							href="/logout"
-							usePrefetch={false}
-							isLightVersion
-						/>
+							className={clsx(
+								buttonStyles.button,
+								buttonStyles.lightButton,
+							)}
+						>
+							Logout
```
taesungh, line 97: Technically out of scope, but "Log In" and "Log Out" as verbs rather than adjectives would make more sense as action descriptions.

samderanova: It's a minor change so I'll just add it in this PR.

taesungh (approved):
Works out. Tried it with the application preface just added and those buttons also navigate seamlessly.

IanWearsHat (approved):
Everything also including the application preface works for me

PR #124, opened 2023-12-18
Fix Flashing Form Validation

alexanderl19:
Only show validation messages if form is failing since the user is going to be redirected if successful anyways.

### DIFF
--- a/apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
+++ b/apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
@@ -17,8 +17,8 @@ function ValidatingForm(props: PropsWithChildren<FormProps>) {
 		if (!form.checkValidity()) {
 			// prevent submission to display validation feedback
 			event.preventDefault();
+			setValidated(true);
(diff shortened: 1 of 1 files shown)


samderanova (approved):
Thanks for the quick fix!

apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
```diff
@@ -17,8 +17,8 @@ function ValidatingForm(props: PropsWithChildren<FormProps>) {
 		if (!form.checkValidity()) {
 			// prevent submission to display validation feedback
 			event.preventDefault();
+			setValidated(true);
 		}
-		setValidated(true);
```
taesungh, line 21: While this does technically work for our current use cases, it is not entirely flexible. Hypothetically, the form might want to show positive feedback for a valid field. I think the real problem is in the CSS selectors for the equivalent of `was-validated` and `invalid-feedback`.

taesungh:
Closing in favor of complete validation feedback in #135.

Issue #108, opened 2023-12-17
Application Preface


samderanova:
As requested from applications, we'll need to set up a page for applicants where applicants must acknowledge that they understand our in-person policy before proceeding to the application.

PR #125, opened 2023-12-18
Implement client-side application form submission

### LINKED ISSUE #90: Enable application submission to backend route

taesungh:
To handle error cases and show the submission completed message, the application form can be submitted asynchronously from the client with axios. This handles the unauthorized case and the success case can be handled separately as part of the entire Apply page flow in #108.

I decided not to omit the empty resume file (empty octet stream) since the API should be reinforced to handle that (in case JavaScript is disabled) as part of #93.

The form submission handler could have been implemented as a [Next.js server action](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations) which would alleviate some of our concerns about JavaScript being disabled, but we are currently using [Next.js v13 where server actions are considered experimental](https://nextjs.org/docs/app/api-reference/next-config-js/serverActions#enabling-server-actions-v13). To avoid potentially breaking changes, we can revisit upgrading to v14 after our application release.

Closes #90.

## Changes
- Implement client-side application form submission
  - Use axios to handle application submission on the client
    - Substitute write-in values marked as "other"
- Disable Apply form submit button while submitting
  - When submitting, disable the submit button so the user doesn't accidentally submit twice
- Handle session expiration in application form
  - When apply form submission encounters status 401, show session expired message instructing user to log in again from a new tab

## Testing
1. Open the application form at `/apply`
2. Select "Other" options for the fields that have such and provide a write-in value
3. Submit the application and observe the other values are substituted in the form data (console.debug)
4. Observe the session expired message appears below the submit button

### DIFF
--- a/apps/site/src/app/apply/sections/Form/Form.tsx
+++ b/apps/site/src/app/apply/sections/Form/Form.tsx
@@ -18,6 +18,7 @@ const FIELDS_WITH_OTHER = ["pronouns", "ethnicity", "school", "major"];
 
 export default function Form() {
 	const [submitting, setSubmitting] = useState(false);
+	const [sessionExpired, setSessionExpired] = useState(false);
 
 	const handleSubmit = async (
 		event: FormEvent<HTMLFormElement>,
@@ -25,6 +26,7 @@ export default function Form() {
 		// Disable native post submission
 		event.preventDefault();
 		setSubmitting(true);
+		setSessionExpired(false);
 
 		const formData = new FormData(event.currentTarget);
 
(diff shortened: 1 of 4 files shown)


apps/site/src/app/apply/sections/Form/Form.tsx
```diff
@@ -8,20 +13,82 @@ import ResumeInformation from "./ResumeInformation";
...
+			<a
+				href="/login"
+				target="_blank"
+				className="text-blue-600 underline"
+			>
+				log in from a new tab
+			</a>{" "}
+			to restore your session and then try submitting again.
+		</p>
+	);
```
samderanova, line 75: Issue: We'll want to add some margin on the sides of this element on mobile views.

image

samderanova (changes requested):
Looking good, thanks for getting this done and for preventing a user from spamming the submit button. A few things to note:

apps/site/src/app/apply/sections/Form/Form.tsx
```diff
@@ -8,20 +13,82 @@ import ResumeInformation from "./ResumeInformation";
...
+
+		const formEntries = Object.fromEntries(formData.entries());
+		console.debug(formEntries);
+
+		try {
+			const res = await axios.post(APPLY_PATH, formData);
+			if (res.status === 201) {
+				console.log("Application submitted");
+				// TODO: show submitted message
+			}
```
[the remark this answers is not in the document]
taesungh: I left it blank to decide as part of the entire flow (see PR description).

apps/site/src/app/apply/sections/Form/Form.tsx
taesungh: Looks like we're gonna need a container design overhaul but added for now in 196d33f.

samderanova (approved):

IanWearsHat (approved):

Issue #126, opened 2023-12-18
Collapsed Navbar creates dead zone


taesungh:
When collapsed, the navigation bar is too tall and leaks below its actual content box. As a result, interactive elements near the top of the screen do not work.

<img src="https://github.com/HackAtUCI/irvinehacks-site-2024/assets/44419552/1eb879da-ae1a-47a5-a5da-5883f60c40be" width=340 alt="Screenshot of web inspector showing navigation bar element occupying more space than it should" />

PR #127, opened 2023-12-18
#126 resolved, clip overflow region when not needed

Bl20052005:


apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -34,8 +34,8 @@ function Navbar({ identity }: NavbarProps) {
 
 	return (
 		<NavMenu.Root
-			className={`${
-				hasScrolled ? "md:bg-opacity-50" : ""
+			className={`${hasScrolled ? "md:bg-opacity-50" : ""} ${
+				hidden ? "max-[768px]:h-[4rem] max-[768px]:overflow-hidden" : ""
```
IanWearsHat, line 38: Nitpick: Instead of `[768px]`, you could use `md`

IanWearsHat (commented):
it looks like it works, there's just one small nitpick

IanWearsHat (approved):

samderanova (commented):
Thanks for the quick fix. Please *squash* merge.

samderanova (approved):

PR #128, opened 2023-12-18
feat: handle application when JS disabled

samderanova:
In scenarios like these, the value "other" will be sent instead of the value the user provided when selecting the "other" option.

- Raise an `HTTPException` with status `422 Unprocessable Entity` when receiving such an error
- Write a unit test to test an application with a field containing "other" as its value

### DIFF
--- a/apps/api/src/routers/user.py
+++ b/apps/api/src/routers/user.py
@@ -72,6 +72,15 @@ async def apply(
         log.error("User %s has already applied.", user)
         raise HTTPException(status.HTTP_400_BAD_REQUEST)
 
+    raw_app_data_dump = raw_application_data.model_dump()
+
+    for field in ["pronouns", "ethnicity", "school", "major"]:
+        if raw_app_data_dump[field] == "other":
+            raise HTTPException(
+                status.HTTP_422_UNPROCESSABLE_ENTITY,
+                "Please enable JavaScript on your browser.",
+            )
+
     if resume is not None:
         try:
             resume_url = await resume_handler.upload_resume(
@@ -95,7 +104,7 @@ async def apply(
 
     now = datetime.now(timezone.utc)
     processed_application_data = ProcessedApplicationData(
-        **raw_application_data.model_dump(),
+        **raw_app_data_dump,
         resume_url=resume_url,
         submission_time=now,
     )
--- a/apps/api/tests/test_user_apply.py
+++ b/apps/api/tests/test_user_apply.py
@@ -36,20 +36,6 @@ SAMPLE_APPLICATION = {
     "frq_dream_job": "I am pkfire",
 }
 
-SAMPLE_APPLICATION_NO_JS = {
-    "first_name": "pk",
-    "last_name": "fire",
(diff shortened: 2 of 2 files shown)


apps/api/src/routers/user.py
```diff
@@ -72,6 +72,15 @@ async def apply(
         log.error("User %s has already applied.", user)
         raise HTTPException(status.HTTP_400_BAD_REQUEST)
 
+    raw_app_data_dump = raw_application_data.model_dump()
+
+    for field in ["pronouns", "ethnicity", "school", "major"]:
+        if raw_app_data_dump[field] == "other":
```
taesungh, line 78: Thought [non-blocking]: unsure if dump or `__getattr__` would be more performant, but this works fine especially cause we'll need the dump later anyway.

apps/api/src/routers/user.py
```diff
@@ -72,6 +72,15 @@ async def apply(
...
+            raise HTTPException(
+                status.HTTP_422_UNPROCESSABLE_ENTITY,
```
taesungh, line 80: Praise: thank you for using a named status value

apps/api/tests/test_user_apply.py
```diff
@@ -36,6 +36,21 @@
...
+    "ethnicity": "fire",
+    "is_18_older": "true",
+    "school": "UC Irvine",
+    "education_level": "Fifth+ Year Undergraduate",
+    "major": "Computer Science",
+    "is_first_hackathon": "false",
+    "portfolio": "https://github.com",
+    "frq_collaboration": "I am pkfire",
+    "frq_dream_job": "I am pkfire",
+}
```
taesungh, line 52: Suggestion: instead of writing out a full copy of the sample data, we could use `model_copy` and manually override fields, see `test_application_data_is_bson_encodable`

apps/api/tests/test_user_apply.py
```diff
@@ -260,3 +275,19 @@ def test_application_data_is_bson_encodable() -> None:
     data.linkedin = HttpUrl("https://linkedin.com")
     encoded = bson.encode(EXPECTED_APPLICATION_DATA.model_dump())
     assert len(encoded) == 415
+
+
+@patch("utils.email_handler.send_application_confirmation_email", autospec=True)
+@patch("services.mongodb_handler.update_one", autospec=True)
+@patch("routers.user.datetime", autospec=True)
+@patch("services.gdrive_handler.upload_file", autospec=True)
+@patch("services.mongodb_handler.retrieve_one", autospec=True)
```
taesungh, line 284: Chore: pretty sure we don't need all of these patches. I think all we care about is retrieve_one is not called. See `test_apply_with_invalid_data_causes_422`

taesungh (commented):
Nice start, some suggestions about the unit tests, but seems to work out functionally.

taesungh (approved):
Nice work on this and figuring out the unit tests. Reminder that we'll still need to handle the empty resume file on the API side. Let's **squash** when ready.

PR #134, opened 2023-12-18
NavLinkItems are now composed of nextLinks

IanWearsHat:
Every nav link should now load new components without reloading the page

Navbar will also show which link is currently active

### DIFF
--- a/apps/site/src/lib/components/Navbar/NavbarHelpers.tsx
+++ b/apps/site/src/lib/components/Navbar/NavbarHelpers.tsx
@@ -11,15 +11,14 @@ const NavLinkItem = React.forwardRef<
 >(({ children, className, href, ...props }, forwardedRef) => {
 	return (
 		<NavMenu.Item>
-			<Link href={href as string} passHref legacyBehavior>
-				<NavMenu.Link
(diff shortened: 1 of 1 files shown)


apps/site/src/lib/components/Navbar/NavbarHelpers.tsx
```diff
@@ -1,22 +1,25 @@
...
 	return (
 		<NavMenu.Item>
-			<NavMenu.Link
-				className={clsx(styles.navMenuLink, className)}
-				{...props}
-				ref={forwardedRef}
-			>
-				{children}
-			</NavMenu.Link>
+			<Link href={href as string} passHref legacyBehavior>
```
taesungh, line 14: Thought: there may be a way to do this with Radix's `asChild` prop instead of using `legacyBehavior` although I'm not sure if it will actually work.

IanWearsHat: The docs have an example where they use legacyBehavior, so I just adapted that example to here.

I also have no idea how to use the `asChild` prop.

samderanova (approved):

Bl20052005 (approved):

PR #136, opened 2023-12-18
Redirect applicants who have successfully submitted applications (if possible)

samderanova:
Redirect applicant to `/portal` if JS is enabled, otherwise have the API display a simple message telling them to check their portal.

### DIFF
--- a/apps/site/src/app/apply/sections/Form/Form.tsx
+++ b/apps/site/src/app/apply/sections/Form/Form.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import { FormEvent, useState } from "react";
-import { redirect } from "next/navigation";
+import { useRouter } from "next/router";
 import axios from "axios";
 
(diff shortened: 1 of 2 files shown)


apps/site/src/app/apply/sections/Form/Form.tsx
```diff
@@ -46,7 +47,7 @@ export default function Form() {
 			const res = await axios.post(APPLY_PATH, formData);
 			if (res.status === 201) {
 				console.log("Application submitted");
-				// TODO: show submitted message
+				redirect("/portal");
```
taesungh, line 50: Couldn't get the redirect working, I think because this is a Client Component, so we need to call push with `useRouter`.

taesungh (approved):
Thanks for considering the non-JavaScript case. Tested and sends me to Portal after applying. Lingering issue redirect when already applied: #137.

PR #139, opened 2023-12-18
fix: use correct version of useRouter

### LINKED ISSUE #138: Switch import for `useRouter`
`useRouter` is imported from two places: `next/router` and `next/navigation`. We are currently using the first, which should only be used in the Pages Router, not the App Router.

samderanova:
Resolves #138 .


taesungh (approved):
Oof, this would have been caught in #136 if we kept preview deployments on. Thanks for the quick fix.

Issue #137, opened 2023-12-18
Redirect to `/portal` from `/apply` when user has already applied


taesungh:
If a user has already applied, visiting `/apply` should redirect to `/portal`.

PR #140, opened 2023-12-18
fix: redirect user from /apply to /portal

samderanova:
Although the `/apply` route is already obscured from applicants who have already submitted an application and the backend prevents the resubmission of applications, applicants can still access the `/apply` page by directly navigating to it (i.e. typing in the URL search bar). If a user has already submitted an application, they should not be able to access the apply page again, so we should redirect them to `/portal`.

### DIFF
--- a/apps/site/src/app/apply/page.tsx
+++ b/apps/site/src/app/apply/page.tsx
@@ -1,23 +1,30 @@
 import Image from "next/image";
-
-import koiLeft from "@/assets/images/koi-swim-left.png";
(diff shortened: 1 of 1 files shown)


Bl20052005 (approved):

taesungh (approved):
Can't test entire application on preview since this was based after #137 but before #139, but redirect works purely based on status. Trusting this doesn't break the application itself. Do sometimes see a slight flash of footer coming up, maybe we need to revisit global min-height.

PR #142, opened 2023-12-18
fix: set portal section height to height of screen

### LINKED ISSUE #141: Resolve min-height issue in portal
In the portal page, the footer is not aligned to the bottom of the screen. This can be fixed by applying a `min-height` to the portal container.

samderanova:
Resolves #141.


taesungh (approved):
Layout could be improved more (e.g. centered vertically), but this will work for now. Thanks for the quick fix.

PR #144, opened 2023-12-18
Revert "Temporarily replace guest token email with print" to re-enable guest email authentication

samderanova:
This reverts commit 7bd7405ef8f9415328ce05df53b309d71f71edd1.


taesungh (approved):
Nice use of `git revert`. Assuming the handler will work even in prod.

Issue #147, opened 2023-12-18
Fix `Too Many Requests` when logging in as a guest


samderanova:
One method of reproducing this is by logging in as a guest, logging back out, and then logging in again as a guest. This is because passphrases take 10 minutes to expire.

This can be resolved by deleting the passphrase once it's verified and in case this situation comes up again, we should provide a more helpful message instead of `Too Many Requests`.

PR #148, opened 2023-12-18
Fix/guest auth 429

samderanova:
Addresses #147.

### DIFF
--- a/apps/api/tests/test_guest.py
+++ b/apps/api/tests/test_guest.py
@@ -105,6 +105,8 @@ def test_successful_guest_verification_provides_identity(
     assert res.status_code == 303
     assert res.headers["Set-Cookie"].startswith("irvinehacks_auth=")
 
+    mock_remove_guest_key.assert_awaited_once_with("edu.caltech.beaver")
+
 
 @patch("auth.guest_auth._get_existing_key", autospec=True)
 @patch("auth.guest_auth._remove_guest_key", autospec=True)
--- a/apps/api/tests/test_guest.py
+++ b/apps/api/tests/test_guest.py
@@ -84,11 +84,11 @@ def test_requesting_login_when_previous_key_exists_causes_429(
     assert res.status_code == 429
 
 
-@patch("auth.guest_auth._get_existing_key", autospec=True)
 @patch("auth.guest_auth._remove_guest_key", autospec=True)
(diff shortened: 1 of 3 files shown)


rileysw (approved):
Code looks good, did you test and try to reproduce the error?

IanWearsHat:
Logging in twice as guest still raises the exception for me, although it is the updated message now.

apps/api/tests/test_guest.py
```diff
@@ -85,7 +85,9 @@ def test_requesting_login_when_previous_key_exists_causes_429(
 
 
 @patch("auth.guest_auth._get_existing_key", autospec=True)
+@patch("auth.guest_auth._remove_guest_key", autospec=True)
 def test_successful_guest_verification_provides_identity(
+    mock_remove_guest_key: AsyncMock,
     mock_get_existing_key: AsyncMock,
```
taesungh, line 91: Suggestion [repeating]: patch order should usually reflect order of usage/calls during test.

apps/api/tests/test_guest.py
```diff
@@ -105,7 +107,9 @@ def test_successful_guest_verification_provides_identity(
 
 
 @patch("auth.guest_auth._get_existing_key", autospec=True)
+@patch("auth.guest_auth._remove_guest_key", autospec=True)
 def test_invalid_guest_verification_is_unauthorized(
+    mock_remove_guest_key: AsyncMock,
```
taesungh, line 112: Suggestion [repeating]: the mock is currently unused in the function. We could call the `assert_awaited_once_with` method to further validate its usage.

taesungh (approved):
Simple enough, thanks for figuring out how to work with the unit tests.

samderanova:
I tested it on the most recent preview and it works for me.

IanWearsHat:
> Logging in twice as guest still raises the exception for me, although it is the updated message now.

Logging in, logging out, and logging back in works.

Inputting an email, backing out, and then inputting the same email triggers the exception, although I don't know if this is intended behavior.

taesungh:
> Inputting an email, backing out, and then inputting the same email triggers the exception, although I don't know if this is intended behavior.

Yes that is intended because we don't want someone who started the process to be able to spam us, especially when each initiation causes an email to be sent.

IanWearsHat (approved):
Looks good then

Issue #146, opened 2023-12-18
Navbar doesn't update identity when submitting application


taesungh:
The user identity on the site is based on server-side fetch rather than client context, so the Navbar doesn't know the user identity is updated when the user completes an application. The Navbar still shows **Apply** when that should be replaced by **Portal**.

PR #150, opened 2023-12-18
fix: use window.location.href instead of useRouter

samderanova:
Addresses #146 .

### DIFF
--- a/apps/site/src/app/apply/sections/Form/Form.tsx
+++ b/apps/site/src/app/apply/sections/Form/Form.tsx
@@ -46,6 +46,10 @@ export default function Form() {
 			const res = await axios.post(APPLY_PATH, formData);
 			if (res.status === 201) {
 				console.log("Application submitted");
+
+				// Use window.location instead of router.push in order
+				// to force reload the page to allow user identity to
+				// update with the new status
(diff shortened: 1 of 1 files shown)


apps/site/src/app/apply/sections/Form/Form.tsx
```diff
@@ -41,14 +39,14 @@ export default function Form() {
...
-		console.debug(formEntries);
+		// const formEntries = Object.fromEntries(formData.entries());
+		// console.debug(formEntries);
 
 		try {
 			const res = await axios.post(APPLY_PATH, formData);
 			if (res.status === 201) {
 				console.log("Application submitted");
-				router.push("/portal");
+				window.location.href = "/portal";
```
taesungh, line 49: Could we include a comment indicating why we use `window.location` instead of `router.push`?

taesungh (approved):
I'm trying to think of ways to use cache invalidation to maintain seamless navigation, but caching might not apply when the components using the API calls are dynamically rendered because they rely on cookies. Manually assigning the location will work for now, and this is mostly how we did it last year.

waalbert (approved):
LGTM

PR #152, opened 2023-12-19
Enforce FRQ character limit on application form

### LINKED ISSUE #151: Application Form: limit FRQ answer length
The `Textfield` inputs currently do not have a max length, so an applicant could enter a quite lengthy answer which would be rejected by the API with status 422 which won't be surface during the client-side submission.

To avoid this, we can set a `maxLength` on the `textarea` fields which corresponds with the limit in the `ApplicationData` model. This is currently 1024 characters (~200 words), whereas the form label suggests 100/150 words, but I'm not sure if this is a soft or hard limit, so we may need to adjust the character limit accordingly.

taesungh:
We received a user report of being unable to submit the application since the API rejected the application with status 405 because the FRQ answer was too long and no limit was present on the client-side form.

Resolves #151.

## Changes
- Double the character limit on FRQ answers (1024 -> 2048)
- Add `maxLength` to the `<textarea>` elements for FRQ answers
  - Form limit is slightly less since newlines are counted weirdly

## Testing
1. Open the application form
2. Complete all of the fields
3. Use a really long answer on either FRQ question
4. Observe the browser validation message appears when trying to submit
    > Use no more than 2000 characters (Safari)
    - Chrome does not seem to show any validation message
    - Both browsers will limit the amount of input characters but not exactly because of newline characters

### DIFF
--- a/apps/api/src/models/ApplicationData.py
+++ b/apps/api/src/models/ApplicationData.py
@@ -31,8 +31,8 @@ class RawApplicationData(BaseModel):
     is_first_hackathon: bool
     linkedin: Union[HttpUrl, None] = None
     portfolio: Union[HttpUrl, None] = None
-    frq_collaboration: Union[str, None] = Field(None, max_length=1024)
-    frq_dream_job: str = Field(max_length=1024)
+    frq_collaboration: Union[str, None] = Field(None, max_length=2048)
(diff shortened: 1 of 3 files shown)


rileysw (approved):
Thanks for the quick response!

PR #156, opened 2023-12-20
Navbar is rendered over stands

IanWearsHat:
z-index for navbar is now 50, which should render over everything


samderanova (approved):
Thanks for the quick fix! Please *squash* merge.

taesungh (approved):
Nice and easy.

PR #157, opened 2023-12-20
Dropdown option no longer shows dom error

IanWearsHat:
Removed selected and disabled attribute from option

This should stop the dom error from the Dropdown component


rileysw (dismissed):
Thanks for correcting this!

apps/site/src/app/apply/sections/Components/DropdownSelect.tsx
```diff
@@ -58,7 +58,7 @@ export default function DropdownSelect({
 				onChange={(e) => setValue(e.target.value)}
 				required
 			>
-				<option value="" selected disabled hidden></option>
+				<option value="" hidden></option>
```
taesungh, line 61: Thought: unsure if keeping or removing the `disabled` attribute is best for UX. Since the field is required it might make sense to leave it disabled? However, a user might want to unset the value to "come back to it later"?

IanWearsHat: I think we can have both.

Removing the `hidden` attribute would allow the option to still be selected to "come back to it later," but the form will still tell the user that the field is required.

I only have chrome and firefox to test this on, so I don't know about Safari.

taesungh: Interesting, the behavior is a bit different on Safari (and presumably WebKit) where `hidden` doesn't seem to have any effect:
- When using just `hidden`
  - Safari will show the empty option by default just as any other which can be selected again
  - Chrome does not show the empty option at all but has it selected by default
- When using `disabled hidden`
  - Safari shows the empty option by default, but it is disabled and cannot be selected again after selecting a different value
  - Chrome selects the first option (e.g. High School/UC Irvine) by default instead of the empty option
    - This can be resolved by making the `<select>` element a controlled component using `value={value}`

IanWearsHat: If Safari will also show the empty option by default which can be selected again regardless of whether the `hidden` attribute is set, then we could remove `hidden` to allow the user to select the empty option again.

samderanova: I think we should leave `disabled` in, mainly because the field is required. Also, I'm not sure if a user unsetting a value is a common occurrence because a user could also just skip the question after viewing the options to come back to it later.

taesungh (approved):
Sorry for dragging this on so long, this will work for now.

samderanova (approved):
Thanks for all of your work on this PR! Please *squash* merge when ready.

PR #159, opened 2023-12-28
feat: import ESLint config from last year

### LINKED ISSUE #155: Enforce more ESLint rules
In some recent PRs, we've been commenting about some code improvements (such as `eqeqeq`) which could be caught by our checks workflow, but the ESLint config currently does not check for the basic rules for ESLint and TypeScript.

samderanova:
Resolves #155


apps/site/src/components/canvas/View.tsx
```diff
@@ -1,28 +1,32 @@
...
-});
+		return (
+			<>
+				<div ref={localRef} className={className} />
+				{/* Sends View component thru tunnel to render on global Canvas in Scene.tsx */}
+				<Three>
+					{/* View component to render 3D view in a div (uses gl.scissor to cut viewport into segments) */}
+					{/* eslint-disable-next-line @typescript-eslint/ban-ts-comment */}
+					{/* @ts-ignore */}
+					<ViewImpl track={localRef}>{children}</ViewImpl>
```
taesungh, line 24: Using a non-null assertion on the initial value would probably safer than completely ignoring types for the entire line. See [Typing with useRef in the React Three Fiber Tutorials](https://docs.pmnd.rs/react-three-fiber/tutorials/typescript#typing-with-useref).

samderanova: This is exactly what I was looking for, thank you!

taesungh: Or maybe the `ViewImpl` should be conditionally rendered when `localRef` is populated?

apps/site/src/components/canvas/View.tsx
```diff
@@ -1,28 +1,32 @@
...
 import { Three } from "@/helpers/components/Three";
 
-const View = forwardRef(({ children, orbit, ...props }: any, ref) => {
-	const localRef = useRef<any>();
-	useImperativeHandle(ref, () => localRef.current);
+const View = forwardRef(
+	(
+		{ className, children }: PropsWithChildren<{ className: string }>,
+		ref,
+	) => {
```
taesungh, line 12: Question: partly out of scope, but is there a particular reason this component needs to use `forwardRef`? Don't see any consumers passing in refs.

apps/site/src/components/canvas/View.tsx
samderanova: Hmm, based on the article you sent, I think the non-null assertion operation is the best approach (and the simplest).

apps/site/src/components/canvas/View.tsx
samderanova: I don't think there is. I'll go ahead and convert it to a standard functional component.

taesungh (approved):
Thanks for adding these, will definitely help to have these improved automated checks.

waalbert (approved):
LGTM

PR #161, opened 2023-12-28
Update Sanity resource schema

samderanova:
- Create new schema for resource categories
- Adjusted category field in current resource schema to reference this new schema
- Changed FAQ icon as it conflicts with the icon for the new schema

### DIFF
--- a/apps/sanity/schemas/faqs.ts
+++ b/apps/sanity/schemas/faqs.ts
@@ -6,6 +6,7 @@ export default defineType({
 	name: "faqs",
 	title: "FAQs",
 	type: "document",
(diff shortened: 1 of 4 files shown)


waalbert (approved):
LGTM

taesungh (approved):
If we're following a similar design as last year, I think we'll also eventually need to add the individual resource items to the category? We can probably handle that when the time comes.

samderanova:
Actually, this new schema is meant for *creating* categories. I modified the `category` field in the current `resource` schema so that you can select any category that was created through the new one.

PR #162, opened 2023-12-29
fix/mentor and volunteer app forms dont cause cors errors

IanWearsHat:
Disabled prefetch for mentor and volunteer links on the application preface to stop cors preflight errors

### DIFF
--- a/apps/site/src/app/apply/sections/ApplyConfirmation/ConfirmationDetails.tsx
+++ b/apps/site/src/app/apply/sections/ApplyConfirmation/ConfirmationDetails.tsx
@@ -35,13 +35,13 @@ export default async function ConfirmationDetails({
 			<hr className="mt-5 w-full h-0.5 bg-[#432810]" />
 			<p className="text-display">
 				Interested in helping out instead? Consider applying to be a{" "}
-				<Link className="underline" prefetch={false} href="/mentor">
+				<a className="underline" href="/mentor">
 					mentor
-				</Link>{" "}
+				</a>{" "}
 				or a{" "}
-				<Link className="underline" prefetch={false} href="/volunteer">
+				<a className="underline" href="/volunteer">
 					volunteer
-				</Link>
+				</a>
 				.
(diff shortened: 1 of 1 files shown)


apps/site/src/app/apply/sections/ApplyConfirmation/ConfirmationDetails.tsx
```diff
@@ -35,13 +35,13 @@ export default async function ConfirmationDetails({
 			<hr className="mt-5 w-full h-0.5 bg-[#432810]" />
 			<p className="text-display">
 				Interested in helping out instead? Consider applying to be a{" "}
-				<Link className="underline" href="/mentor">
+				<a className="underline" href="/mentor">
```
samderanova, line 38: Issue: Instead of using `a` tags, you can set the `prefetch` attribute to false. See the [`next/link`](https://nextjs.org/docs/app/api-reference/components/link) documentation for an example.

samderanova (changes requested):
Thanks for picking this up! One small thing to address:

apps/site/src/app/apply/sections/ApplyConfirmation/ConfirmationDetails.tsx
IanWearsHat: I tried just disabling prefetch and that stopped the cors error from happening on the application preface. 

However, clicking the next link raised another cors error, but still allowed me to access the Google form.

taesungh (approved):

apps/site/src/app/apply/sections/ApplyConfirmation/ConfirmationDetails.tsx
samderanova: I see. We can stick to `a` tags then, especially since the form isn't a page on our site.

samderanova (approved):

samderanova:
Please *squash* merge when ready!

PR #163, opened 2023-12-29
Add event cards for each day to Schedule

waalbert:
- Creates the schedule page according to Figma using [Radix Tabs](https://www.radix-ui.com/primitives/docs/components/tabs)
- Updates Sanity Event schema to allow for Event Types (Main, Workshop, Social, Announcement, Miscellaneous)
- [date-fns-tz](https://github.com/marnusw/date-fns-tz) and [dayjs](https://day.js.org/) are used to calculate the duration of events using Dates set in Pacific Standard Time
- If there isn't an organization hosting an event, then whoever's in the Sanity "hosts" field will be displayed instead
- For announcements, the description field from the event Sanity schema is used as the rendered title for the announcement, allowing for the use of Portable Text
- Adds new smaller Tailwind breakpoint `"xs"` for even smaller phone screens


samderanova (changes requested):
Great start on the schedule page so far! A few things to point out:

apps/site/src/app/schedule/page.tsx
```diff
@@ -1,16 +1,29 @@
+import { PortableText } from "@portabletext/react";
+
 import ShiftingCountdown from "./components/ShiftingCountdown/ShiftingCountdown";
+import { getSchedule } from "./components/getSchedule";
+import SchedulePage from "./sections/SchedulePage";
 
 export const revalidate = 60;
```
samderanova, line 8: Issue: Could we export metadata like how we do for the login `page.tsx`? We should have a descriptive title for our pages (such as `Schedule | IrvineHacks 2024`).

samderanova (changes requested):

apps/site/package.json
```diff
@@ -17,12 +17,15 @@
...
 		"@radix-ui/react-toast": "^1.1.5",
+		"@radix-ui/themes": "^2.0.3",
 		"@react-three/drei": "^9.88.14",
 		"@react-three/fiber": "^8.15.11",
 		"@sanity/image-url": "^1.0.2",
 		"@types/three": "^0.158.2",
 		"axios": "^1.6.2",
 		"clsx": "^2.0.0",
+		"date-fns-tz": "^2.0.0",
+		"dayjs": "^1.11.10",
```
taesungh, line 28: Thought/Suggestion: Day.js provides their own [Timezone plugin](https://day.js.org/docs/en/plugin/timezone), so is there a particular advantage in using this other date-fns-tz library?

apps/site/src/app/schedule/sections/SchedulePage.tsx
```diff
@@ -0,0 +1,140 @@
...
+		};
+	}, []);
+
+	return (
+		<Tabs.Root
+			value={day}
+			defaultValue="Friday"
+			onValueChange={setDay}
+			className="w-11/12 sm:w-4/5"
+		>
```
taesungh, line 82: Thought/question: Did we confirm with graphics that we're doing this new tabbed approach? For reference, a few years back we did "anchor navigation" with "scroll spy", and last year we just had a long list with no separations.

apps/site/src/app/schedule/components/EventRegular.tsx
```diff
@@ -0,0 +1,122 @@
...
+			</div>
+		);
+	} else if (eventType === "Social") {
+		return (
+			<div className="inline-block px-4 py-1.5 font-semibold text-[#0D272D] bg-[#DFA9A9] rounded-2xl sm:py-2">
+				{eventType}
+			</div>
+		);
+	}
+}
```
samderanova, line 54: Issue: The only difference between the returned elements is their background color. You could simplify this even further by initializing a string variable, setting it to be the appropriate color based on `eventType`, and then adding that variable to the `className` string.

apps/site/src/app/schedule/sections/SchedulePage.tsx
```diff
@@ -0,0 +1,140 @@
...
+		title: string;
+		eventType: string;
+		location?: string | undefined;
+		virtual?: string | undefined;
+		startTime: Date;
+		endTime: Date;
+		organization?: string | undefined;
+		hosts?: string[] | undefined;
+		description: JSX.Element;
+	}[][];
```
taesungh, line 29: Thought: it seems we're repeating a lot of `EventProps`. Would there be a way to derive the necessary field types?

apps/site/src/app/schedule/sections/SchedulePage.tsx
samderanova: This is what was on the Figma, so that was what Albert proceeded with. I can double check with graphics.

samderanova (changes requested):

apps/site/src/app/schedule/sections/SchedulePage.tsx
samderanova: Graphics just said yes

apps/site/src/app/schedule/sections/SchedulePage.tsx
taesungh: Or better yet, would there be a way to use inference from zod, e.g. `z.infer<typeof Events>`?

apps/site/src/app/schedule/components/EventRegular.tsx
waalbert: I tried doing this before which worked locally, but the Vercel preview wasn't able to set the right background color
594aca7cd2747c38a801af4a902307d38fafde39

samderanova: Just found out that it's actually not recommended to dynamically declare classes like this because Tailwind won't be able to [find them](https://tailwindcss.com/docs/content-configuration). In that case, please use inline CSS and set the background color that way.

apps/site/src/app/schedule/components/EventRegular.tsx
```diff
@@ -32,25 +32,27 @@ interface EventMomentProps {
...
+		Main: string;
+		Workshop: string;
+		Social: string;
 	}
+
+	const eventBackgroundColors: EventBackgroundColors = {
+		Main: "bg-[#DFBA73]",
+		Workshop: "bg-[#94A9BD]",
+		Social: "bg-[#DFA9A9]",
+	};
```
samderanova, line 45: Nitpick: It might be simpler to use [string enums](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) instead of an interface. Additionally, we generally declare constant declarations like these outside of the function.

samderanova (changes requested):

apps/site/src/app/schedule/components/EventRegular.tsx
samderanova: On second thought, string enums can't be used the way I thought they could be used, so please disregard my comment about using the enum, but still move it outside of the function.

apps/site/package.json
waalbert: I wasn't able to get a Date object in PST using Day.js. From what I've tried, Day.js was able to print the date in PST but when the dayjs object gets converted into a JS Date object, the Date object is still in UTC time.

samderanova (approved):
Looking good for now, approving to unblock #185.

taesungh (approved):
Approving to unblock, we can revisit cleaning up some of the type definitions.

apps/site/package.json
taesungh: Will revisit date processing.

PR #164, opened 2023-12-29
fix/Query parameter for application preface enforced

IanWearsHat:
if query parameters match, but the user is not logged in, they will be redirected to log in

### DIFF
--- a/apps/site/src/app/apply/page.tsx
+++ b/apps/site/src/app/apply/page.tsx
@@ -20,55 +20,53 @@ export default async function Apply({
 		prefaceAccepted?: string;
 	};
 }) {
+	const hasAcceptedQueryParam =
+		searchParams !== undefined && searchParams?.prefaceAccepted === "true";
+
 	const identity = await getUserIdentity();
+
 	if (identity.status !== null) {
 		redirect("/portal");
 	}
 
-	if (
-		searchParams !== undefined &&
-		searchParams?.prefaceAccepted === "true" &&
-		identity.uid === null
-	) {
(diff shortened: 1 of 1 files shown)


apps/site/src/app/apply/page.tsx
```diff
@@ -25,6 +25,14 @@ export default async function Apply({
...
+		searchParams !== undefined &&
+		searchParams?.prefaceAccepted === "true" &&
+		identity.uid === null
+	) {
+		redirect("/login");
+	}
+
 	const applyBody =
 		searchParams !== undefined &&
 		searchParams?.prefaceAccepted === "true" ? (
```
taesungh, line 38: Chore: We're repeating the same conditional logic for checking the `prefaceAccepted` state. Could we extract this into a variable we can reference in both spots?

apps/site/src/app/apply/page.tsx
```diff
@@ -20,47 +20,53 @@ export default async function Apply({
 		prefaceAccepted?: string;
 	};
 }) {
+	const hasAcceptedQueryParam =
+		searchParams !== undefined && searchParams?.prefaceAccepted === "true";
```
taesungh, line 24: Partly out of scope, but since we're using optional chaining, is the first check for `undefined` actually necessary?

IanWearsHat: You're right, the first check is not necessary.

taesungh (approved):
Thanks for resolving this.

samderanova (approved):

samderanova:

Issue #165, opened 2023-12-30
API deployment encounters ImportModuleError due to missing shared object


taesungh:
Some instances of the Vercel Serverless Function for the API were unable to start from an AWS Lambda error
```python
[ERROR] Runtime.ImportModuleError: Unable to import module 'vc__handler__python': libexslt.so.0: cannot open shared object file: No such file or directory
```

taesungh:
This issue is present on all (preview) deployments after around midday (12–2pm) today.
I am suspecting the culprit is the [release of lxml 5.0.0](https://github.com/lxml/lxml/releases) today.

PR #172, opened 2023-12-30
fix: copy additional shared objects for vercel-lib dependencies

samderanova:
Attempted resolution of #165 . Adding `libxslt-devel` as a package to install in `vercel-lib.sh` might resolve the issue.

## Update

Adding `libxslt-devel`, `libgcrypt-devel`, and `libgpg-error-devel` and then adding the missing files to the list of `.so` files to copy over in `vercel-lib.sh` resolved this issue. This was discovered mainly through trial and error, in which I installed the first package, then noticed that a different `.so` file was missing so I installed the package for that as well. This continued until no more missing `.so` file errors appeared in Vercel logs (i.e. after adding `libgpg-error-devel`).

### DIFF
--- a/apps/site/vercel-lib.sh
+++ b/apps/site/vercel-lib.sh
@@ -13,6 +13,9 @@ lib_files=(
    libxmlsec1.so.1.2.20
    libxslt.so.1
    libxslt.so.1.1.28
+   libexslt.so.0
+   libgcrypt.so.11
+   libgpg-error.so.0
 )
 
 mkdir lib
--- a/apps/site/vercel-lib.sh
+++ b/apps/site/vercel-lib.sh
@@ -1,7 +1,7 @@
 # Install python3-saml dependencies on Vercel's Serverless Function 
 # environment (Amazon Linux)
 
-yum install -y libxml2-devel xmlsec1-devel xmlsec1-openssl-devel libtool-ltdl-devel libxslt-devel libgcrypt-devel
+yum install -y libxml2-devel xmlsec1-devel xmlsec1-openssl-devel libtool-ltdl-devel libxslt-devel libgcrypt-devel libgpg-error-devel
 lib_files=(
    libltdl.so.7
    libltdl.so.7.3.0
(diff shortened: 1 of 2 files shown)


apps/site/vercel-lib.sh
```diff
@@ -1,7 +1,7 @@
 # Install python3-saml dependencies on Vercel's Serverless Function 
 # environment (Amazon Linux)
 
-yum install -y libxml2-devel xmlsec1-devel xmlsec1-openssl-devel libtool-ltdl-devel
+yum install -y libxml2-devel xmlsec1-devel xmlsec1-openssl-devel libtool-ltdl-devel libxslt-devel libgcrypt-devel libgpg-error-devel
```
taesungh, line 4: Thought: the extra libraries added should already be transitive dependencies of the ones we need, so I don't think we need to manually specify them.

apps/site/vercel-lib.sh
```diff
@@ -13,6 +13,9 @@ lib_files=(
    libxmlsec1.so.1.2.20
    libxslt.so.1
    libxslt.so.1.1.28
+   libexslt.so.0
+   libgcrypt.so.11
+   libgpg-error.so.0
```
taesungh, line 18: Thought: I'm confused how this resolves the issue (or why we're even encountering the issue) when each of these are symlinks to the actual library file with the full version number (e.g. `libexslt.so.0 -> libexslt.so.0.8.17`), but if it works, so be it 🤷.

apps/site/vercel-lib.sh
taesungh: Chore [partly out of scope]: while you're at it, I think `libtool-ltdl-devel` is also a transitive dependency and can be removed.

apps/site/vercel-lib.sh
taesungh: On the surface, it appears the libraries aren't needed, but let's include the actual library files anyway to avoid potential issues down the line. See the lines above for how we specify the symlink and the actual library file.

taesungh (approved):
Weird how the latter two are already part of the AL2 image but not part of an actual deployed AWS Lambda Function. Thanks for putting up with all of the quirks and finding a suitable fix.

waalbert (approved):
LGTM

PR #174, opened 2023-12-30
feat: admin routes

### LINKED ISSUE #173: Import admin routes from last year's website
In order for applications to be able to effectively and efficiently view the applications to last year's IrvineHacks, we set up an administration portal for them on the frontend. This issue involves setting up the infrastructure for that portal by importing the backend routes it relies on.

samderanova:
Resolves #173 .

### DIFF
--- a/apps/api/src/auth/authorization.py
+++ b/apps/api/src/auth/authorization.py
@@ -1,0 +1,25 @@
+from typing import Any, Callable, Coroutine, Sequence
+
+from fastapi import Depends, HTTPException, status
+
+from auth.user_identity import User, require_user_identity
+from services import mongodb_handler
+from services.mongodb_handler import Collection
+from utils.user_record import Role, UserRecord
+
+
+def require_role(
+    allowed_roles: Sequence[Role],
+) -> Callable[[User], Coroutine[Any, Any, User]]:
+    """Return a dependency which requires a user to have an allowed role."""
+
+    async def require_allowed_role(user: User = Depends(require_user_identity)) -> User:
+        """Require a user to have a role in the allowed roles."""
+        record = await mongodb_handler.retrieve_one(Collection.USERS, {"_id": user.uid})
+        user_record = UserRecord.model_validate(record)
+
+        if user_record.role not in allowed_roles:
+            raise HTTPException(status.HTTP_403_FORBIDDEN)
+        return user
+
+    return require_allowed_role
--- a/apps/api/src/routers/admin.py
+++ b/apps/api/src/routers/admin.py
@@ -22,7 +22,7 @@ ADMIN_ROLES = (Role.DIRECTOR, Role.REVIEWER)
 class ApplicationDataSummary(BaseModel):
     first_name: str
     last_name: str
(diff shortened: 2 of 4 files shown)


taesungh (commented):
Could we also get the unit tests for the admin routes?

apps/api/src/routers/admin.py
```diff
@@ -0,0 +1,112 @@
...
+
+router = APIRouter()
+
+ADMIN_ROLES = (Role.DIRECTOR, Role.REVIEWER)
+
+
+class ApplicationDataSummary(BaseModel):
+    first_name: str
+    last_name: str
+    university: str
```
taesungh, line 25: Issue: we changed the name of this field in `ApplicationData` from `university` to `school`, so we'll need to use the new name.

apps/api/src/routers/admin.py
```diff
@@ -0,0 +1,112 @@
...
+    try:
+        return TypeAdapter(list[ApplicantSummary]).validate_python(records)
+    except ValidationError:
+        raise RuntimeError("Could not parse applicant data.")
+
+
+@router.get("/applicant/{uid}")
+async def applicant(
+    uid: str,
+) -> Applicant:
```
taesungh, line 69: Issue: this endpoint does not enforce admin authorization and could be accessed by anyone. We'll need to include the `require_role` dependency.

apps/api/src/auth/authorization.py
```diff
@@ -0,0 +1,25 @@
...
+) -> Callable[[User], Coroutine[Any, Any, User]]:
+    """Return a dependency which requires a user to have an allowed role."""
+
+    async def require_allowed_role(user: User = Depends(require_user_identity)) -> User:
+        """Require a user to have a role in the allowed roles."""
+        record = await mongodb_handler.retrieve_one(Collection.USERS, {"_id": user.uid})
+        user_record = UserRecord.model_validate(record)
+
+        if user_record.role not in allowed_roles:
+            raise HTTPException(status.HTTP_403_FORBIDDEN)
```
taesungh, line 22: Issue [non-blocking]: although unlikely, it is possible for a user without a database record to attempt to access this endpoint, and Pydantic will raise `ValidationError`. This may however become irrelevant if we resolve #179.

apps/api/src/routers/admin.py
taesungh: Partly my fault from last year: we can [Add `dependencies` to the _path operation decorator_](https://fastapi.tiangolo.com/tutorial/dependencies/dependencies-in-path-operation-decorators/#add-dependencies-to-the-path-operation-decorator) instead of as a function parameter to avoid confusion that the parameter is unused.

taesungh (approved):
Thanks for adding.

waalbert (approved):
LGTM

apps/api/src/auth/authorization.py
samderanova: We can address this in a separate issue. Merging to unblock other issues related to the admin portal.

PR #175, opened 2023-12-30
fix/Redirects to portal if guest logged in

IanWearsHat:
If the `/guest-login` route is accessed while the user is logged in, they will be redirected to `/portal`.


taesungh (approved):

samderanova (approved):
Thanks for working on this!

PR #181, opened 2024-01-01
fix: remove call to `_remove_guest_key`

### LINKED ISSUE #178: [Guest Auth] `last_login` field is incorrect when removing expired keys
The `_remove_guest_key` function stores the last login time, but this function is also called when a key has expired in `_get_existing_key` which should not be considered a successful login.

samderanova:
Resolves #178 by removing the call to `_remove_guest_key` in the `_get_existing_key` function.

### DIFF
--- a/apps/api/src/auth/guest_auth.py
+++ b/apps/api/src/auth/guest_auth.py
@@ -88,6 +88,9 @@ async def _get_existing_key(email: EmailStr) -> Optional[str]:
     # Reject expired key
     now = utc_now()
     if now > auth.exp:
+        await mongodb_handler.update_one(
+            Collection.USERS, {"_id": uid}, {"guest_auth": None}
+        )
         return None
 
     return auth.key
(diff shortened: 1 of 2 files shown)


apps/api/src/auth/guest_auth.py
```diff
@@ -88,7 +88,9 @@ async def _get_existing_key(email: EmailStr) -> Optional[str]:
     # Reject expired key
     now = utc_now()
     if now > auth.exp:
-        await _remove_guest_key(uid)
+        await mongodb_handler.update_one(
+            Collection.USERS, {"_id": uid}, {"guest_auth": None}
+        )
```
taesungh, line 93: Could we extract this database call to another named function such as `_remove_expired_guest_key`?

taesungh (approved):
I think we have a separate issue with the guest confirmation cookie expiring, but the fix here is still relevant. Thanks for handling this.

waalbert (approved):
LGTM

Issue #182, opened 2024-01-01
[Guest Auth] Token removed even when verification fails


taesungh:
Following up from #147 and #148, a user could mistype/paste a well-formed but incorrect passphrase. From #148, we remove the guest key unconditionally, so even a correct passphrase will be rejected afterwards.
- Remove the consumed guest key only when verification succeeds
- Update the docstring for `verify_guest_credentials`

PR #183, opened 2024-01-02
Guest auth token is now removed only after successful verification

samderanova:
To resolve the issue pointed out in #182, we should remove the token only when the call to the `_validate` function returns `True`, or in other words, if the passphrase and confirmation are valid.

### DIFF
--- a/apps/api/tests/test_guest.py
+++ b/apps/api/tests/test_guest.py
@@ -109,9 +109,7 @@ def test_successful_guest_verification_provides_identity(
 
 
 @patch("auth.guest_auth._get_existing_key", autospec=True)
-@patch("auth.guest_auth._remove_guest_key", autospec=True)
 def test_invalid_guest_verification_is_unauthorized(
-    mock_remove_guest_key: AsyncMock,
     mock_get_existing_key: AsyncMock,
 ) -> None:
     """Test that a guest with invalid credentials is unauthorized."""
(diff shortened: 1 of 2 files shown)


apps/api/tests/test_guest.py
```diff
@@ -124,5 +122,3 @@ def test_invalid_guest_verification_is_unauthorized(
     )
 
     assert res.status_code == 401
-
-    mock_remove_guest_key.assert_awaited_once_with("edu.caltech.beaver")
```
taesungh, line 128: Chore: We can use [`AsyncMock.assert_not_awaited`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.AsyncMock.assert_not_awaited) to check that the guest key is not removed.

apps/api/tests/test_guest.py
```diff
@@ -109,9 +109,7 @@ def test_successful_guest_verification_provides_identity(
 
 
 @patch("auth.guest_auth._get_existing_key", autospec=True)
-@patch("auth.guest_auth._remove_guest_key", autospec=True)
 def test_invalid_guest_verification_is_unauthorized(
-    mock_remove_guest_key: AsyncMock,
     mock_get_existing_key: AsyncMock,
```
taesungh, line 113: Nitpick: also missed on #148, but for consistency, it would make more sense for `mock_get_existing_key` to be the first parameter (and switch the order of the patches). Refer to `test_successful_guest_verification_provides_identity` for the suggested order.

taesungh (approved):
Thanks for fixing this.

Issue #170, opened 2023-12-30
Login forms do not prevent resubmission


taesungh:
We had a user report of seeing the Unauthorized response from the API when attempting passphrase verification on the Guest Login page with the verification form.

According to the logs, the user successfully authenticated as guest via `/api/guest/verify` but then was unauthorized on the same endpoint two seconds later. I suspect this happened because the user pressed the submit button twice causing the browser to ignore the response from the original request which would have contained the user identity cookie.

As such, we should consider using some form of client-side submission to send the guest passphrase or at the very least disable the submit button before letting the browser process the form submission.

taesungh:
Actually, the regular Login page would also have the same issue. I think the best approach is updating the `ValidatingForm` component to include the submit button with the logic to disable on submit.

PR #186, opened 2024-01-04
Prevent accidental resubmission in `ValidatingForm`

Bl20052005:
added state that makes buttons / inputs unclickable after successful submission

### DIFF
--- a/apps/site/src/lib/components/ValidatingForm/ValidatingForm.module.scss
+++ b/apps/site/src/lib/components/ValidatingForm/ValidatingForm.module.scss
@@ -7,3 +7,9 @@
 		display: block;
 	}
 }
+
+.unclickable {
+	button, input {
+		pointer-events: none;
+	}
+}
--- a/apps/site/src/lib/components/ValidatingForm/ValidatingForm.module.scss
+++ b/apps/site/src/lib/components/ValidatingForm/ValidatingForm.module.scss
@@ -9,7 +9,8 @@
 }
 
 .unclickable {
-	button, input {
+	button,
+	input {
 		pointer-events: none;
 	}
 }
--- a/apps/site/src/lib/components/ValidatingForm/ValidatingForm.module.scss
+++ b/apps/site/src/lib/components/ValidatingForm/ValidatingForm.module.scss
@@ -7,10 +7,3 @@
 		display: block;
 	}
 }
-
-.unclickable {
-	button,
-	input {
-		pointer-events: none;
-	}
-}
--- a/apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
+++ b/apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
@@ -12,7 +12,9 @@ interface FormProps {
 
 function ValidatingForm(props: PropsWithChildren<FormProps>) {
 	const [validated, setValidated] = useState<boolean>(false);
-	const [isDisabled, setIsDisabled] = useState<boolean>(false);
+	const [submitting, setSumbitting] = useState<boolean>(false);
+
+	const { children, ...rest } = props;
 
 	const handleSubmit = (event: FormEvent<HTMLFormElement>): void => {
 		const form = event.currentTarget;
@@ -20,7 +22,7 @@ function ValidatingForm(props: PropsWithChildren<FormProps>) {
 			// prevent submission to display validation feedback
 			event.preventDefault();
 		} else {
-			setIsDisabled(true);
+			setSumbitting(true);
 		}
 		setValidated(true);
 	};
@@ -29,14 +31,12 @@ function ValidatingForm(props: PropsWithChildren<FormProps>) {
 		<form
 			onSubmit={handleSubmit}
 			noValidate // use custom validation feedback
-			className={`bg-white p-5 md:p-10 rounded-2xl mx-5 text-black ${
-				validated ? styles.validated : styles.notYetValidated
-			}`}
+			className={validated ? styles.validated : styles.notYetValidated}
 			// validated={validated}
-			{...props}
+			{...rest}
 		>
-			{props.children}
-			<Button text="Continue" disabled={isDisabled} />
+			{children}
+			<Button text="Continue" disabled={submitting} />
 		</form>
 	);
 }
--- a/apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
+++ b/apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
@@ -12,7 +12,7 @@ interface FormProps {
 
 function ValidatingForm(props: PropsWithChildren<FormProps>) {
 	const [validated, setValidated] = useState<boolean>(false);
-	const [isClickable, setIsClickable] = useState<boolean>(false);
+	const [isDisabled, setIsDisabled] = useState<boolean>(false);
 
 	const handleSubmit = (event: FormEvent<HTMLFormElement>): void => {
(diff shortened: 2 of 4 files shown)


samderanova (approved):
Thanks for working on this!

apps/site/src/lib/components/ValidatingForm/ValidatingForm.module.scss
```diff
@@ -7,3 +7,10 @@
 		display: block;
 	}
 }
+
+.unclickable {
+	button,
+	input {
+		pointer-events: none;
+	}
+}
```
taesungh, line 16: Issue: disabling pointer events does not necessarily disable interactions with the element (e.g. when focusing on the submit button and pressing enter).

apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
```diff
@@ -11,12 +11,15 @@ interface FormProps {
...
 	const [validated, setValidated] = useState<boolean>(false);
+	const [isClickable, setIsClickable] = useState<boolean>(true);
 
 	const handleSubmit = (event: FormEvent<HTMLFormElement>): void => {
 		const form = event.currentTarget;
 		if (!form.checkValidity()) {
 			// prevent submission to display validation feedback
 			event.preventDefault();
+		} else {
+			setIsClickable(false);
```
taesungh, line 22: Issue: when navigating back in the browser, the state is permanently disabled.

apps/site/src/lib/components/ValidatingForm/ValidatingForm.module.scss
samderanova: Thanks for pointing this out. I think maybe a better approach would be setting the `disabled` attribute of the button to false? We have access to the form element so we can access the button via the form's children attribute.

Bl20052005: how would we make sure that all the buttons in children are disabled though? Should we pass it as a prop and attach it to all the buttons?

apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
samderanova: I think this can also be resolved by the method I mentioned above. Also, I didn't run into this issue. Could this be a Safari thing?

apps/site/src/lib/components/ValidatingForm/ValidatingForm.module.scss
samderanova: The only button we really need to disable (at least for the login form) is the submit one. We can access the button via [`getElementsByTagName`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getElementsByTagName).

taesungh: As I mentioned in #170, we can move the submit button element to inside `ValidatingForm`

Bl20052005: but the form can also be submitted for example via pressing enter on the input field, would we need to have a fix for that as well?

taesungh: Hmm, we'll need to investigate more, some browsers will disable the form when it contains a button that is disabled, some browsers won't.

Bl20052005: there is also technically an "inert" attribute we can put that makes browsers ignore user input, but I don't know if that's appropriate here
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert

samderanova: We could also surround all of the form elements with a `fieldset` and then set the `disabled` attribute to `"disabled"`.

taesungh: I can't find it in the [Safari 16 Release Notes](https://developer.apple.com/documentation/safari-release-notes/safari-16-release-notes), but the issue appears to be fixed through WebKit/WebKit#3093.

Since most modern browsers will disable the form submission, I think we can proceed with the disabled button approach.

Bl20052005: slight issue: I'm not quite certain how to integrate button into ValidatingForm as the form's children are already passed in props and I can't edit the props or pass any more children into the form

Bl20052005: nevermind I think I figured it out

apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
```diff
@@ -25,10 +29,15 @@ function ValidatingForm(props: PropsWithChildren<FormProps>) {
 		<form
 			onSubmit={handleSubmit}
 			noValidate // use custom validation feedback
-			className={validated ? styles.validated : styles.notYetValidated}
+			className={`bg-white p-5 md:p-10 rounded-2xl mx-5 text-black ${
```
taesungh, line 32: Suggestion: styling should be the responsibility of the parent, not the form itself, which is why we had the styles on the consumers.

apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
```diff
@@ -25,10 +29,15 @@ function ValidatingForm(props: PropsWithChildren<FormProps>) {
...
+				validated ? styles.validated : styles.notYetValidated
+			}`}
 			// validated={validated}
 			{...props}
-		/>
+		>
+			{props.children}
```
taesungh, line 38: Thought: we are already spreading props, so providing children to the element might be in conflict.
We can extract the children separately and provide the rest of the props.
```tsx
const { children, ...rest } = props;
<form {...rest}>
	{children}
	<button ...>
</form>
```

apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
```diff
@@ -11,12 +12,15 @@ interface FormProps {
 
 function ValidatingForm(props: PropsWithChildren<FormProps>) {
 	const [validated, setValidated] = useState<boolean>(false);
+	const [isDisabled, setIsDisabled] = useState<boolean>(false);
```
taesungh, line 15: Suggestion/nitpick: I think `submitting, setSubmitting` would be more descriptive than `isDisabled`.

taesungh (approved):
Nice work following along with the suggestions and finding your way around the `ValidatingForm`.

apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
```diff
@@ -11,12 +12,17 @@ interface FormProps {
 
 function ValidatingForm(props: PropsWithChildren<FormProps>) {
 	const [validated, setValidated] = useState<boolean>(false);
+	const [submitting, setSumbitting] = useState<boolean>(false);
```
taesungh, line 15: Typo in `setSumbitting`, should be `setSubmitting`

taesungh (approved):

samderanova (approved):

samderanova:
Thanks for resolving all of the issues and apologies for my oversight. Please *squash* when ready.

PR #188, opened 2024-01-05
Setup/windows docker

### LINKED ISSUE #187: Create Windows version of Docker startup script
To allow Windows developers to run the MongoDB Docker containers, a `.bat` script should be created that contains the same commands as `run-docker.sh`.

samderanova:
Resolves #187 by creating a `.bat` script and using [`run-script-os`](https://www.npmjs.com/package/run-script-os) to run the appropriate startup script based on architecture. Assuming that most of us are on either Windows or a Unix-based operating system, `run-script-os` will execute `run-docker.bat` for those on Windows (but not developing directly on WSL)  when running `pnpm dev` and `run-docker.sh` for all others.

To test this, make sure to run `pnpm install` to install the `run-script-os` library and then run `pnpm dev` from the root of the project. If you start seeing output related to Docker or installing `pip` libraries, this means the Docker containers are being set up properly. Lastly, to make sure the applications are running properly, check `localhost:8000` (the port where the backend is hosted) and `localhost:8081` (the port where the frontend interface used to interact with the local Docker MongoDB instance is hosted).


waalbert (approved):
LGTM

Bl20052005 (approved):

Issue #185, opened 2024-01-04
Import admin console from last year's website


samderanova:
In relation to #173, this issue involves setting up the frontend of the admin console to allow the applications team to review applications properly.

Specifically, for now, we can directly import the admin console from last year's website, meaning that it will mainly be a client side application, with state and client side API queries. However, in the long term, we should eventually migrate this to server side rendering like the rest of our application. This is something I'd like the interns to work on and take their time with, in order for them to gain a better understanding of what SSR is and how it works.

taesungh:
Admin site mostly imported, will get #118 done first since waiting on #163 before submitting PR for #189 after which I can submit a PR for this.

PR #193, opened 2024-01-06
Add local-only dev route to simulate user identity

### LINKED ISSUE #118: Provide internal login endpoint for local development
Following up from #113, to locally test the application system as well as identity-based routing, the API could provide an internal development-only endpoint to issue a spoofed user identity.
- Create new `admin` router and include a `/dev-login` route to issue a user identity
  - This route will need to be restricted to be accessible only when developing locally and should not be accessible in any deployments
  - The route could have a query parameter or provided a static identity
  - Since the development server uses http, the user identity cookie will need to be provided with `secure=False`

taesungh:
Ahead of #185, adding a local-only development endpoint to simulate a super identity. This will be useful in more easily testing all portions of the site including the Apply page, Portal page, and Admin site.

Resolves #118.

## Changes
- Add dev router with `/impersonate` endpoint to simulate user identity
- Mount dev router to app only when `DEPLOYMENT` environment is `LOCAL`
  - Add environment variable to docker-compose file
- Include associated tests for dev router and mounting

## Testing
1. Run the API through Docker and the development server for the site (e.g. `pnpm run dev`)
2. Access the impersonation endpoint through the site proxy
    - e.g. `http://localhost:3000/api/dev/impersonate/boo`
3. Confirm a user identity is provided
    - Navbar should indicate the user is logged in (button should say **Log Out**)
    - If the associated user has submitted an application, the Portal page should be shown
    - Otherwise, the Application Preface should be shown

### DIFF
--- a/apps/api/src/routers/dev.py
+++ b/apps/api/src/routers/dev.py
@@ -1,0 +1,23 @@
+from fastapi import APIRouter
+from fastapi.responses import RedirectResponse
+
+from auth import user_identity
+from auth.user_identity import COOKIE_NAME, NativeUser
+
+router = APIRouter()
+
+
+@router.get("/impersonate/{ucinetid}")
+async def impersonate(ucinetid: str) -> RedirectResponse:
+    """Simulate a user identity during local development (does not require https)."""
+    user = NativeUser(
+        ucinetid=ucinetid,
+        display_name="Local Dev",
+        email="dev@irvinehacks.com",
+        affiliations=[],
+    )
+
+    res = RedirectResponse("/portal", status_code=303)
+    jwt_token = user_identity._generate_jwt_token(user)
+    res.set_cookie(COOKIE_NAME, jwt_token, max_age=4000, httponly=True)
+    return res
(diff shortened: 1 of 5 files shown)


apps/api/src/routers/dev.py
```diff
@@ -0,0 +1,23 @@
...
+async def impersonate(ucinetid: str) -> RedirectResponse:
+    """Simulate a user identity during local development (does not require https)."""
+    user = NativeUser(
+        ucinetid=ucinetid,
+        display_name="Local Dev",
+        email="dev@irvinehacks.com",
+        affiliations=[],
+    )
+
+    res = RedirectResponse("/portal", status_code=303)
```
samderanova, line 20: Note: One thing for everyone to note is that when you fill out the application and click submit, nothing seems to happen. In the console, there is an error that is thrown about how the SendGrid API key is missing, but fortunately, the application data is still submitted to the local database, which can be verified by refreshing the page and seeing the portal instead of the application.

samderanova (approved):
Thanks for getting this set up! This will help a lot when testing the admin portal.

apps/api/src/routers/dev.py
waalbert: This happens to me as well

waalbert (approved):
LGTM

PR #194, opened 2024-01-07
Move non-Admin page components into route group

### LINKED ISSUE #189: Address layout folder change due to admin portal
Because the admin portal has a completely different layout and design compared to the rest of the site, one of the steps we plan on taking is moving the current site's files to a subfolder so that the admin page does not adhere to the current site's layout (i.e. it will not share the same navbar and footer as other pages do).

taesungh:
During #185, we'll need to use a separate layout for the Admin site since we don't want to show the usual Navbar and Footer. In last year's pages router ([HackAtUCI/HackUCI-Site/src/pages/_app.tsx:44-46](https://github.com/HackAtUCI/HackUCI-Site/blob/0b181fb750094373e2ab3f689d2691312bb88ee7/src/pages/_app.tsx#L44-L46)), we were able to select a specific layout based on `router.pathname`. However, with App Router, `usePathname` is allowed only in a Client Component.
> Reading the current URL from a Server Component is not supported. This design is intentional to support layout state being preserved across page navigations. ([usePathname, Next.js API Reference](https://nextjs.org/docs/app/api-reference/functions/use-pathname))

Therefore, we'll need to use [Route Groups](https://nextjs.org/docs/app/building-your-application/routing/route-groups) for the main site to allow the Admin site to have its own layout. For now, the two share a plain [Root Layout](https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts#root-layout-required), but if needed, the individual layouts could each be Root Layouts.

Closes #189.
As this is a major refactor, I'll be planning for a fast-forward merge by pre-tagging the commit.

## Changes
- Since `usePathname` requires a Client Component and the Root Layout must be a Server Component, create a route group for the main site
- The Admin site will have it's own layout (without main Navbar/Footer)
- Move application form components into `@/lib/components/forms`
  - Update imports in application form sections accordingly
- Refactor `SceneLayout` for canvas scene
- Include brown background color on `body` for main layout


waalbert (approved):
LGTM

apps/site/src/components/dom/SceneLayout.tsx
```diff
@@ -37,4 +37,4 @@ const Layout = ({ children }: { children: ReactNode }) => {
 	);
 };
 
-export { Layout };
+export { SceneLayout };
```
samderanova, line 40: Nitpick [partially out of scope]: Since this is the only thing that's exported from this file, we should make `SceneLayout` a default export.

apps/site/src/components/dom/SceneLayout.tsx
```diff
@@ -3,7 +3,7 @@
 import { ReactNode, useRef } from "react";
 import Scene from "@/components/canvas/Scene";
 
-const Layout = ({ children }: { children: ReactNode }) => {
+const SceneLayout = ({ children }: { children: ReactNode }) => {
```
samderanova, line 6: Nitpick: Since you changed the other instances of `ReactNode` to `PropsWithChildren`, should we do it here too?

samderanova (changes requested):
Looking good, a few suggestions:

samderanova:
Also, can we not just squash merge here? According to the [GitHub docs](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/about-merge-methods-on-github#squashing-your-merge-commits):

> When you select the Squash and merge option on a pull request on GitHub.com, the pull request's commits are squashed into a single commit. Instead of seeing all of a contributor's individual commits from a topic branch, the commits are combined into one commit and merged into the default branch. Pull requests with squashed commits are merged using the [fast-forward option](https://git-scm.com/docs/git-merge#_fast_forward_merge).

samderanova:
Also, for other people reviewing this PR and who are unfamiliar with fast-forwarding and pre-tagging, could you explain what both of those are?

apps/site/src/components/dom/SceneLayout.tsx
taesungh: Kinda feeling (in general) that having named exports helps preserve references, but that's also out of scope.

taesungh:
Squashing isn't entirely safe since we need to ensure this refactor is the latest commit on top of main. Otherwise, other commits may have added new files to the old locations which would not be seen by rebasing or squashing and would not be detected as a merge conflict. Fast-forwarding allows the commit in this PR to be pushed exactly as is without any modification. To maintain the usual PR number added to the commit message, the PR number has already been included in the message of the commit to be fast-forwarded..

apps/site/src/components/dom/SceneLayout.tsx
taesungh: Uhh, the `Three.tsx` file also uses `ReactNode`, so I didn't bother messing with it too much.

samderanova (approved):

waalbert (approved):

PR #195, opened 2024-01-08
Import Admin site from last year

### LINKED ISSUE #185: Import admin console from last year's website

taesungh:
Importing last year's Admin site to use again this year.

This PR focuses on providing a minimal working version of the Admin site to view applications and submit reviews based on last year's version. Please limit review comments to functional issues, code style (for modifications after the initial import), and relevant questions. We can discuss design and feature improvements separately.

To keep things simple, we are keeping the original client-side data fetching hooks and user identity context for now, but the pages and navigation were updated to use [App Router](https://nextjs.org/docs/app) instead of [Pages Router](https://nextjs.org/docs/pages). While the entire PR is quite lengthy, the first commit is a raw import, and **the remaining changes should be viewed separately**: [Compare abbfd47..a6480f0](https://github.com/HackAtUCI/irvinehacks-site-2024/compare/abbfd471335040288f51dcaaf028de3621e933e1..a6480f0cc3dee3ba79d0cf87a84db7abd618e270).

Resolves #185.

## Changes
- Import last year's Admin site components as-is
  - Import Admin site components of [HackAtUCI/HackUCI-Site](https://github.com/HackAtUCI/HackUCI-Site) exactly as-is
    - Includes page views, subcomponents, hooks, and layout
  - Organize components into approximate locations but without changes
- Install cloudscape-design and swr
  - Install cloudscape-design components and global-styles for Admin site
  - Install swr for client-side data fetching on Admin site
- Fix first-party imports of Admin components
- Fix missing `OptionDefinition` type
  - Extract `MultiselectProps.Option` to use instead of `OptionDefinition`
- Use `UserContext` for Admin site components
  - For now, we are using client-side data fetching, so the user identity cannot be fetched by the server and must be provided through context
- Use dynamic segment params for Applicant UID
  - Replace router with props for dynamic route segment in `Applicant`
- Expose Admin site pages for Applicant(s)
  - Add default exports to `page.tsx` to serve `Applicant`/`Applicants`
  - Include `use client` directive since fetching data from client by hook
- Provide identity through context in Admin layout
  - Add `useUserIdentity` hook for client-side identity on Admin site
  - Provide `UserContext` with identity in layout for Admin site
- Migrate Admin navigation to use App Router hooks
  - Replace old `next/router` hooks with new hooks from `next/navigation`
  - Redesign `followWithNextLink` into hook since `useRouter` is a hook
  - Replace state and effect for `breadcrumbItems` with regular function
- Use Server Component for Admin Layout
  - Metadata cannot be provided from a Client Component
    - Similar reasoning, if we wanted the Admin Layout to be a [Root Layout](https://nextjs.org/docs/app/building-your-application/routing/pages-and-layouts#root-layout-required)
  - Transfer `UserContext.Provider` to `AdminLayout` which is Client
  - Simplify redirects when not logged in or unauthorized
- Fix `CardHeader` links in Admin `Applicants` view
  - Using `passHref` and `legacyBehavior` does not work with Cloudscape
  - Use Cloudscape `Link` with custom `followWithNextLink` router
  - Simplify event type for `BaseNavigationDetail`
- Update hackathon name in Admin site layout
  - Replace "HackUCI 2023" with "IrvineHacks 2024"
- Update `ApplicationData` fields in Admin site
  - Change `university` to `school` and remove `gender`
  - Update profile links and FRQ question field names
- Serve placeholder page for `AdminDashboard`
  - Remove old `AdminDemo` page
- Reformat Admin site: include all trailing commas
- Fix type error with `fetcher` in `useApplicant`
  - Specify explicit `SWRKey` generic type

## Testing
1. Start the development servers for the site and API
2. Submit some sample applications under various UIDs
3. Add the "reviewer" role to your own user record
4. Acquire a cookie for your own user identity
5. Visit the Admin site at `/admin/dashboard`
6. Navigate to the **Applicants** view from the sidebar
7. Observe the list of applicants is shown
8. Select an applicant by name
9. Observe the application data is shown


apps/site/src/app/admin/applicants/[uid]/components/ApplicationReviews.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+function ApplicationReviews({ reviews }: ApplicationReviewsProps) {
+	const { uid } = useContext(UserContext);
+
+	if (reviews.length === 0) {
+		return <p>-</p>;
+	}
+
+	const formatUid = (uid: uid) => uid.split(".").at(-1);
```
Comment (samderanova) [inline apps/site/src/app/admin/applicants/[uid]/components/ApplicationReviews.tsx:18]: Nitpick: The `uid` variable and the `uid` type may be a bit confusing, at least to me. Should there be some sort distinction made between the two (perhaps in name)?

apps/site/src/app/admin/applicants/[uid]/components/ApplicationReviews.tsx
```diff
@@ -0,0 +1,43 @@
...
+	const formatDate = (timestamp: string) =>
+		new Date(timestamp).toLocaleDateString();
```
Comment (samderanova) [inline apps/site/src/app/admin/applicants/[uid]/components/ApplicationReviews.tsx:20]: Note for later: Same as with submission, it might also be good to include the time a review was made instead of just the date.

apps/site/src/app/admin/applicants/[uid]/components/ApplicantOverview.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+interface ApplicantOverviewProps {
+	applicant: Applicant;
+}
+
+function ApplicantOverview({ applicant }: ApplicantOverviewProps) {
+	const { application_data, status } = applicant;
+	const { submission_time, reviews } = application_data;
+
+	const submittedDate = new Date(submission_time).toDateString();
```
Comment (samderanova) [inline apps/site/src/app/admin/applicants/[uid]/components/ApplicantOverview.tsx:19]: Note for later: Might be good to include the time of submission instead of just the date.

samderanova (changes requested):
Thanks for working on this! Testing works as expected. A few notes and nitpicks:

apps/site/src/app/admin/applicants/Applicants.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+	const filteredApplicants = applicantList.filter(
+		(applicant) =>
+			(selectedStatuses.length === 0 ||
+				selectedStatusValues.includes(applicant.status)) &&
+			(selectedDecisions.length === 0 ||
+				selectedDecisionValues.includes(applicant.decision || "-")),
+	);
+
+	const items = filteredApplicants;
```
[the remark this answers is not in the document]
taesungh: Hmm, seems to be an artifact of HackAtUCI/HackUCI-Site@f7e127f: there was previously two stages of filtering. I agree it is a little redundant.

apps/site/src/app/admin/applicants/[uid]/components/ApplicationReviews.tsx
Comment (taesungh) [reply apps/site/src/app/admin/applicants/[uid]/components/ApplicationReviews.tsx:18]: Convention is type names are usually capitalized, not sure why I didn't do that before.

apps/site/src/lib/admin/useApplicant.ts
```diff
@@ -0,0 +1,79 @@
...
+export type Status = ReviewStatus | Decision;
+
+export interface Applicant {
+	_id: Uid;
+	role: string;
+	status: Status;
+	application_data: ApplicationData;
+}
+
+const fetcher = async ([api, uid]: [string, string]) => {
```
samderanova, line 53: Nitpick: Should we use the `Uid` type here instead of `string`?

samderanova (changes requested):

apps/site/src/lib/admin/useApplicant.ts
taesungh: Used to be `string[]` before the type fix, but makes sense since we're switching to tuple type.

samderanova (approved):

waalbert (approved):

PR #202, opened 2024-01-10
Redirect to /login on 401 in admin console and remove unrelated unused import

### LINKED ISSUE #201: [Admin site] Redirect to Login page when Admin data hooks encounter status 401
If an admin's authorization cookie expires, the data hooks for applicant(s) will fail with status 401, and SWR will continue retrying. To instruct the admin user to reauthenticate, we can handle the error and redirect to the Login page.

To avoid code duplication in the both the fetchers and mutations for both hooks, we can use a [Global Configuration](https://swr.vercel.app/docs/global-configuration) by providing `SWRConfig` in `AdminLayout` where the configuration has an `onError` handler to process the error with `axios.isAxiosError` (see [`Form.tsx:57:61`](https://github.com/HackAtUCI/irvinehacks-site-2024/blob/68bd937cebf5f52cbaf2542fabd1c93b19633243/apps/site/src/app/(main)/apply/sections/Form/Form.tsx#L57-L61) for reference).

samderanova:
Resolves #201. Use [`SWRConfig`](https://swr.vercel.app/docs/error-handling#global-error-report) to provide a global error handler when receiving an `AxiosError`.

Using the `redirect` function from `next/navigation` to redirect to `/login` resulted in an error message that simply states `NEXT_REDIRECT`, which upon further investigation, can only be used in server components, not client ones. Thus, I switched to `useRouter` from `next/navigation` and replaced each use of `redirect` with `useRouter`, resolving the error.

I also included a very minor out of scope change in this PR to remove an unused import. This was previously causing ESLint checks to fail, although deployments still worked properly because that "failure" was only an error.

To test that the unused import warning no longer works:
1. Run `pnpm build` in the root directory of the project or in the `site` directory.

To test that the redirect works properly on local:
1. Impersonate yourself by accessing `/api/dev/impersonate/[some identifier]`
2. Submit an application
3. In the MongoDB UI (http://localhost:8081, add the `reviewer` role to the the document associated with your application. This will grant you access to the admin console locally.
4. Access the admin console and go to the page with the list of applicants. Then delete your authentication cookie and click on one of the applications (perhaps your own). An error should show in the console and you should be redirected to the login page.

### DIFF
--- a/apps/site/src/app/admin/layout/AdminLayout.tsx
+++ b/apps/site/src/app/admin/layout/AdminLayout.tsx
@@ -41,7 +41,7 @@ function AdminLayout({ children }: PropsWithChildren) {
 	return (
 		<SWRConfig
 			value={{
-				onError: (err, _) => {
+				onError: (err) => {
 					if (axios.isAxiosError(err)) {
 						router.push("/login");
 					}
--- a/apps/site/src/app/admin/layout/AdminLayout.tsx
+++ b/apps/site/src/app/admin/layout/AdminLayout.tsx
@@ -41,10 +41,12 @@ function AdminLayout({ children }: PropsWithChildren) {
 	return (
 		<SWRConfig
 			value={{
-				onError: (err) => {
+				shouldRetryOnError: (err) => {
 					if (axios.isAxiosError(err) && err.response?.status === 401) {
 						router.push("/login");
+						return false;
 					}
+					return true;
 				},
 			}}
 		>
--- a/apps/site/src/app/admin/layout/AdminLayout.tsx
+++ b/apps/site/src/app/admin/layout/AdminLayout.tsx
@@ -1,8 +1,10 @@
 "use client";
 
-import { redirect } from "next/navigation";
+import axios from "axios";
+import { useRouter } from "next/navigation";
 
 import { PropsWithChildren } from "react";
+import { SWRConfig } from "swr";
 
 import AppLayout from "@cloudscape-design/components/app-layout";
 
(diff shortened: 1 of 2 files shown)


apps/site/src/app/admin/layout/AdminLayout.tsx
```diff
@@ -30,19 +33,29 @@ function AdminLayout({ children }: PropsWithChildren) {
...
-				breadcrumbs={<Breadcrumbs />}
-			/>
-		</UserContext.Provider>
+		<SWRConfig
+			value={{
+				onError: (err) => {
+					if (axios.isAxiosError(err)) {
+						router.push("/login");
+					}
+				},
```
taesungh, line 48: Issue: we don't necessarily want to redirect to Login on every API request error, e.g. if the server has a hiccup, we should let SWR retry. The Login redirect would make sense for only status 401.
Thought: sorry for the mixup, but I think specifying a function for `shouldRetryOnError` would be a little more appropriate to prevent retries even though the redirect should theoretically happen immediately.

apps/site/src/app/admin/layout/AdminLayout.tsx
```diff
@@ -30,19 +33,31 @@ function AdminLayout({ children }: PropsWithChildren) {
 	const authorized = isAdminRole(role);
 
 	if (!loggedIn) {
-		redirect("/login");
+		router.push("/login");
 	} else if (!authorized) {
-		redirect("/unauthorized");
+		router.push("/unauthorized");
```
taesungh, line 38: Thought: Kind of surprised the server-side redirect didn't cause any issues on the deployment.

taesungh (approved):
Thanks for fixing. Not sure what the best course of action is for other API errors, but retry by default seems okay for now.

waalbert (approved):

PR #203, opened 2024-01-11
Close applications page and endpoint after deadline

### LINKED ISSUE #199: Close off apply routes after application deadline
Applications close on January 14 at 11:59PM PST, so we'll need to make the apply routes on the frontend and the backend inaccessible to users. The main tasks for this issue are:

- Changing the "Apply" button on the home screen to a disabled button saying "Applications closed", similar to the ZotHacks site.
- On the apply page of the frontend, perform a hardcoded date check to see whether applications have closed. If they are closed, show a card similar to the application preface stating that applications are closed. See last year's [application closed message](https://github.com/HackAtUCI/HackUCI-Site/blob/main/src/views/apply/Apply.tsx) for an example application closed message.
- For the `/apply` route of the backend, a similar date check should be performed. If applications are closed, a `403 Forbidden` error should be raised with an additional argument stating that applications have been closed.

IanWearsHat:
Resolves #199.

- Created function to check whether the deadline has passed using js dates.
- Hardcoded datetime in utc in `/apply` route in `routers/user.py`
  - Checks whether or not the deadline has passed using datetime

If deadline has passed:
- The apply link will be removed from the navbar.
- Accessing the application page will show an "applications closed" card.
- A `403 FORBIDDEN` will be raised in `/apply` route with an "applications closed" message.
- The center apply button in the center gets disabled and reads, "Applications have been closed!"


apps/api/src/routers/user.py
```diff
@@ -63,6 +70,15 @@ async def apply(
     raw_application_data: Annotated[RawApplicationData, Depends(RawApplicationData)],
     resume: Optional[UploadFile] = None,
 ) -> str:
+    # Check if current datetime is past application deadline
+    now = datetime.now(timezone.utc)
```
samderanova, line 74: Issue: There's a `now` variable in this function and in the new function you created above. It might be simpler to add a time parameter to that new function and pass the now variable in so we don't have redundant variables.

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -34,6 +35,9 @@ function Navbar({ identity }: NavbarProps) {
 		window.addEventListener("scroll", scrollHandler);
 	}, []);
 
+	const deadlineSeconds = 1705305540;
```
samderanova, line 38: Issue: Can we move this to a utility function (maybe in the `lib/` directory) to reduce the repetition across the two files?

samderanova (changes requested):
Nice work so far, a few suggestions

apps/api/src/routers/user.py
```diff
@@ -25,6 +25,13 @@ class IdentityResponse(BaseModel):
     role: Union[Role, None] = None
 
 
+def getHasDeadlinePassed() -> bool:
```
taesungh, line 28: Chore: Python names are usually snake_case. Additionally, including `get` is unnecessary, and if this function isn't used outside of this module, we can denote it as private: consider `_is_past_deadline` (and update the test mocks accordingly).

apps/api/src/routers/user.py
```diff
@@ -25,6 +25,13 @@ class IdentityResponse(BaseModel):
...
+    now = datetime.now(timezone.utc)
+    deadline = datetime(2024, 1, 15, 7, 59, tzinfo=timezone.utc)
```
taesungh, line 30: Chore: Since this deadline is a constant, could we extract it to maybe between `router` and `IdentityResponse`?

apps/api/src/routers/user.py
```diff
@@ -63,6 +70,15 @@ async def apply(
     raw_application_data: Annotated[RawApplicationData, Depends(RawApplicationData)],
     resume: Optional[UploadFile] = None,
 ) -> str:
+    # Check if current datetime is past application deadline
+    now = datetime.now(timezone.utc)
+
+    if getHasDeadlinePassed():
+        raise HTTPException(
+            status.HTTP_403_FORBIDDEN,
+            "Applications have been closed."
```
taesungh, line 79: Nitpick [non-blocking]: double perfect `been` is unnecessary

apps/site/src/lib/components/Navbar/Navbar.tsx
taesungh: Or alternatively, I think we could export the utility function from the Apply page.

apps/api/tests/test_user_apply.py
```diff
@@ -272,3 +278,12 @@ def test_application_data_with_other_throws_422(
...
     assert res.status_code == 422
+
+
+@patch("routers.user.getHasDeadlinePassed", autospec=True)
+def test_past_deadline_throws_403(
+    mock_hasDeadlinePassed: Mock,
+) -> None:
+    mock_hasDeadlinePassed.return_value = True
+    res = client.post("/apply", data=SAMPLE_APPLICATION, files=SAMPLE_FILES)
+    assert res.status_code == 403
```
taesungh, line 289: Praise: great job figuring out how to use the patches and mocks.

apps/site/src/app/(main)/apply/sections/ApplicationsClosed/ApplicationsClosed.tsx
```diff
@@ -0,0 +1,28 @@
+import Link from "next/link";
+
+function ApplicationsClosed() {
+	return (
+		<div className="rounded-2xl p-10 text-display text-[#432810] bg-white text-center w-8/12 max-[800px]:w-9/12 max-[400px]:w-11/12">
+			<h1 className="text-4xl p-10">
+				<p>Applications for IrvineHacks 2024 were closed on January 14th.</p>
```
taesungh, line 7: Nitpick: passive voice unnecessary, and paragraph unnecessary since already `h1`.
```suggestion
				Applications for IrvineHacks 2024 closed on January 14th.
```

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -76,10 +76,14 @@ function Navbar({ identity }: NavbarProps) {
...
-					{!status && <NavLinkItem href="/apply">Apply</NavLinkItem>}
-					{status !== null && (
-						<NavLinkItem href="/portal">Portal</NavLinkItem>
-					)}
+					{!status &&
+						(deadlinePassed ? (
+							<Button text="Applications closed" disabled />
+						) : (
+							<NavLinkItem href="/apply">Apply</NavLinkItem>
+						))}
```
taesungh, line 84: Thought: not sure if having a button in the middle of the Navigation is the best design. Maybe we could just remove the Apply link entirely? Might want to consult graphics as well.

taesungh (changes requested):
Nice work on this, some small suggestions on naming, grammar, and code organization, and one concern about the Navigation link.

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -76,10 +75,18 @@ function Navbar({ identity }: NavbarProps) {
...
-					)}
+					{!status &&
+						(deadlinePassed ? (
+							<p
+								className={clsx(styles.navMenuLink, styles.disabledLink, "m-0")}
+							>
+								Applications have closed
+							</p>
+						) : (
+							<NavLinkItem href="/apply">Apply</NavLinkItem>
```
samderanova, line 86: Issue: I think we can just not show the `Apply` link when the deadline has passed, mainly because now, this element isn't a link (and therefore shouldn't be in the navbar).

samderanova (changes requested):
One more small change:

apps/api/src/routers/user.py
```diff
@@ -18,13 +18,19 @@
 
 router = APIRouter()
 
+deadline = datetime(2024, 1, 15, 7, 59, tzinfo=timezone.utc)
```
taesungh, line 21: Nitpick [non-blocking]: we generally put constant names in UPPERCASE.

apps/site/src/lib/components/Navbar/Navbar.module.scss
```diff
@@ -32,3 +32,11 @@
...
 	}
 }
+
+.disabledLink {
+	color: #d1d1d1;
+	text-shadow: none;
+	padding-bottom: 0;
+
+	pointer-events: none;
+}
```
taesungh, line 42: Chore: unused style

taesungh (approved):
Thanks for working on this. We can merge this for now to avoid further increasing scope, but one lingering issue we'll need to address separately: if somebody has the application page open while the deadline closes, the client-side submission will fail without any indication of why.

samderanova (approved):

PR #204, opened 2024-01-11
Feature/native user records

### LINKED ISSUE #179: [API] Store user records for native users
Currently, native user records (for UCI students) are created only when the user submits an application. For increased participant retention, we can store user records in the database when native users first log in (this is already done for guest users). This will allow us to send a notice of "incomplete application" closer to the deadline to potential participants who have logged in but not submitted an application.
- During SAML ACS, upsert a user record with the last login time, similar to guest authentication
  - There might be a way to improve the abstraction organization among `guest-auth` and `user_identity`, but each consideration has its own trade-offs

samderanova:
Resolves #179. This pull request creates a helper function in the `/acs` route function to [upsert](https://www.mongodb.com/docs/drivers/node/current/fundamentals/crud/write-operations/upsert/) a native user record containing a user's UID and last login time to the MongoDB database, updating unit tests as needed.

Testing this requires assigning this deployment's domain to `staging.irvinehacks.com`, logging in as a native user, and then checking the MongoDB staging database to see if a record has been properly upserted. I can take care of assigning the domain and viewing the staging database because only directors have access to these, so reviewers should simply access staging, log in via UCI SSO, and let me know.

### DIFF
--- a/apps/api/src/routers/saml.py
+++ b/apps/api/src/routers/saml.py
@@ -10,7 +10,8 @@ from fastapi.responses import RedirectResponse, Response
 from onelogin.saml2.auth import OneLogin_Saml2_Auth
 from onelogin.saml2.settings import OneLogin_Saml2_Settings
 
-from auth import user_identity
+from auth.user_identity import NativeUser, utc_now, issue_user_identity
+from services.mongodb_handler import Collection, update_one
 
 log = getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


apps/api/src/routers/saml.py
```diff
@@ -81,6 +83,13 @@ async def _get_saml_auth(req: Request) -> OneLogin_Saml2_Auth:
     return OneLogin_Saml2_Auth(request_data, old_settings=settings)
 
 
+async def _insert_native_record(user: NativeUser) -> None:
```
taesungh, line 86: Nitpick: "insert" isn't always accurate here since when the record already exists, we're just updating it. Perhaps consider `_update_last_login` or `_save_login_time` or alike.

apps/api/src/routers/saml.py
```diff
@@ -81,6 +83,13 @@ async def _get_saml_auth(req: Request) -> OneLogin_Saml2_Auth:
...
+    now = utc_now()
+    await mongodb_handler.update_one(
+        Collection.USERS, {"_id": user.uid}, {"last_login": now}, upsert=True
+    )
```
taesungh, line 90: Thought [non-blocking]: I just discovered MongoDB supports [`$currentDate`](https://www.mongodb.com/docs/manual/reference/operator/update/currentDate/) as an update operator. I think for consistency with `guest_auth._remove_guest_key`, we can keep this for now.

taesungh (approved):
Thanks for adding. We'll still need to scrape the logs for previous incomplete applications.

PR #207, opened 2024-01-12
fix: add return after setting href to /portal

### LINKED ISSUE #206: [Application Form] Unexpected duplicate submission attempts
According to the production logs, every now and then a user encounters status 400 on `/api/user/apply` with the "user has already applied" message.

This might be because the client-side portal redirection does not happen immediately, and the `setSubmitting(false)` is still executed reenabling the submit button for a short moment. We could `return` immediately after the redirect to potentially resolve this.

samderanova:
Resolves #206 by adding a `return` after changing `window.location.href` to `/portal`.

### DIFF
--- a/apps/site/src/app/(main)/apply/sections/Form/Form.tsx
+++ b/apps/site/src/app/(main)/apply/sections/Form/Form.tsx
@@ -51,6 +51,7 @@ export default function Form() {
 				// to force reload the page to allow user identity to
 				// update with the new status
 				window.location.href = "/portal";
+				return;
 			}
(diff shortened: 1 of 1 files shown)


taesungh (commented):
Do you think we could also include a spinner of some sort inside the button during the `submitting` state to indicate something is happening?

samderanova:
> Do you think we could also include a spinner of some sort inside the button during the `submitting` state to indicate something is happening?

Sure, I can look into that.

taesungh (approved):
Works for now.

Issue #210, opened 2024-01-13
[Home Page] Adjust sponsors section to differentiate tiers


samderanova:
Currently, there is no differentiation between the different sponsor tiers. We'll want to organize them into their respective tiers and render images in their corresponding sizes (platinum are the largest, gold is the next largest, and so on).

PR #214, opened 2024-01-14
update: sponsor zod validation schema

### LINKED ISSUE #213: [Home Page] Update Zod validation schema to include other tiers
Going off of #209, the Zod schema in `getSponsors.ts` should be updated to include the other tiers as valid values.

samderanova:
Resolves #213. Updating the sponsors to their proper tiers at this current moment will cause Zod to throw an error in production, making the site inaccessible. Thus, the Zod schema should be updated first before making any other changes in the frontend.

The main reason for extracting and creating a new `Sponsor` validation is for #210, in which I plan on separating the sponsors into a `Map` by tier in the `getSponsors` function and returning that map instead of just the list of sponsors.


taesungh (approved):
Looks reasonable

PR #223, opened 2024-01-14
Provide endpoint to initiate waiver signing

### LINKED ISSUE #217: [API] Provide endpoint to initiate waiver signing
In the portal, we plan on having the user click a button with a link to initiate filling out the DocuSign [PowerForm](https://www.docusign.com/products/electronic-signature/features/powerforms) waiver.

It is possible to pre-complete some of the fields on the PowerForm by providing query parameters for email and name. However, because we don't have the user's email or name (only UID) in the frontend, we can create an endpoint in our API to generate a link to the PowerForm with those fields prefilled.

samderanova:
Closes #217.

## Changes
- Add `/user/waiver` route to redirect to DocuSign waiver
- Create `require_accepted_application` dependency to allow only accepted applicants
- Add new `WAIVER_SIGNED` status
- Include VSCode PyTest configuration
- Unit tests for all the above

## Testing
1. Set the `DEPLOYMENT` environment variable to `LOCAL`.
2. Start the backend by running `pnpm dev` inside `apps/api`.
3. Create authentication cookie by accessing `http://localhost:8000/dev/impersonate/<ucinetid>`.
4. Complete an application.
5. Navigate to `http://localhost:8000/user/waiver` and you should be redirected to the DocuSign form. If you instead see a DocuSign prompt asking for your name and/or email, please indicate in review.

**Note:** We'll eventually need to update the PowerForm, Account ID, and other DocuSign related information, but we can do that in a separate issue.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,31 @@
+{
+	// Use IntelliSense to learn about possible attributes.
+	// Hover to view descriptions of existing attributes.
+	// For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+	"version": "0.2.0",
+	"configurations": [
+		{
+			"name": "Python: Current File",
+			"type": "python",
+			"request": "launch",
+			"program": "${file}",
+			"purpose": ["debug-in-terminal"],
+			"console": "integratedTerminal",
+			"cwd": "${workspaceFolder}/apps/api",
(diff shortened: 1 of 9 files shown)


taesungh (approved):
Epic

rileysw (approved):
ty taesung 🙏

PR #224, opened 2024-01-15
Application will not send if user is on application page and deadline passes

IanWearsHat:
- `handleSubmit` in `Form.tsx` will reload the page if the deadlines has passed and NOT submit the form
- "If you already applied" message in `ApplicationsClosed.tsx` will not show if user is logged in, but hasn't submitted an application

### DIFF
--- a/apps/site/src/app/(main)/apply/sections/ApplicationsClosed/ApplicationsClosed.tsx
+++ b/apps/site/src/app/(main)/apply/sections/ApplicationsClosed/ApplicationsClosed.tsx
@@ -1,19 +1,25 @@
 import Link from "next/link";
 
-function ApplicationsClosed() {
+import getUserIdentity from "@/lib/utils/getUserIdentity";
+
+async function ApplicationsClosed() {
+	const identity = await getUserIdentity();
+
 	return (
 		<div className="rounded-2xl p-10 text-display text-[#432810] bg-white text-center w-8/12 max-[800px]:w-9/12 max-[400px]:w-11/12">
 			<h1 className="text-4xl p-10 font-semibold max-[800px]:p-0">
 				Applications for IrvineHacks 2024 closed on January 14th.
 			</h1>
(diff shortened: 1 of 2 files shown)


IanWearsHat:
Note: because there is no way to access the `/apply` route from the ui, the Applications Closed message really can't be viewed unless the user manually adds `/apply` to the url.

apps/site/src/app/(main)/apply/sections/ApplicationsClosed/ApplicationsClosed.tsx
```diff
@@ -1,19 +1,25 @@
...
-			<hr className="mt-5 w-full h-0.5 bg-[#432810]" />
-			<p className="pt-10 text-lg">
-				If you already applied, please{" "}
-				<Link href="/login" className="text-blue-600 underline">
-					log in
-				</Link>{" "}
-				to visit your applicant portal.
-			</p>
+			<hr className="my-10 w-full h-0.5 bg-[#432810]" />
+			{(identity.uid === null || identity.status !== null) && (
```
taesungh, line 14: If a user has already applied, they would be redirected to Portal anyway, so I don't think the status check is necessary.

apps/site/src/app/(main)/apply/sections/Form/Form.tsx
```diff
@@ -25,6 +26,12 @@ export default function Form() {
 	): Promise<void> => {
 		// Disable native post submission
 		event.preventDefault();
+
+		if (hasDeadlinePassed()) {
+			window.location.reload();
```
taesungh, line 31: Chore: Could you include a comment to indicate this is to show the messaged that applications have closed.

taesungh (approved):
Thanks for adding, looks alright. Hopefully not too scary for people trying to submit last minute.

samderanova (approved):

PR #226, opened 2024-01-15
fix: extended deadline by 2 minutes

Bl20052005:


samderanova (commented):
The deadline is also in the frontend, so please update that as well.

IanWearsHat (approved):
That is definitely an increase of 2 minutes

apps/site/src/app/(main)/apply/sections/ApplyConfirmation/ConfirmationDetails.tsx
```diff
@@ -24,7 +24,7 @@ export default async function ConfirmationDetails({
 				and will be over the age of 18 by January 26th, 2024.
 			</p>
 			<strong className="text-lg text-[#FF2222]">
-				Applications are due on January 14th, 2024 at 11:59PM PST.
+				Applications are due on January 15th, 2024 at 12:01PM PST.
```
taesungh: Chore [repeating]: Sorry the confusion, but we shouldn't publicly announce this change and leave it as 11:59.

Bl20052005: alright then

Bl20052005: is there anything else on the frontend that relies on dates that we should change, or is this purely a backend fix?

samderanova: Sorry, I should have been clearer in my previous comment. We perform a datetime check in the frontend as well in the hasDeadlinePassed function, so please update the date object there too.

taesungh (approved):
Thanks for updating.

samderanova (approved):
Looks good, thanks for the quick fix! Please *squash* merge.

PR #239, opened 2024-01-16
fix: fetch ResourceReference from sanity ref

njhuey:
Use GROQ join to fetch ResourceReference data from sanity.

Before this pr, only a reference to ResourceReference was being fetched from sanity when fetching Resources. This pr uses a GROQ join to fetch the ResourceReference data. Additionally, appropriate zod types have been created.


samderanova (approved):
Thanks for taking care of this.

Issue #221, opened 2024-01-14
[Portal] Implement next steps for accepted users


samderanova:
Accepted applicants should be prompted to sign the waiver and RSVP. Note that this is blocked by #219.

Relevant portions we'll need:
- [x] #236
- [x] #237
- [x] #238
- [x] #243
- [x] #253

samderanova:
Duplicate of #219

taesungh:
Closing as all tasks are completed (besides DocuSign handler).

PR #240, opened 2024-01-16
Add `waived` status to `PortalStatus`

taesungh:
Following from #223 and as part of #221.
- Include new enum value: `PortalStatus.waived = "WAIVER_SIGNED"`
- Add empty Portal Message for waived status


waalbert (approved):

samderanova (approved):
I'm assuming we'll be filling this with

PR #247, opened 2024-01-16
Add SendGrid templates to send applicant decision emails

### LINKED ISSUE #241: [API] Include new SendGrid templates IDs for applicant decision release
As part of decision release, we'll be using personalized emails to announced the applicant's decision
- Include new template IDs for accepted, waitlisted, and rejected applicants
- Provide a `send_decision_email` function in `email_handler` to pick the appropriate template based on a user's status and pass in the personalization data
  - We might need to make this function handle multiple user's at once to avoid being rate-limited by SendGrid

samderanova:
Resolves #241. Wrote the SendGrid `send_decision_email` function along with unit tests.

### DIFF
--- a/apps/api/src/services/sendgrid_handler.py
+++ b/apps/api/src/services/sendgrid_handler.py
@@ -7,7 +7,10 @@ from typing import Iterable, Literal, Tuple, TypedDict, Union, overload
 
 import aiosendgrid
 from httpx import HTTPStatusError
+from pydantic import EmailStr
 from sendgrid.helpers.mail import Email, Mail, Personalization
+
+from models.ApplicationData import Decision
 
 log = getLogger(__name__)
 
@@ -17,9 +20,9 @@ SENDGRID_API_KEY = os.getenv("SENDGRID_API_KEY")
 class Template(str, Enum):
     CONFIRMATION_EMAIL = "d-83d42cc17b54456183eeb946ba58861a"
     GUEST_TOKEN = "d-1998e588ddf74c6d9ede36b778730176"
-    ACCEPTED_EMAIL = "d-9178c043de134a71a4fdbe513d35f89f"
+    ACCEPTED_EMAIL = "d-062e7106a0d64d49ad9f03325bbc7286"
     WAITLISTED_EMAIL = "d-9178c043de134a71a4fdbe513d35f89f"
-    REJECTED_EMAIL = "d-9178c043de134a71a4fdbe513d35f89f"
+    REJECTED_EMAIL = "d-71ef30ac91a941e0893b7680928d80b7"
 
 
 class PersonalizationData(TypedDict):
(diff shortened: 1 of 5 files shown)


apps/api/src/services/sendgrid_handler.py
```diff
@@ -114,3 +124,31 @@ async def send_email(
...
         log.exception("During SendGrid processing: %s", e)
         raise RuntimeError("Could not send email with SendGrid")
+
+
+async def send_decision_email(
+    sender_email: Tuple[str, str], applicant_batch: dict[tuple[str, EmailStr], Decision]
+):
+    accepted_personalization_data = []
+    rejected_personalization_data = []
+    waitlisted_personalization_data = []
```
taesungh, line 145: Suggestion: given how we're using three if/elifs below, instead of separate lists, could we create a dictionary with each `Decision` value as the key?

apps/api/src/services/sendgrid_handler.py
```diff
@@ -114,3 +124,31 @@ async def send_email(
     except HTTPStatusError as e:
         log.exception("During SendGrid processing: %s", e)
         raise RuntimeError("Could not send email with SendGrid")
+
+
+async def send_decision_email(
+    sender_email: Tuple[str, str], applicant_batch: dict[tuple[str, EmailStr], Decision]
```
taesungh, line 141: Thought [non-blocking]: the format of the batch data here might be a little inconvenient to construct for the consumer. I'll think if there's a way to make it simpler, e.g. by using a `Protocol`.

taesungh (changes requested):
Request for reorganization and one issue with potentially empty data.

apps/api/src/services/sendgrid_handler.py
```diff
@@ -65,6 +75,17 @@ async def send_email(
     ...
 
 
+@overload
+async def send_email(
+    template_id: Literal[Template.CONFIRMATION_EMAIL],
+    sender_email: Tuple[str, str],
+    receiver_data: Iterable[ApplicationUpdatePersonalization],
```
taesungh, line 82: Issue: template doesn't match receiver data, might need to specify a union of the three templates.

taesungh (dismissed):
Thanks for adding, assuming email content is being reviewed by logistics.

apps/api/tests/test_user_apply.py
```diff
@@ -289,3 +291,32 @@ def test_past_deadline_causes_403() -> None:
     assert res.status_code == 403
 
     user.DEADLINE = TEST_DEADLINE
+
+
+@patch("services.sendgrid_handler.send_email")
+async def test_send_decision_email(mock_sendgrid_handler_send_email: AsyncMock) -> None:
```
taesungh, line 297: Chore: could you move this test to a new `test_email_handler.py`

waalbert (dismissed):

taesungh (approved):

PR #248, opened 2024-01-16
Add endpoint for participants to RSVP

### LINKED ISSUE #243: [API] Import user RSVP endpoint
Together with #238, we'll need to import last year's [`/user/rsvp` endpoint](https://github.com/HackAtUCI/HackUCI-Site/blob/0b181fb750094373e2ab3f689d2691312bb88ee7/src/api/routers/user.py#L136).

samderanova:
Resolves #243.

### DIFF
--- a/apps/api/src/routers/user.py
+++ b/apps/api/src/routers/user.py
@@ -173,7 +173,9 @@ async def request_waiver(
 
 
 @router.post("/rsvp")
-async def rsvp(user: User = Depends(require_user_identity)) -> RedirectResponse:
+async def rsvp(
+    user: Annotated[User, Depends(require_user_identity)]
+) -> RedirectResponse:
     """Change user status for RSVP"""
     user_record = await mongodb_handler.retrieve_one(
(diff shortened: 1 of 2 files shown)


apps/api/src/routers/user.py
```diff
@@ -170,3 +170,31 @@ async def request_waiver(
 
     form_url = docusign_handler.waiver_form_url(user_data.email, user_name)
     return RedirectResponse(form_url, status.HTTP_303_SEE_OTHER)
+
+
+@router.post("/rsvp")
+async def rsvp(user: User = Depends(require_user_identity)) -> RedirectResponse:
```
taesungh, line 176: Chore: can we update to the `Annotated` syntax

taesungh (commented):
Would appreciate a unit test but understandable if short on time.

taesungh (approved):
Thanks for adding unit tests and separating the raw import.

waalbert (approved):

PR #254, opened 2024-01-17
Fix: prevent sign waiver from showing when user is confirmed and properly update vertical timeline

### LINKED ISSUE #253: [Portal] Sign Waiver shown when confirmed and Vertical Timeline not updated
When user's status is `CONFIRMED`, the Vertical Timeline still shows an empty circle and says "Confirm attendance", and the `SignWaiver` component is shown when it shouldn't be.

The conditional status was not implemented properly in #244.

waalbert:
Resolves #253

### DIFF
--- a/apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
+++ b/apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
@@ -28,14 +28,14 @@ async function Portal() {
 	const submittedWaiver =
 		status === PortalStatus.waived || status === PortalStatus.confirmed;
 
-	const isAccepted = status === PortalStatus.accepted || submittedWaiver;
+	const isAccepted = status === PortalStatus.accepted;
 
 	return (
 		<div className="bg-white text-black max-w-4xl rounded-2xl p-6 flex flex-col mb-24 w-full">
 			<h2 className="text-4xl font-semibold">Status</h2>
 			<VerticalTimeline status={status as PortalStatus} />
 			<Message status={status as PortalStatus} />
-			{isAccepted && <SignWaiver status={status as PortalStatus} />}
+			{isAccepted && !submittedWaiver && <SignWaiver />}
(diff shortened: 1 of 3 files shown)


apps/site/src/app/(main)/portal/@applicant/SignWaiver.tsx
```diff
@@ -7,7 +7,8 @@ interface SignWaiverProps {
 }
 
 function SignWaiver({ status }: SignWaiverProps) {
-	return status === PortalStatus.waived ? null : (
+	return status === PortalStatus.waived ||
+		status === PortalStatus.confirmed ? null : (
```
taesungh, line 11: Chore: we already conditionally render this `SignWaiver` component from `ApplicantPortal`, so could we fix the condition there and remove the condition here?

apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
```diff
@@ -28,14 +28,14 @@ async function Portal() {
...
-	const isAccepted = status === PortalStatus.accepted || submittedWaiver;
+	const isAccepted = status === PortalStatus.accepted;
 
 	return (
 		<div className="bg-white text-black max-w-4xl rounded-2xl p-6 flex flex-col mb-24 w-full">
 			<h2 className="text-4xl font-semibold">Status</h2>
 			<VerticalTimeline status={status as PortalStatus} />
 			<Message status={status as PortalStatus} />
-			{isAccepted && <SignWaiver status={status as PortalStatus} />}
+			{isAccepted && !submittedWaiver && <SignWaiver />}
```
taesungh, line 38: Thought: all of the states are mutually exclusive, so we shouldn't need the second conditional. Perhaps consider renaming `isAccepted` to `needsToSignWaiver`?

samderanova (approved):
Thanks for simplifying the conditionals and fixing the issue.

taesungh (approved):
Thanks for fixing this.

PR #257, opened 2024-01-17
DocuSign Handler Updates

### LINKED ISSUE #246: [API] PowerForm ID for waiver needs to be updated
Following from #217 and #223 and testing in #245, we'll need to add the actual waiver document to the production instance of DocuSign and include the associated PowerForm ID in our API. The subdomain and query parameters of the PowerForm URL may also need to be updated.

samderanova:
Resolves #246.

## Changes
- Add production PowerForm and account IDs and allow usage of staging ones depending on STAGING_ENV
- Update unit tests to use production PowerForm ID for HMAC signature


taesungh (approved):
Thanks for updating.

PR #259, opened 2024-01-17
Create decision release endpoint for directors

### LINKED ISSUE #242: [API] Add endpoint to release application decisions and send emails
We'll need another admin endpoint to release application decisions.
- Add new route under Admin router to release decisions
  - Update user's `status` with final review value (I think `_include_review_decision` should be sufficient)
- Send decision email to applicants
  - Can use `send_decision_email` from #241
  - We may need to use batching with the SendGrid API to avoid being rate limited as we do have several hundred applicants

samderanova:
Closes #242.

Worked with @taesungh to create a decision release endpoint accessible only by directors at `/api/admin/release`. It updates the statuses of applicants with their most recent review and sends out the respective decision emails via SendGrid.

We tested this locally on Docker MongoDB instances and it works as expected. The only concern is whether SendGrid will rate limit us for sending too many emails at once, although we did implement batching algorithms to work around this.

### DIFF
--- a/apps/api/src/routers/admin.py
+++ b/apps/api/src/routers/admin.py
@@ -1,16 +1,19 @@
+import asyncio
 from datetime import datetime
 from logging import getLogger
 from typing import Any, Optional
 
 from fastapi import APIRouter, Body, Depends, HTTPException, status
-from pydantic import BaseModel, Field, TypeAdapter, ValidationError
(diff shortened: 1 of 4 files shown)


taesungh (approved):
Tested with Riley on staging domain and small sample. Hopefully resilient to several hundred decisions.

rileysw (approved):

PR #265, opened 2024-01-17
Redirect repeated guest login to verification form

IanWearsHat:


apps/api/src/auth/guest_auth.py
```diff
@@ -32,10 +32,6 @@ class GuestRecord(BaseRecord):
 async def initiate_guest_login(email: EmailStr) -> Optional[str]:
     """Generate a login passphrase to be emailed, save the authentication key,
     and a confirmation token to be saved as a cookie."""
-    if await _get_existing_key(email):
-        # To prevent spammed requests, user must wait until previous key expires
-        return None
-
```
taesungh, line 38: Issue: we don't want to regenerate a new key when the user has a recent unexpired key. Instead of changing the return value here, we can just redirect to the verification form from `guest.guest_login` if `confirmation` is `None`

apps/api/tests/test_guest.py
```diff
@@ -72,16 +72,17 @@ def test_guest_login_initiation(
...
-def test_requesting_login_when_previous_key_exists_causes_429(
-    mock_get_existing_key: AsyncMock,
-) -> None:
-    """Test that requesting to log in as guest when the user has an existing,
-    unexpired key causes status 429."""
-
-    mock_get_existing_key.return_value = "some-existing-key"
-    res = client.post("/login", data=SAMPLE_LOGIN_DATA)
-    assert res.status_code == 429
+# Commenting out this test to disable 429 from existing, unexpired key
```
taesungh, line 75: Could we update the test to reflect that in such a situation, the result should be being redirected to the verification and not having the `guest_confirmation` cookie modified?

taesungh (changes requested):

taesungh (approved):
Nice work figuring out how this system works.

samderanova (approved):
Thanks for fixing this issue! Please squash merge when ready.

PR #283, opened 2024-01-20
Warning for `ATTENDING` users who are about to un-rsvp

IanWearsHat:
- Created `ATTENDING` PortalStatus in `ApplicantPortal.tsx`
 - Updated VerticalTimeline logic to include new `ATTENDING` status
 - Moved rsvp form to its own client component
   - Form displays confirmation dialog if `ATTENDING` user tries to un-rsvp
 - Added warning message above un-rsvp button shown only to `ATTENDING` users

### DIFF
--- a/apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
+++ b/apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
@@ -27,7 +27,9 @@ async function Portal() {
 	}
 
 	const submittedWaiver =
-		status === PortalStatus.waived || status === PortalStatus.confirmed;
+		status === PortalStatus.waived ||
+		status === PortalStatus.confirmed ||
+		status === PortalStatus.attending;
 
 	const needsToSignWaiver = status === PortalStatus.accepted;
 
--- a/apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
+++ b/apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
@@ -15,6 +15,7 @@ export const enum PortalStatus {
 	waitlisted = "WAITLISTED",
 	waived = "WAIVER_SIGNED",
 	confirmed = "CONFIRMED",
+	attending = "ATTENDING",
 }
 
 async function Portal() {
--- a/apps/site/src/app/(main)/portal/@applicant/RsvpForm.tsx
+++ b/apps/site/src/app/(main)/portal/@applicant/RsvpForm.tsx
@@ -4,10 +4,10 @@ import Button from "@/lib/components/Button/Button";
 
 interface RsvpFormProps {
 	buttonText: string;
-	showMessage: boolean;
+	showWarning: boolean;
 }
 
-export default function RsvpForm({ buttonText, showMessage }: RsvpFormProps) {
+export default function RsvpForm({ buttonText, showWarning }: RsvpFormProps) {
 	const confirmationMessage =
 		"WARNING: You will not be able to RSVP again. Are you sure you want to continue?";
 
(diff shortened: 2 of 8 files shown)


apps/site/src/app/(main)/portal/@applicant/VerticalTimeline.tsx
```diff
@@ -75,7 +76,9 @@ function VerticalTimeline({ status }: VerticalTimelineProps) {
 				/>
 				Sign waiver
 			</li>
-		) : status === PortalStatus.waived || status === PortalStatus.confirmed ? (
+		) : status === PortalStatus.waived ||
+		  status === PortalStatus.confirmed ||
+		  status === PortalStatus.attending ? (
```
taesungh, line 81: Rather than fighting the linters, could we extract some of these conditions to named boolean values, e.g. `attendanceConfirmed` and `signedWaiver`

apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
```diff
@@ -15,6 +15,7 @@ export const enum PortalStatus {
 	waitlisted = "WAITLISTED",
 	waived = "WAIVER_SIGNED",
 	confirmed = "CONFIRMED",
+	attending = "ATTENDING",
 }
```
samderanova, line 20: Issue: If someone has a status of `ATTENDING`, that means they submitted the waiver. Could you update the `submittedWaiver` variable later in this file to include `PortalStatus.attending`? Otherwise, the `ConfirmAttendance` component will not be shown.

apps/site/src/app/(main)/portal/@applicant/RsvpForm.tsx
```diff
@@ -0,0 +1,27 @@
...
+		<form
+			method="post"
+			action="/api/user/rsvp"
+			onSubmit={(event) => {
+				if (showMessage && !confirm(confirmationMessage)) {
+					event.preventDefault();
+				}
+			}}
+		>
+			<Button className="text-color-red" text={buttonText} />
```
samderanova, line 24: Issue: On mobile views, the button extends past the width of the screen. Could we add a `style` prop to `Button` with a default value of `width: max-content` (so we don't affect other `Button`s) and then for this instance, pass in `width: 100%`?

samderanova (changes requested):

apps/site/src/app/(main)/portal/@applicant/VerticalTimeline.tsx
```diff
@@ -12,6 +12,10 @@ interface VerticalTimelineProps {
 }
 
 function VerticalTimeline({ status }: VerticalTimelineProps) {
+	const signedWaiver = status === PortalStatus.waived;
+	const attendanceConfirmed = status === PortalStatus.confirmed;
+	const sentRSVP = status === PortalStatus.attending;
```
taesungh, line 17: Suggestion: we could make the three conditions cascading so we don't have to write the triple disjunction `signedWaiver || attendanceConfirmed || sentRSVP` in two different places below.
```suggestion
	const confirmedAttendance = status === PortalStatus.confirmed || status === PortalStatus.attending;
	const signedWaiver = status === PortalStatus.waived || confirmedAttendance;
```
We would then need to revert the change to the first condition for `rsvp_component`.

taesungh (commented):
Some additional suggestions besides the issues Sam mentioned.

samderanova (approved):
Thanks for implementing this along with the `confirm` message. We'll be running into merge conflicts with #289, so let's merge this first since it has less files before merging this one.

taesungh (approved):
Nice work, let's merge this before #289.

PR #286, opened 2024-01-20
Specify file inputs to Turborepo workspace of site

### LINKED ISSUE #285: Turborepo configuration needs explicit file inputs
As described in #271, files from `.next/cache/` and `apps/site/.next/static/` are being considered [file inputs](https://turbo.build/repo/docs/core-concepts/caching/file-inputs) because Vercel does not receive the `.gitignore` file when deploying through our workflow which is based on the Vercel CLI.

As such, the build cache for the site was sometimes missed because Turborepo would spuriously consider changes in the cached `.next` files as inputs
- Specify relevant source files as `pipeline.build.inputs` in `apps/site/turbo.json`

taesungh:
Fixes #285. In the build logs for the preview deployment, you should see "FULL TURBO".

## Changes
- To optimize build cache and enable Full Turbo, add explicit paths to source files as inputs for build stage

## Impact
This will ensure cache can be used to reduce build times by around 40 seconds during redeployments, deployments of pull requests with no diverging tree, or deployments with changes unrelated to Site source files (e.g. changes only to API).

### DIFF
--- a/apps/site/turbo.json
+++ b/apps/site/turbo.json
@@ -1,14 +1,18 @@
 {
-	"extends": [
-		"//"
-	],
+	"extends": ["//"],
 	"pipeline": {
 		"build": {
-			"outputs": [
-				".vercel/**",
-				".next/**",
-				"!.next/cache/**"
-			]
+			"inputs": [
+				"src/**",
+				"public/**",
+				".eslintrc.json",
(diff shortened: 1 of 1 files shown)


samderanova (approved):
Thanks for working on this. It'll definitely speed up build times.

PR #288, opened 2024-01-20
feat: new API route to apply status changes after RSVP deadline

Bl20052005:


samderanova (changes requested):

apps/api/src/routers/admin.py
```diff
@@ -165,3 +194,15 @@ def _include_review_decision(applicant_record: dict[str, Any]) -> None:
...
+def _set_confirmations(applicant_record: dict[str, Any]) -> None:
+    """Sets the applicant's status based on their RSVP status"""
+
+    status = applicant_record["status"]
+    if status == Status.CONFIRMED:
+        applicant_record["status"] = Status.ATTENDING
+    elif status == Status.WAIVER_SIGNED:
+        applicant_record["status"] = Status.VOID
+    elif status == Decision.ACCEPTED:
+        applicant_record["status"] = Status.VOID
```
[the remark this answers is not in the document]
taesungh: Or alternatively write `status in (...)`

apps/api/src/routers/admin.py
```diff
@@ -125,12 +125,41 @@ async def release_decisions() -> None:
...
+    )
+
+    for record in records:
+        _set_confirmations(record)
+
+    for cur_status in (Status.ATTENDING, Status.VOID):
+        group = [record for record in records if record["status"] == cur_status]
+        await asyncio.gather(
+            *(_process_status(batch, cur_status) for batch in batched(group, 100))
+        )
```
samderanova, line 151: Issue: While this code properly changes the status of every applicant that has `CONFIRMED`, `WAIVER_SIGNED`, or `ACCEPTED`, it runs update queries for applicants that already have a status of `ATTENDING` or `VOID`, which leads to unnecessary queries sent to the database. To cut down on the number of queries, could you iterate only over the statuses that need to be changed and perform updates on those?

samderanova (changes requested):

apps/api/src/routers/admin.py
Bl20052005: I have updated the logic, however, now it becomes pretty much untestable as it's now all inside the router function, but I'm not sure if we truly require it to be tested

I have commented the previous tests and helper functions out for now

apps/api/src/routers/admin.py
```diff
@@ -125,12 +125,68 @@ async def release_decisions() -> None:
...
+        ]
+
+        for record in void_record:
+            record["status"] = Status.VOID
+        await asyncio.gather(
+            *(
+                _process_status(batch, Status.VOID)
+                for batch in batched(void_record, 100)
+            )
+        )
```
taesungh, line 169: Suggestion: instead of repeating this logic, we could use a mapping of `status` to `new_status` to describe the enforcement process.
```python
{
    Status.CONFIRMED: Status.ATTENDING,
    Status.ACCEPTED: Status.VOID,
    Status.WAIVER_SIGNED: Status.VOID,
}

File: apps/api/src/routers/admin.py [hunk 4]
```diff
@@ -125,12 +125,68 @@ async def release_decisions() -> None:
...
+
+    # for cur_status in (Status.ATTENDING, Status.VOID):
+    #     group = [record for record in records if record["status"] == cur_status]
+    #     await asyncio.gather(
+    #         *(_process_status(batch, cur_status) for batch in batched(group, 100))
+    #     )
+
+
+async def _process_status(batch: tuple[dict[str, Any], ...], status: Status) -> None:
+    uids: list[str] = [record["_id"] for record in batch]
```
taesungh, line 182: Suggestion: instead of having this function receive all of the user data, we could provide just the UIDs from the consumer.
```suggestion
async def _process_status(batch: Sequence[str], status: Status) -> None:
```

apps/api/tests/test_admin.py
```diff
@@ -128,6 +140,37 @@ def test_no_decision_from_no_reviews() -> None:
...
+    """Test that confirmed status changes to attending."""
+
+    mock_mongodb_handler_retrieve_one.return_value = DIRECTOR_IDENTITY
+    mock_mongodb_handler_retrieve.return_value = [
+        {
+            "_id": "edu.uc.tester",
+            "role": "applicant",
+            "status": Status.CONFIRMED,
+        }
+    ]
```
taesungh, line 160: Chore: Could we also include the two other conversion situations in the test data?

apps/api/tests/test_admin.py
```diff
@@ -164,13 +164,75 @@ def test_confirm_attendance_when_status_confirmed(
...
+
+@patch("services.mongodb_handler.update", autospec=True)
+@patch("services.mongodb_handler.retrieve_one", autospec=True)
+@patch("services.mongodb_handler.retrieve", autospec=True)
+def test_confirm_attendance_when_status_accepted(
+    mock_mongodb_handler_retrieve: AsyncMock,
+    mock_mongodb_handler_retrieve_one: AsyncMock,
+    mock_mognodb_handler_update: AsyncMock,
+) -> None:
+    """Test that confirmed status changes to attending."""
```
samderanova: Issue: Could you update the docstrings to reflect the new tests?

samderanova (changes requested):

apps/api/tests/test_admin.py
taesungh: Instead of repeating the same test three times, we could include three applicants of different statuses in the mocked return value, also to simulate a more realistic scenario.

samderanova (approved):

taesungh (approved):
Can improve tests a bit more after we enforce.

PR #291, opened 2024-01-20
Fix `NEXT_PUBLIC_` environment variables in cache

### LINKED ISSUE #290: `NEXT_PUBLIC_VERCEL_URL` incorrectly reused from build cache for site
In some recent deployments, the server-side API calls were going to a different deployment because the `NEXT_PUBLIC_VERCEL_URL` was being baked into the build, and the Turborepo cache doesn't consider this as an input.

> All `NEXT_PUBLIC_` variables will be frozen with the value evaluated at build time, so these values need to be set appropriately when the project is built ([Configuring: Environment Variables, Next.js](https://nextjs.org/docs/app/building-your-application/configuring/environment-variables))

If we do specify `NEXT_PUBLIC_VERCEL_URL` as a cache input key, this means we will never be able to utilize the cache since this environment variable changes for each deployment. To avoid this, we might be able to use a dynamic lookup to prevent inlining as shown in the documentation example.
```js
// This will NOT be inlined, because it uses a variable
const varName = 'NEXT_PUBLIC_ANALYTICS_ID'
setupAnalyticsService(process.env[varName])
 
// This will NOT be inlined, because it uses a variable
const env = process.env
setupAnalyticsService(env.NEXT_PUBLIC_ANALYTICS_ID)
```

taesungh:
Fixes #290.

## Changes
- Replace `NEXT_PUBLIC_VERCEL_URL` with `VERCEL_URL` to avoid inlining
- Add `NEXT_PUBLIC_SANITY_*` to Turborepo `pipeline.build.env` cache key

## Testing
Confirm deployment logs show that server-side API calls are going to the current deployment, even when the site build was from Full Turbo cache.

### DIFF
--- a/apps/site/src/lib/utils/api.ts
+++ b/apps/site/src/lib/utils/api.ts
@@ -2,7 +2,7 @@ import axios from "axios";
 import { cookies } from "next/headers";
 
 const LOCAL_API_URL = "http://localhost:8000";
-const SERVER_HOST = process.env.NEXT_PUBLIC_VERCEL_URL;
+const SERVER_HOST = process.env.VERCEL_URL;
 
 // The Vercel Serverless Function for the API lives outside the scope of Next.js
 // so the publicly deployed URL must be used instead of a rewrite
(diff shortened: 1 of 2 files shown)


samderanova (approved):
Epic

PR #293, opened 2024-01-20
feat: add waitlist release SendGrid email template

### LINKED ISSUE #292: [API] Add waitlist email SendGrid template to list of templates
Since we are releasing applicants off the waitlist soon, we'll want to be able to send them emails about their release via SendGrid, which means we need to add the email template for it.

samderanova:
Resolves #292.


taesungh (approved):
Trusting the templated ID is correct.

Issue #269, opened 2024-01-17
[Admin Site] Create separate page to facilitate participant check-in process


samderanova:
The third step of the check-in construction is to create a new page on the admin site to provide check-in members with an easy way to check people in. It should contain a table view for participants containing the name, UID, status, and role obtained from #268, along with an actions menu where one can choose whether to promote the current user off the waitlist or check them in. Since we use Cloudscape for the admin portal at the moment, see the [Cloudscape demo](https://cloudscape.design/examples/react/table.html) for an example on how to implement this. No need to provide a column for checkboxes because we'll be checking in one person at a time instead of checking multiple people in at once.

- [x] #334
- [x] #339
- [x] #340
- [x] #343
- [x] #344
- [x] If time permits: collection preferences

taesungh:
As noted in #315, only certain check-in leads will be allowed to promote users from the waitlist through the endpoint provided by #307/#316. We'll also want a feature to confirm attendance for non-hackers as noted in #318.

taesungh:
Considering complete as core functionality was added and used during the event. Continuation work for improvements described in #407 and #408.

PR #296, opened 2024-01-20
Create new volunteer and organizer roles limited to check-ins only

### LINKED ISSUE #267: Create new volunteer and organizer roles limited to check-ins only
To start off the construction of our check-in process, we'll have to create new `volunteer` and `organizer` roles to grant them access to only the check-in page on the admin site.

While there is a `tech_organizer` role currently in use, it is not being used, so this can simply be replaced by the `organizer` role. Additionally, the `volunteer` role already exists, so the only role that really needs to be created is the `organizer` one.

waalbert:
- `tech_organizer` role replaced with `organizer`
- Update `isAdminRole` to allow organizer and volunteer
- Use separate `isApplicationManager` function for the previous admin roles (director and reviewer).
- Create an empty page called Participants for https://github.com/HackAtUCI/irvinehacks-site-2024/issues/269 and show this in the AdminSidebar
- Hide the Applicants path and restrict the related pages to roles which satisfy `isApplicationManager`
- Resolves #267

### DIFF
--- a/apps/api/src/utils/user_record.py
+++ b/apps/api/src/utils/user_record.py
@@ -13,7 +13,7 @@ class Role(str, Enum):
     HACKER = "hacker"
     MENTOR = "mentor"
     REVIEWER = "reviewer"
-    TECH_ORGANIZER = "tech_organizer"
+    ORGANIZER = "organizer"
     VOLUNTEER = "volunteer"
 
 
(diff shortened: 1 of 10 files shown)


samderanova (approved):
Looking good, thanks for working on this.

taesungh (approved):
Nice work, thanks for adding.

PR #308, opened 2024-01-21
feat: include log messages for rsvp changes

IanWearsHat:
Also added unit test for exception case in `/rsvp`

### DIFF
--- a/apps/api/tests/test_user.py
+++ b/apps/api/tests/test_user.py
@@ -7,6 +7,7 @@ from auth.user_identity import GuestUser, UserTestClient
 from routers import user
 from services.mongodb_handler import Collection
 from utils.user_record import Status
+from models.ApplicationData import Decision
 
 app = FastAPI()
 app.include_router(user.router)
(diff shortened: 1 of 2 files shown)


apps/api/tests/test_user.py
```diff
@@ -116,3 +116,19 @@ def test_user_with_status_attending_un_rsvp_changes_status_to_void(
...
     assert res.status_code == 303
+
+@patch("services.mongodb_handler.update_one", autospec=True)
+@patch("services.mongodb_handler.retrieve_one", autospec=True)
+def test_user_with_status_reviewed_un_rsvp_returns_403(
+    mock_mongodb_handler_retrieve_one: AsyncMock,
+    mock_mongodb_handler_update_one: AsyncMock,
+) -> None:
+    """Test user with REVIEWED status has no status change after RSVP."""
+    mock_mongodb_handler_retrieve_one.return_value = {"status": Status.REVIEWED}
```
taesungh, line 128: Suggestion: a more appropriate test would be with `Status.ACCEPTED` to emphasize that users cannot RSVP without submitting the waiver.

IanWearsHat: `ACCEPTED` is currently not a status in `Status` for `user_record.py`. Should I add it?

taesungh: Sorry I meant `Decision.ACCEPTED`

taesungh (approved):
Thanks for adding this is as well as the unit test.

samderanova (approved):
Thanks for working on this.

PR #320, opened 2024-01-21
fix: override model_dump instead of dict

### LINKED ISSUE #319: UID being inserted twice into database
This isn't a major issue, but we've found that many of our records in the database contain a user's uid twice: once as the ID of the document and another as the value of the field `uid`. We'll need to override the `model_dump` method in `BaseRecord` so that the `uid` field gets appropriately aliased by `_id`.

samderanova:
Resolves #319.

### DIFF
--- a/apps/api/src/services/mongodb_handler.py
+++ b/apps/api/src/services/mongodb_handler.py
@@ -33,7 +33,7 @@ class BaseRecord(BaseModel):
 
     uid: str = Field(alias="_id")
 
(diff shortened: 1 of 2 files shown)


taesungh (commented):
Could we get a unit test to confirm this works: e.g. the key `"uid"` should not be in the result of using `model_dump` on an Applicant record.

waalbert (approved):

Issue #268, opened 2024-01-17
[API] Create endpoints for participant check-in system


samderanova:
Two new endpoints need to be created in the second part of the check-in construction:

1. One endpoint should fetch the name, UID, status, and role from the `ATTENDING` applicants in the database. We'll want to verify that the person accessing this route has a role of `organizer`, `volunteer`, or `director` (see #267) first via `Depends`. 
2. The second endpoint should allow check-ins to be recorded in our database. A new key called "checkins" should be created whose value is an array. We'll want to store the UID of the associate that checked the participant in along with the time as a tuple and then [push](https://www.mongodb.com/docs/manual/reference/operator/update/push/) it to the array. See `admin.submit_review` for a similar example.

Issue #266, opened 2024-01-17
[Admin] Implement participant check-in system


samderanova:
Organizers on check in will need an easy interface to check people in to our event. The main tasks for this are as follows:

- [x] #267
- [x] #268
- [x] #269
- [x] #279
- [x] #318
- [x] #355
- [x] #364

and if time permits:
- Include [Onboarding](https://cloudscape.design/patterns/general/onboarding/) process with [Help system](https://cloudscape.design/patterns/general/help-system/) and [Hands-on tutorials](https://cloudscape.design/patterns/general/onboarding/hands-on-tutorials/)

taesungh:
Closing since core tasks were completed and successfully used during the event. Will create follow-up issue about onboarding/help system.

PR #321, opened 2024-01-21
Feature/api particpant check in system

njhuey:
Resolves https://github.com/HackAtUCI/irvinehacks-site-2024/issues/268.

Add endpoint `/admin/attending` and `/admin/checkin to facilitate check-in procedure.

### DIFF
--- a/apps/api/src/routers/admin.py
+++ b/apps/api/src/routers/admin.py
@@ -217,6 +217,7 @@ async def confirm_attendance() -> None:
     ],
 )
 async def attending() -> list[dict[str, object]]:
+    """Fetch all applicants who have a status of ATTENDING"""
     records = await mongodb_handler.retrieve(
         Collection.USERS,
         {"role": Role.APPLICANT, "status": Status.ATTENDING},
@@ -239,7 +240,16 @@ async def checkin(
         require_role([Role.DIRECTOR, Role.ORGANIZER, Role.VOLUNTEER])
     ),
 ) -> None:
-    new_checkin_entry = {"uid": associate.uid, "time": datetime.now()}
+    """Check-in applicant at IrvineHacks"""
+    record: Optional[dict[str, object]] = await mongodb_handler.retrieve_one(
+        Collection.USERS, {"_id": uid, "role": Role.APPLICANT}
+    )
+    if not record:
+        raise HTTPException(status.HTTP_404_NOT_FOUND)
+    elif record["status"] != Status.ATTENDING:
+        raise RuntimeError("Applicant status must be ATTENDING")
+
+    new_checkin_entry = {"uid": associate.uid, "time": utc_now()}
 
     update_status = await mongodb_handler.raw_update_one(
         Collection.USERS,
@@ -251,6 +261,8 @@ async def checkin(
 
     if not update_status:
         raise HTTPException(status.HTTP_404_NOT_FOUND)
+    else:
+        log.info(f"Applicant {uid} checked in by {associate.uid}")
 
 
 async def _process_status(uids: Sequence[str], status: Status) -> None:
--- a/apps/api/src/routers/admin.py
+++ b/apps/api/src/routers/admin.py
@@ -6,7 +6,7 @@ from typing import Any, Optional, Sequence
 from fastapi import APIRouter, Body, Depends, HTTPException, status
 from pydantic import BaseModel, EmailStr, Field, TypeAdapter, ValidationError
 
-from admin import summary_handler
+from admin import participant_manager, summary_handler
 from auth.authorization import require_role
 from auth.user_identity import User, utc_now
 from models.ApplicationData import Decision, Review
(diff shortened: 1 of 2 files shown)


samderanova (approved):
Worked with @njhuey on local to test both endpoints. Since this is blocking @taesungh's work on the admin page, we can merge this first and then create a separate issue to write unit tests for these.

apps/api/src/routers/admin.py
```diff
@@ -210,6 +210,49 @@ async def confirm_attendance() -> None:
...
 
 
+@router.get(
+    "/attending",
+    dependencies=[
+        Depends(require_role([Role.DIRECTOR, Role.ORGANIZER, Role.VOLUNTEER]))
+    ],
+)
+async def attending() -> list[dict[str, object]]:
+    records = await mongodb_handler.retrieve(
```
taesungh, line 220: Nitpick [non-blocking, repeating]: missing docstring.

apps/api/src/routers/admin.py
```diff
@@ -210,6 +210,49 @@ async def confirm_attendance() -> None:
...
+
+
+@router.post("/checkin/{uid}")
+async def checkin(
+    uid: str,
+    associate: User = Depends(
+        require_role([Role.DIRECTOR, Role.ORGANIZER, Role.VOLUNTEER])
+    ),
+) -> None:
+    new_checkin_entry = {"uid": associate.uid, "time": datetime.now()}
```
taesungh, line 242: Issue: I think MongoDB will require a zoned time, we can use `utc_now` similar to as in `submit_review`.
Suggestion: given such a basic structure, could we just use a tuple rather than a dict?

apps/api/src/routers/admin.py
```diff
@@ -210,6 +210,49 @@ async def confirm_attendance() -> None:
...
+    update_status = await mongodb_handler.raw_update_one(
+        Collection.USERS,
+        {"_id": uid},
+        {
+            "$push": {"checkins": new_checkin_entry},
+        },
+    )
+
+    if not update_status:
+        raise HTTPException(status.HTTP_404_NOT_FOUND)
```
taesungh, line 253: Chore: could we include a log statement indicating the associate checked in the participant?

apps/api/src/routers/admin.py
```diff
@@ -210,6 +210,49 @@ async def confirm_attendance() -> None:
...
+@router.post("/checkin/{uid}")
+async def checkin(
+    uid: str,
+    associate: User = Depends(
+        require_role([Role.DIRECTOR, Role.ORGANIZER, Role.VOLUNTEER])
+    ),
+) -> None:
+    new_checkin_entry = {"uid": associate.uid, "time": datetime.now()}
+
+    update_status = await mongodb_handler.raw_update_one(
```
taesungh, line 244: Issue: a check-in should not be permitted if a user does not have a status of `ATTENDING`.

apps/api/src/routers/admin.py
```diff
@@ -210,6 +210,61 @@ async def confirm_attendance() -> None:
...
+
+
+@router.post("/checkin/{uid}")
+async def checkin(
+    uid: str,
+    associate: User = Depends(
+        require_role([Role.DIRECTOR, Role.ORGANIZER, Role.VOLUNTEER])
+    ),
+) -> None:
+    """Check-in applicant at IrvineHacks"""
```
samderanova, line 243: Very small grammar nitpick: Should be "Check in" instead of "check-in", since it's being used as a verb.

taesungh (commented):
Could you also fix the merge conflict? Please place the new check-in routes **after** the waitlist_release route.

apps/api/src/routers/admin.py
```diff
@@ -210,6 +210,61 @@ async def confirm_attendance() -> None:
...
+    record: Optional[dict[str, object]] = await mongodb_handler.retrieve_one(
+        Collection.USERS, {"_id": uid, "role": Role.APPLICANT}
+    )
+    if not record:
+        raise HTTPException(status.HTTP_404_NOT_FOUND)
+    elif record["status"] != Status.ATTENDING:
+        raise RuntimeError("Applicant status must be ATTENDING")
```
samderanova, line 250: Issue: Instead of raising a runtime error, a `403 Forbidden` status should be returned (`HTTPException`).

apps/api/src/routers/admin.py
```diff
@@ -210,6 +210,61 @@ async def confirm_attendance() -> None:
...
+        {"_id": uid},
+        {
+            "$push": {"checkins": new_checkin_entry},
+        },
+    )
+
+    if not update_status:
+        raise HTTPException(status.HTTP_404_NOT_FOUND)
+    else:
+        log.info(f"Applicant {uid} checked in by {associate.uid}")
```
samderanova, line 265: Nitpick: No need to include `else` here since the log statement will run only if `update_status` is a truthy value.

apps/api/src/routers/admin.py
```diff
@@ -210,6 +210,61 @@ async def confirm_attendance() -> None:
...
+    update_status = await mongodb_handler.raw_update_one(
+        Collection.USERS,
+        {"_id": uid},
+        {
+            "$push": {"checkins": new_checkin_entry},
+        },
+    )
+
+    if not update_status:
+        raise HTTPException(status.HTTP_404_NOT_FOUND)
```
samderanova, line 263: Issue: 404 isn't exactly the best status code to raise here. If we get to this point, then the user must exist in the database and has a status of `ATTENDING`, so if `raw_update_one` returns a falsy value, that means something is seriously wrong. A 500 would therefore be better.

samderanova (changes requested):
A few more things to address:

taesungh:
I'll go ahead and fix the merge conflict and extract the logic to a new handler.

taesungh (approved):
Can make further adjustments as needed while working on #269 and #266 as a whole.

samderanova (approved):

PR #49, opened 2023-11-30
Add form links to mentor and volunteer section

samderanova:
Replace the placeholder links in `MentorVolunteer.tsx` with the appropriate links.

PR #329, opened 2024-01-23
feat: added incident report and devpost for navbar

Bl20052005:


apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -75,8 +75,24 @@ function Navbar({ identity }: NavbarProps) {
 					<NavLinkItem href="/">Schedule</NavLinkItem>
 					<NavLinkItem href="/">Resources</NavLinkItem>
 					<NavLinkItem href="/">Stage</NavLinkItem> */}
+					<NavLinkItem
+						href="http://tinyurl.com/ih24-incident-form"
```
taesungh, line 79: Could we add named redirects for these, see #49 for reference.
Also we could use the actual Google Form short URL (`forms.gle/...`) instead of sending this through TinyURL.

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -75,8 +75,24 @@ function Navbar({ identity }: NavbarProps) {
...
+						target="_blank"
+						rel="noopener noreferrer"
+					>
+						Devpost
+					</NavLinkItem>
 					{!status && !deadlinePassed && (
-						<NavLinkItem href="/apply">Apply</NavLinkItem>
+						<>
+							<NavLinkItem href="/apply">Apply</NavLinkItem>
+						</>
```
taesungh, line 95: Chore: unnecessary fragment

taesungh (changes requested):

apps/site/src/lib/components/Navbar/Navbar.tsx
```diff
@@ -75,8 +75,24 @@ function Navbar({ identity }: NavbarProps) {
...
 					<NavLinkItem href="/">Stage</NavLinkItem> */}
+					<NavLinkItem
+						href="http://tinyurl.com/ih24-incident-form"
+						target="_blank"
+						rel="noopener noreferrer"
+					>
+						Report Incident
+					</NavLinkItem>
+					<NavLinkItem
+						href="https://irvinehacks-2024.devpost.com/"
```
samderanova, line 86: Could we add a redirect in `next.config.js` so that `https://irvinehacks.com/devpost` redirects to this URL?

samderanova (changes requested):

apps/site/src/lib/components/Navbar/Navbar.tsx
Bl20052005: I could not find the short form of forms.gle, but the full link for docs.google.com/forms does exist

apps/site/next.config.js
```diff
@@ -44,6 +44,17 @@ const nextConfig = {
 				destination: DOCUSIGN_FORM_URL,
 				permanent: true,
 			},
+			{
+				source: "/incident",
+				destination:
+					"https://docs.google.com/forms/d/e/1FAIpQLSehTGeZxoYy5-zQ-dBeEXjvAzy2kc25X8-74LBWbYbW-xMJNQ/viewform",
```
samderanova, line 50: Please use https://forms.gle/A6BdsSzYSiyeTP8Y6 for the incident response form instead. I was able to get this by going to the edit view of the form, clicking "Send", then clicking the link tab and "Shorten URL".

samderanova (changes requested):

apps/site/next.config.js
Bl20052005: I do not have access to the edit view, so maybe that's why

samderanova (approved):
Thanks for working on this. Please squash merge after a second approval.

IanWearsHat (approved):

PR #332, opened 2024-01-23
update: add other non-hacker roles

### LINKED ISSUE #331: Add `Role`s for other non-hackers
We'll want to add roles for non-hackers including mentors, volunteers, sponsors, judges, and workshop leads. Currently, mentors and volunteers have been added but we'll need to add the rest.

samderanova:
Resolves #331


taesungh (approved):
Not sure how I feel about the term "workshop lead", but that's what logistics uses. Equivalent roles added to Admin site in #335.

Issue #340, opened 2024-01-23
[PMS] Incorporate waitlist promotion system into Participants page


taesungh:
For our walk-in hackers, check-in leads will need to be allowed to promote waitlisted hackers as an action on the Participants page.

- [x] Include hackers with intermediate statuses (`WAITLISTED`, `ACCEPTED`, `WAIVER_SIGNED`, `CONFIRMED`) in the `/admin/participants` endpoint
- [x] Show **Promote** action instead of **Check In** for participants with a `WAITLISTED` status
- [x] Disable the **Promote** action for organizers who are not check-in leads
    - Include a tooltip suggesting that only check-in leads are permitted to promote walk-ins
- [x] Disable the **Check In** action for participants with `WAIVER_SIGNED` status
    - Include a tooltip suggesting that the hacker must confirm attendance in their portal
    - Will need to be careful with roles and how #318 ends up
- [x] Allow check-in leads to use the Promote action which shows a modal with a confirmation button and instructions to provide the participant (log into portal, sign waiver, confirm attendance, etc.)
- [x] Allow check-in associates to use `/admin/checkin` endpoint on `CONFIRMED` participants
    - Previously considering making only leads able to verify walk-ins, but only leads can promote anyway

PR #342, opened 2024-01-24
API check-in system modifications

samderanova:
Completes the API side of the tasks mentioned in #340. Waiting on the modal in the admin frontend to be completed first by #341.

## Changes
- Include hackers with status (`ACCEPTED`, `WAITLISTED`, `WAIVER_SIGNED`, `CONFIRMED`) in `/admin/participants` endpoint
- Allow check-in leads, along with directors to use the `/admin/checkin` endpoint on `CONFIRMED` participants

### DIFF
--- a/apps/api/src/admin/participant_manager.py
+++ b/apps/api/src/admin/participant_manager.py
@@ -2,6 +2,7 @@ from logging import getLogger
 from typing import Any, Optional
 
 from auth.user_identity import User, utc_now
+from models.ApplicationData import Decision
 from services import mongodb_handler
 from services.mongodb_handler import Collection
 from utils.user_record import Role, Status, UserRecord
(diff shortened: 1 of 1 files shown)


apps/api/src/admin/participant_manager.py
```diff
@@ -18,10 +19,21 @@ class Participant(UserRecord):
 
 
 async def get_attending_applicants() -> list[Participant]:
-    """Fetch all applicants who have a status of ATTENDING"""
+    """Fetch all applicants who have a status of ATTENIDNG, WAIVER_SIGNED, CONFIRMED,
+    or WAITLISTED."""
```
taesungh, line 23: Thought: unsure if function name `get_attending_applicants` is still accurate.
Chore: typo on `ATTENIDNG`

taesungh (approved):
Nice work, please update the task list on #340 once merged.

apps/api/src/admin/participant_manager.py
```diff
@@ -17,11 +18,22 @@ class Participant(UserRecord):
     status: Status
```
taesungh, line 18: Issue: validation will fail since `Decision.Waitlisted` is a possible "status" value, so we'll need to use a `Union` here.

apps/api/src/admin/participant_manager.py
```diff
@@ -14,14 +15,25 @@ class Participant(UserRecord):
...
         Collection.USERS,
-        {"role": Role.APPLICANT, "status": Status.ATTENDING},
+        {
+            "role": Role.APPLICANT,
+            "status": {
+                "$in": [
+                    Status.ATTENDING,
+                    Status.WAIVER_SIGNED,
+                    Status.CONFIRMED,
+                    Decision.WAITLISTED,
```
taesungh, line 33: Issue: during the walk-in promotion process, there will be a slight moment when the user's status will be `Decision.ACCEPTED`, so we should account for that to show in the list of participants.

taesungh (approved):

Issue #322, opened 2024-01-22
[API] Add Admin endpoint for directors to confirm attendance of a non-hacker


samderanova:
Non-hackers (mentors, volunteers, sponsors, etc.) are special cases as they are not tracked in our application. There should be an endpoint to confirm their attendance.
- Add new Admin endpoint which directors can use to update the status of non-hackers from `WAIVER_SIGNED` to `ATTENDING`
  - Important: user must have a valid role along with status `WAIVER_SIGNED`
- Participant action button to be updated separately

PR #347, opened 2024-01-25
Feature/confirm attendance non hacker

njhuey:
Resolves https://github.com/HackAtUCI/irvinehacks-site-2024/issues/322.

Implements logic to check in non-hackers.


taesungh:
We'll still need to check `attending` status on non-hackers, so I think most of the changes here are unnecessary. One thing we do need to add is checking that participants have a non-empty role, but other than that, I think we can just adapt the existing check-in function.

taesungh:
Apologies for the confusion, I'll update the description on #322, please work on #355 instead.

PR #350, opened 2024-01-25
Make Schedule visible on Navbar and fix countdown and scheule page

waalbert:
- Fixes countdown so that fireworks only show up after IH ends (1/28 at 12pm PST).
- Fixes ordering of the events so now events are sorted by start date then end date.
- Adds Schedule to the Navbar
- There is still a lingering issue where the fireworks will block interaction with the Schedule Tabs, but this can be taken care of in another PR


samderanova (dismissed):

taesungh:
Could you address the merge conflict

samderanova (approved):

PR #357, opened 2024-01-25
feat: add check in capability for non hackers

### LINKED ISSUE #355: [API] Allow checking in other types of participants
Following from #268, we added an endpoint to check in hackers, but we still need to be able to check in non-hackers. In conjunction with #322 as part of #318, we'll assume for now the status of non-hackers will be updated to `CONFIRMED` or `ATTENDING` (exact TBD), so the logic is almost the same as for non-hackers. The only additional check we should need to make is that the user role is not empty.
- Repurpose `participant_manager.check_in_applicant` to allow checking in non-hackers.
- Make sure the user's role is not empty during the database query
- Use the same existing status check (`ATTENDING` or `CONFIRMED`)

njhuey:
Resolve #355

### DIFF
--- a/apps/api/src/admin/participant_manager.py
+++ b/apps/api/src/admin/participant_manager.py
@@ -47,12 +47,16 @@ async def get_hackers() -> list[Participant]:
     return [Participant(**user, **user["application_data"]) for user in records]
 
 
-async def check_in_applicant(uid: str, associate: User) -> None:
-    """Check in applicant at IrvineHacks"""
+async def check_in_participant(uid: str, associate: User) -> None:
+    """Check in participant at IrvineHacks"""
     record: Optional[dict[str, object]] = await mongodb_handler.retrieve_one(
-        Collection.USERS, {"_id": uid, "role": Role.APPLICANT}
+        Collection.USERS, {"_id": uid, "role": {"$exists": True}}
     )
-    if not record or record["status"] not in (Status.ATTENDING, Status.CONFIRMED):
+
+    if not record or record.get("status", "") not in (
+        Status.ATTENDING,
+        Status.CONFIRMED,
+    ):
         raise ValueError
 
     new_checkin_entry = (utc_now(), associate.uid)
--- a/apps/api/src/routers/admin.py
+++ b/apps/api/src/routers/admin.py
@@ -253,13 +253,12 @@ async def participants() -> list[Participant]:
 
 
 @router.post("/checkin/{uid}")
-async def checkin(
+async def check_in_participant(
     uid: str, associate: Annotated[User, Depends(require_checkin_associate)]
 ) -> None:
     """Check in participant at IrvineHacks."""
     try:
-        # TODO: non-hackers
-        await participant_manager.check_in_applicant(uid, associate)
+        await participant_manager.check_in_participant(uid, associate)
     except ValueError:
         raise HTTPException(status.HTTP_404_NOT_FOUND)
     except RuntimeError as err:


apps/api/src/admin/participant_manager.py
```diff
@@ -52,7 +52,15 @@ async def check_in_applicant(uid: str, associate: User) -> None:
     record: Optional[dict[str, object]] = await mongodb_handler.retrieve_one(
         Collection.USERS, {"_id": uid, "role": Role.APPLICANT}
```
taesungh, line 53: Issue: we'll need to update the query to select roles other than just applicant. I think we could just remove the `role` selector all together since we're checking status anyway, and it's not possible for anyone to have reached the final two statuses, but for completeness, `{"role": {"$exists": True}}` could also work.

apps/api/src/routers/admin.py
```diff
@@ -258,7 +258,6 @@ async def checkin(
 ) -> None:
     """Check in participant at IrvineHacks."""
     try:
-        # TODO: non-hackers
         await participant_manager.check_in_applicant(uid, associate)
```
taesungh, line 261: Chore: could you update the function name (and docstring) to be more generic, e.g. `check_in_participant`.

apps/api/src/admin/participant_manager.py
```diff
@@ -52,7 +52,15 @@ async def check_in_applicant(uid: str, associate: User) -> None:
...
     )
-    if not record or record["status"] not in (Status.ATTENDING, Status.CONFIRMED):
+    if not record:
+        raise ValueError
+
+    try:
+        status = record["status"]
+    except KeyError:
+        raise RuntimeError
+
```
taesungh, line 62: Rather than allowing a `KeyError` to happen, we can ensure the key exists with `"status" in record`, or use `dict.get` to retrieve a default value.
```suggestion
    status = record.get("status")
```

taesungh (changes requested):
Nice start, need to update the query to select all roles.

apps/api/src/routers/admin.py
taesungh: I meant the name of the function in `participant_manager`

apps/api/src/admin/participant_manager.py
```diff
@@ -50,9 +50,16 @@ async def get_hackers() -> list[Participant]:
...
+        Collection.USERS, {"_id": uid, "role": {"$exists": True}}
     )
-    if not record or record["status"] not in (Status.ATTENDING, Status.CONFIRMED):
+    if not record:
+        raise ValueError
+
+    status = record.get("status", "")
+    if not status:
+        raise RuntimeError
+
```
taesungh, line 61: A `RuntimeError` is usually used in our codebase to indicate an exception out of our control such as a database operation failing. We can use the same `ValueError` here by just omitting these lines and letting the next `not in` condition handle it since `None` is not in the tuple.

njhuey: QQ: Since we have updated all Users to have a status attribute, wouldn't a user not having a status signify an internal server error? If I switch this to raise a `ValueError`, a 404 will be returned which doesn't seem right.

taesungh: There are some user records who actually won't have statuses such as people who logged in but never applied. We could change the parent to status 400 if that would make the error more generic, but I think 404 still kinda makes sense, as in the person the user asked to check is found not be a participant?

njhuey: Ah gotcha, I was missing that part. I'll go ahead and make the change.

samderanova (approved):

taesungh (approved):
Should work out.

PR #358, opened 2024-01-26
fix: schedule tabs can be clicked on through fireworks

IanWearsHat:


samderanova (changes requested):
Issue: There's an extra scrollbar on the right side of the screen when the fireworks appear. Could we apply `overflow: hidden` on them?

waalbert (approved):
Thanks for fixing this!

samderanova (approved):
Thanks for working on this. Will squash merge.

PR #362, opened 2024-01-26
feat: add confirm attendance route for non-hackers

### LINKED ISSUE #322: [API] Add Admin endpoint for directors to confirm attendance of a non-hacker

njhuey:
Resolves #322

### DIFF
--- a/apps/api/src/admin/participant_manager.py
+++ b/apps/api/src/admin/participant_manager.py
@@ -68,3 +68,33 @@ async def check_in_applicant(uid: str, associate: User) -> None:
         raise RuntimeError(f"Could not update check-in record for {uid}.")
 
     log.info(f"Applicant {uid} checked in by {associate.uid}")
+
+
+async def confirm_attendance_non_hacker(uid: str, director: User) -> None:
+    """Update status for Role.Attending for non-hackers."""
+    allowed_roles = (
+        Role.MENTOR,
+        Role.ORGANIZER,
+        Role.VOLUNTEER,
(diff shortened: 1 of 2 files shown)


apps/api/src/admin/participant_manager.py
```diff
@@ -68,3 +68,33 @@ async def check_in_applicant(uid: str, associate: User) -> None:
...
+        Role.SPONSOR,
+        Role.JUDGE,
+        Role.WORKSHOP_LEAD,
+    )
+
+    record: Optional[dict[str, object]] = await mongodb_handler.retrieve_one(
+        Collection.USERS, {"_id": uid, "status": Status.WAIVER_SIGNED}
+    )
+
+    if not record or record["status"] not in allowed_roles:
```
taesungh, line 88: Issue: we need to check for `role`, not `status`

taesungh (commented):
Good work so far, one issue/typo and one suggestion on logging.

taesungh (dismissed):
Thanks for working on this.

taesungh (approved):

samderanova (approved):

Issue #344, opened 2024-01-24
[PMS] Display check-in status for each day of the event


taesungh:
Following from #341, to indicate who has checked into the event, we can have three columns, one for each day, based on the check-in entries provided by the API. To offload the work from the API, the actual computation of whether or not a participant has checked in on a given day can happen on the client. To avoid a full data refetch, we can [Mutate Based on Current Data](https://swr.vercel.app/docs/mutation#mutate-based-on-current-data) in the `useParticipants` hook where the `/admin/checkin/{uid}` endpoint returns the update data for the current participant.
- Make checkin endpoint return updated check-in entries for the current participant
- Implement the mutation in `useParticipant` to update the data for only the current participant
- Add three new columns to the participants table showing an icon indicating if a participant has checked in for that day

taesungh:
Considering as completed in #363. Will create new issue about implementing minimal mutations.

PR #363, opened 2024-01-26
Show check-ins per day in Participants table

taesungh:
As part of #344.
- Provide checkins list in participants endpoint
- Add columns for each day of the event with an icon on check-in status
- Disable resizable columns

## Testing
1. Add sample applicant records with `ATTENDING` status if you don't have them already
2. Load the Participants page on the Admin site
3. Observe the check-in icons are all pending
4. Check in an applicant
5. Refocus the page to trigger a refetch (since we don't have the mutation yet)
6. Observe the icon for Friday has changed to positive

### DIFF
--- a/apps/site/src/app/admin/participants/components/ParticipantsTable.tsx
+++ b/apps/site/src/app/admin/participants/components/ParticipantsTable.tsx
@@ -9,8 +9,13 @@ import TextFilter from "@cloudscape-design/components/text-filter";
 import ApplicantStatus from "@/app/admin/applicants/components/ApplicantStatus";
 import { Participant } from "@/lib/admin/useParticipants";
 
+import CheckinDayIcon from "./CheckinDayIcon";
 import ParticipantAction from "./ParticipantAction";
 import RoleBadge from "./RoleBadge";
+
+const FRIDAY = new Date("2024-01-26T12:00:00");
+const SATURDAY = new Date("2024-01-27T12:00:00");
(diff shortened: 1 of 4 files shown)


apps/site/src/app/admin/participants/components/ParticipantsTable.tsx
```diff
@@ -9,9 +9,14 @@ import TextFilter from "@cloudscape-design/components/text-filter";
 import ApplicantStatus from "@/app/admin/applicants/components/ApplicantStatus";
 import { Participant } from "@/lib/admin/useParticipants";
 
+import CheckinDayIcon from "./CheckinDayIcon";
 import ParticipantAction from "./ParticipantAction";
 import RoleBadge from "./RoleBadge";
 
+const FRIDAY = new Date("2024-01-26T12:00:00");
+const SATURDAY = new Date("2024-01-27T12:00:00");
+const SUNDAY = new Date("2024-01-28T12:00:00");
```
samderanova, line 18: Do we need to use UTC here to prevent potential timezone issues like in the schedule countdown?

taesungh: The timezones are encoded within the `CheckinDayIcon`, and since the time of noon is used, any timezone shift should still be within the same day.

samderanova (approved):

IanWearsHat (approved):

PR #366, opened 2024-01-26
Add non-hackers to `/participants` endpoint

### LINKED ISSUE #364: [API] Provide non-hackers in participants list
Following from #268 and together with #355, we'll need to provide the list of non-hackers in the `/admin/participants` endpoint. To ensure security, this should include only participants with a valid role but with any status (in practice would only be `WAIVER_SIGNED` or `ATTENDING`).
- Create new function in `participant_manager` to fetch records of non-hacker roles from the database
- Return concatenation of non-hackers and hackers in `/admin/participants`

samderanova:
Resolves #364.

## Changes
- Create a new function `get_non_hackers` in `participant_manager` to pull non-hackers from the database
  - The query for this function checks whether a role exists and if it does, whether that role is not an applicant or organizer role
- Return concatenation of hackers and non-hackers in the function for the `/admin/participants` route

### DIFF
--- a/apps/api/src/admin/participant_manager.py
+++ b/apps/api/src/admin/participant_manager.py
@@ -23,7 +23,7 @@ class Participant(UserRecord):
 
     first_name: str
     last_name: str
-    status: Optional[Union[Status, Decision]] = None
+    status: Union[Status, Decision] = Status.REVIEWED
 
 
 async def get_hackers() -> list[Participant]:
--- a/apps/api/src/admin/participant_manager.py
+++ b/apps/api/src/admin/participant_manager.py
@@ -52,24 +52,15 @@ async def get_non_hackers() -> list[Participant]:
     records: list[dict[str, Any]] = await mongodb_handler.retrieve(
         Collection.USERS,
         {
-            "$and": [
-                {"role": {"$exists": True}},
-                {
-                    "role": {
-                        "$not": {
-                            "$regex": "applicant|"
-                            + "|".join(
-                                (
-                                    Role.DIRECTOR,
-                                    Role.REVIEWER,
-                                    Role.CHECKIN_LEAD,
-                                    Role.ORGANIZER,
-                                )
-                            )
-                        }
-                    }
-                },
-            ]
+            "role": {
+                "$in": [
+                    Role.JUDGE,
+                    Role.SPONSOR,
+                    Role.MENTOR,
+                    Role.VOLUNTEER,
+                    Role.WORKSHOP_LEAD,
+                ]
+            }
         },
         ["_id", "status", "role", "first_name", "last_name"],
     )
--- a/apps/api/src/admin/participant_manager.py
+++ b/apps/api/src/admin/participant_manager.py
@@ -9,6 +9,14 @@ from utils.user_record import Role, Status, UserRecord
 
 log = getLogger(__name__)
 
+NON_HACKER_ROLES = (
+    Role.MENTOR,
+    Role.VOLUNTEER,
+    Role.SPONSOR,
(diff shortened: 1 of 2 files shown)


apps/api/src/routers/admin.py
```diff
@@ -248,8 +248,9 @@ async def waitlist_release(uid: str) -> None:
 @router.get("/participants", dependencies=[Depends(require_checkin_associate)])
 async def participants() -> list[Participant]:
     """Get list of participants."""
-    # TODO: non-hackers
-    return await participant_manager.get_hackers()
+    return (await participant_manager.get_hackers()) + (
+        await participant_manager.get_non_hackers()
+    )
```
taesungh, line 253: Suggestion/nitpick: might look cleaner to extract these two to variables instead of inlining the `await`s

apps/api/src/admin/participant_manager.py
```diff
@@ -47,6 +47,35 @@ async def get_hackers() -> list[Participant]:
...
+                                (
+                                    Role.DIRECTOR,
+                                    Role.REVIEWER,
+                                    Role.CHECKIN_LEAD,
+                                    Role.ORGANIZER,
+                                )
+                            )
+                        }
+                    }
+                },
```
taesungh, line 71: Was there a particular reason you chose to use a exclusive filter here? Think it would be simpler to just get the exact roles we need.

apps/api/src/admin/participant_manager.py
```diff
@@ -15,7 +15,7 @@ class Participant(UserRecord):
 
     first_name: str
     last_name: str
-    status: Union[Status, Decision]
+    status: Optional[Union[Status, Decision]] = None
```
taesungh, line 18: Question: in what cases would a participant not have a status?

taesungh (commented):
Some interesting choices, think it could be simplified.

apps/api/src/admin/participant_manager.py
samderanova: I had to alter this to fit with non-hackers because there are some non-hackers who haven't signed the waiver yet and therefore don't have a status.

apps/api/src/admin/participant_manager.py
```diff
@@ -52,24 +52,15 @@ async def get_non_hackers() -> list[Participant]:
...
-                },
-            ]
+            "role": {
+                "$in": [
+                    Role.JUDGE,
+                    Role.SPONSOR,
+                    Role.MENTOR,
+                    Role.VOLUNTEER,
+                    Role.WORKSHOP_LEAD,
+                ]
```
taesungh, line 62: Sorry for the merge conflicts but could you rebase with latest `main` and then refactor the new `allowed_roles` from #362 into a shared global?

apps/api/src/admin/participant_manager.py
taesungh: Hmm, do you think you could manually assign a status to those who don't? Don't wanna mess too much with entire system right now, so maybe something like `REVIEWED`, even though it technically doesn't make sense?

samderanova: How about `WAIVER_UNSIGNED`?

samderanova: Oh never mind, that's not part of `Status`. I'll give them `REVIEWED`.

taesungh: Oh i meant the database records, but I suppose this also technically works.

taesungh (approved):
Approving to unblock, thanks for the patience with the rebase and merge conflict. Will have another conflict with #363.

PR #368, opened 2024-01-26
Confirm non-hacker attendance from PMS

### LINKED ISSUE #356: [PMS] Include confirm attendance option for non-hackers
Following from #322, we'll need to update the `ParticipantAction` button to allow confirming the attendance of a non-hacker.
- Update `ParticipantAction` to show **Confirm** button when participant's role is not `applicant`
  - Send POST request to endpoint from #322
- Disable the button when the user is not a director and show a corresponding tooltip/popover
  - See #340 and #346 for reference
- A modal would be nice but is probably unnecessary
  - Directors can notify tech if a confirmation needs to be undone for some reason

samderanova:
Resolves #356.

## Changes
- Update ParticipantAction to show Confirm button when participant's role is not applicant
  - Send POST request to endpoint from https://github.com/HackAtUCI/irvinehacks-site-2024/issues/322
- Disable the button when the user is not a director and show a corresponding tooltip/popover

### DIFF
--- a/apps/site/src/app/admin/participants/components/ParticipantAction.tsx
+++ b/apps/site/src/app/admin/participants/components/ParticipantAction.tsx
@@ -61,7 +61,7 @@ function ParticipantAction({
 		</Button>
 	);
 
-	if (nonHacker) {
+	if (nonHacker && isWaiverSigned) {
 		if (role !== "director") {
 			return (
 				<ParticipantActionPopover content="Only directors are allowed to confirm non-hackers.">
--- a/apps/site/src/app/admin/participants/components/ParticipantAction.tsx
+++ b/apps/site/src/app/admin/participants/components/ParticipantAction.tsx
@@ -3,7 +3,7 @@ import { useContext } from "react";
 import Button from "@cloudscape-design/components/button";
 
 import UserContext from "@/lib/admin/UserContext";
-import { isCheckinLead } from "@/lib/admin/authorization";
+import { isCheckinLead, isNonHacker } from "@/lib/admin/authorization";
 import { Status } from "@/lib/admin/useApplicant";
 import { Participant } from "@/lib/admin/useParticipants";
 import ParticipantActionPopover from "./ParticipantActionPopover";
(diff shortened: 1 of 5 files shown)


apps/site/src/lib/admin/useParticipants.ts
```diff
@@ -38,12 +38,20 @@ function useParticipants() {
 		console.log("Checking in", participant);
 		// TODO: implement mutation for showing checked in on each day
 		await axios.post(`/api/admin/checkin/${participant._id}`);
+		mutate(data);
```
taesungh, line 41: Issue: I'm pretty sure if you provide a data object, it needs to be the updated data. We can do a full mutation by not providing anything, see `submitReview` in `useApplicant`.

apps/site/src/app/admin/participants/components/ParticipantAction.tsx
```diff
@@ -47,7 +50,27 @@ function ParticipantAction({
...
+
+	if (nonHacker) {
+		if (role !== "director") {
+			return (
+				<ParticipantActionPopover content="Only directors are allowed to confirm non-hackers.">
+					{confirmButton}
+				</ParticipantActionPopover>
+			);
+		}
+		return confirmButton;
```
taesungh, line 72: Issue: confirmation button should be shown only if status is `WAIVER_SIGNED`, should be allowed to check-in otherwise.

apps/site/src/lib/admin/authorization.ts
```diff
@@ -16,3 +23,7 @@ export function isAdminRole(role: string | null) {
 export function isCheckinLead(role: string | null) {
 	return role !== null && CHECKIN_ROLES.includes(role);
 }
+
+export function isNonHacker(role: string | null) {
+	return role !== null && NONHACKER_ROLES.includes(role);
+}
```
taesungh, line 29: Nitpick [non-blocking]: while this is related to roles, it's not really related to authorization. At the moment, not sure where else to put this, maybe just add a comment suggesting to refactor elsewhere later.

taesungh (changes requested):
Nice start, need to account for other cases with confirm button.

waalbert (approved):

taesungh (approved):
Works for now. Separate issue for the default `REVIEWED` status used in #366.

PR #373, opened 2024-01-26
fix: non-hacker confirm button disable conditions

### LINKED ISSUE #371: [PMS] Non-Hacker Check In Fixes
Following from #368, we need to fix the disable condition for `confirmButton` to match its corresponding condition for showing `ParticipantActionPopover`. Additionally, we can simply the `confirm_attendance_non_hacker` function to use `update_one` instead of `raw_update_one`.

samderanova:
Resolves #371.

### DIFF
--- a/apps/site/src/app/admin/participants/components/ParticipantAction.tsx
+++ b/apps/site/src/app/admin/participants/components/ParticipantAction.tsx
@@ -55,7 +55,7 @@ function ParticipantAction({
 			variant="inline-link"
 			ariaLabel={`Confirm attendance for ${participant._id}`}
 			onClick={() => initiateConfirm(participant)}
-			disabled={!isCheckin}
+			disabled={nonHacker && isWaiverSigned && role !== "director"}
 		>
 			Confirm
 		</Button>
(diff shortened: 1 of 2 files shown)


apps/site/src/app/admin/participants/components/ParticipantAction.tsx
```diff
@@ -55,16 +55,20 @@ function ParticipantAction({
 			variant="inline-link"
 			ariaLabel={`Confirm attendance for ${participant._id}`}
 			onClick={() => initiateConfirm(participant)}
-			disabled={!isCheckin}
+			disabled={nonHacker && isWaiverSigned && role !== "director"}
```
taesungh, line 58: Sorry, can you also allow check-in leads to confirm following from #372.

apps/site/src/app/admin/participants/components/ParticipantAction.tsx
```diff
@@ -55,16 +55,19 @@ function ParticipantAction({
 			variant="inline-link"
 			ariaLabel={`Confirm attendance for ${participant._id}`}
 			onClick={() => initiateConfirm(participant)}
-			disabled={!isCheckin}
+			disabled={!isCheckinLead}
```
taesungh, line 58: Issue: should be value not function?

taesungh (approved):

Issue #360, opened 2024-01-26
[PMS] Incorporate barcode input with check-in system


taesungh:
To track badges of participants during subevents, we'll need to register their barcodes with the system.
- Update check-in endpoint to accept a badge number and store this on the user record
- In the check-in modal, include an input field for the participant's badge number
- Integrate a barcode scanning system to read the barcode through the user's camera

taesungh:
Interim version shipped in #374.

taesungh:
Considering as completed in #374 and #380, used during event last year. Possible improvements if desired again in the future are fixing double camera issue in strict mode and minor styling quirks.

PR #378, opened 2024-01-27
Display badge numbers in Participants table

taesungh:
As part of #360
- Include column of badge numbers in particpants table
- Allows searching by badge number for potentially faster check-in


alanchangxyz (approved):
functionality/code wise looks good, maybe we can `<Badge></Badge>` these like with participant roles but not big deal

njhuey (approved):
Yes Chef

PR #379, opened 2024-01-27
Set `minWidth` on participant actions column

taesungh:
As part of #269.
- To avoid text wrapping on Check In/Promote action text
- Include `ColumnDefinition` type


IanWearsHat (approved):

njhuey (approved):
Yes Chef

Issue #168, opened 2023-12-30
Next.js build complains about dynamic server usage


taesungh:
While running `next build` for the site, Next tries rendering the site content as static pages but encounters a `DynamicServerError` since our `getUserIdentity` function relies on `cookies` to personalize the content for each user.
> DynamicServerError: Dynamic server usage: Page couldn't be rendered statically because it used `cookies`. See more info here: https://nextjs.org/docs/messages/dynamic-server-error

We'll need to make sure the cookies are properly contained in the async context and bound to the same call stack.

taesungh:
Note: this may become irrelevant with the new [`dynamicIO` flag in Next.js 15 Canary](https://nextjs.org/docs/app/api-reference/config/next-config-js/dynamicIO), but that is currently still experimental.

Noticing we currently don't access the API from server-side rendering except for [`getUserIdentity`](https://github.com/HackAtUCI/irvinehacks-site/blob/3752272e3bba64e62ded6441b3a79d095e11ed2d/apps/site/src/lib/utils/getUserIdentity.ts), so a short-term solution may be forwarding the cookies there instead of using a [request interceptor on the Axios instance](https://github.com/HackAtUCI/irvinehacks-site/blob/3752272e3bba64e62ded6441b3a79d095e11ed2d/apps/site/src/lib/utils/api.ts#L13).

Alternatively, we can explicitly specify dynamic rendering on such pages with the [`dynamic` Route Segment Config](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamic):
```tsx
const dynamic = 'force-dynamic'
```

taesungh:
Can fix by including a dynamic function (e.g. `cookies`) inside `getUserIdentity` that is unused but indicates that dynamic rendering is needed. Will handle while working on #524.

PR #403, opened 2024-12-01
Updated dependencies versions

Bl20052005:
Updated next.js v13 -> 14, framer motion v10 -> 11

technically framer motion is supposed to work with React 17 and next 13; however there seems to be no issue for now.

Other dependencies don't need to be updated as far as I can tell


taesungh:
Looks like we need to handle #168 before being able to upgrade to Next.js 14.

taesungh:
Actually I misread, there's an issue with CSR bailout:
> site:build:  ⨯ useSearchParams() should be wrapped in a suspense boundary at page "/guest-login". Read more: https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout

but I think Next.js gets confused because it thinks the route is statically rendered when in practice, dynamic rendering is happening.

PR #404, opened 2024-12-01
Update/constrain database query for user identity

### LINKED ISSUE #402: Constrain database query for user identity
The `/user/me` endpoint currently queries the database for all fields to obtain the user record. In practice, only the `role` and `status` fields are needed in the response, so the query should specify those fields to avoid using unnecessary bandwidth.

https://github.com/HackAtUCI/irvinehacks-site/blob/7e8175001f074a0e715ab6af0a8734066691173f/apps/api/src/routers/user.py#L69-L71

For better type checking, it might also help to make the return value an explicit `IdentityResponse` rather than relying on `response_model` for data filtering.

IanWearsHat:
Closes #402 

- Added query projection to `mongodb_handler.retrieve_one` call in `/me` route
- Replaced FastAPI `response_model` with explicit `IdentityResponse`  
- Added test to check that `IdentityResponse` fields are returned from `/me` route

### DIFF
--- a/apps/api/src/routers/user.py
+++ b/apps/api/src/routers/user.py
@@ -67,7 +67,7 @@ async def me(
     if not user:
         return dict()
     user_record = await mongodb_handler.retrieve_one(
-        Collection.USERS, {"_id": user.uid}
+        Collection.USERS, {"_id": user.uid}, ["role", "status"]
     )
     if not user_record:
         return {"uid": user.uid}
--- a/apps/api/src/routers/user.py
+++ b/apps/api/src/routers/user.py
@@ -65,7 +65,7 @@ async def me(
 ) -> IdentityResponse:
     log.info(user)
     if not user:
-        return dict()
+        return IdentityResponse()
     user_record = await mongodb_handler.retrieve_one(
         Collection.USERS, {"_id": user.uid}, ["role", "status"]
     )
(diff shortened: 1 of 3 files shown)


apps/api/src/routers/user.py
```diff
@@ -59,19 +59,21 @@ async def logout() -> RedirectResponse:
...
 
-@router.get("/me", response_model=IdentityResponse)
+@router.get("/me")
 async def me(
     user: Annotated[Union[User, None], Depends(use_user_identity)]
-) -> dict[str, object]:
+) -> IdentityResponse:
     log.info(user)
     if not user:
         return dict()
```
waalbert, line 68: Issue: return type should be IdentityResponse

waalbert (changes requested):
Awesome work Ian! One thing to fix

apps/api/src/routers/user.py
taesungh: Weird, mypy should have caught this, will need to troubleshoot the checks workflow.

taesungh: Was getting filtered out to only added lines, removed filter in 6c03c32.

waalbert (approved):
Thanks for the work Ian!

PR #413, opened 2024-12-03
Adding Landing Page

Bl20052005:


apps/site/src/app/(main)/(home)/sections/Landing/LandingUnlocked.tsx
waalbert, line 1: Issue: on larger screen widths `> 1920px`, you can start to see the edges outside of the assets, so to fix this can you split up the assets into halves and load them in that way?

waalbert (changes requested):
Looks awesome Bo! Here are a few fixes

apps/site/src/app/(main)/(home)/sections/Landing/LandingUnlocked.tsx
Bl20052005: This is possible for the pillars and background castle, but there are a couple of assets like clouds and background pillars that cannot be split up as it would make for a very jarring cut-off

Bl20052005: As it turns out this is not needed at all

waalbert (approved):
Thanks for the work Bo and Ian!

waalbert (approved):

PR #414, opened 2024-12-03
Update SendGrid Template IDs

waalbert:
Because we're using SendGrid to send out custom guest login, application confirmation, and decision emails like IH 2024, these email template IDs need to be updated for IH 2025.


apps/api/src/utils/email_handler.py
```diff
@@ -10,8 +10,8 @@
     Template,
 )
 
-IH_SENDER = ("apply@irvinehacks.com", "IrvineHacks 2024 Applications")
-REPLY_TO_HACK_AT_UCI = ("hack@uci.edu", "Hack at UCI")
+IH_SENDER = ("apply@irvinehacks.com", "IrvineHacks 2025 Applications")
+REPLY_TO_HACK_AT_UCI = ("irvinehacks2025@gmail.com", "Hack at UCI")
```
taesungh, line 14: Thought: if we set up forwarding from `apply@irvinehacks.com` to the Gmail address, I think we could get away with not specifying a reply-to address and have the default of replying to the sender.

Issue #416, opened 2024-12-04
Update UI for Guest Passphrase Login


waalbert:
The logic of users entering guest passphrase can be reused from IH 2024. Only the UI needs to change. The UI can be the same as the regular login page from [Figma](https://www.figma.com/design/CyxJ0OSjgNPwb0eeTgq84s/IrvineHacks-2025?version-id=2157408819231333036&node-id=1-2&node-type=frame&t=CakJ2lXb9BwC8802-0)

PR #418, opened 2024-12-05
Update Login UI

### LINKED ISSUE #415: Update UI for Login
The login logic can be reused from IH 2024. Only the login page UI needs to be updated.

IanWearsHat:
Closes #415 and #416
Updated login UI at `/login` and `/guest-login`


waalbert (approved):
Thanks for updating the Login UI Ian!

apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx
```diff
@@ -31,12 +31,20 @@ function ValidatingForm(props: PropsWithChildren<FormProps>) {
...
+				validated ? styles.validated : styles.notYetValidated
+			} flex flex-col items-center`}
 			{...rest}
 		>
 			{children}
-			<Button text="Continue" disabled={submitting} />
+			<div className="w-1/3">
+				<Button
+					className="text-xs sm:text-base md:text-2xl"
+					text="Login"
```
taesungh, line 43: Nitpick: "log in" is being used as a verb, so it should be two words. Personally think having it say "Continue" made more sense since the login process has multiple steps.

apps/site/src/app/(main)/login/Login.tsx
```diff
@@ -1,25 +1,15 @@
...
-			<h1
-				className="font-display text-5xl md:text-7xl mb-10"
-				style={{
-					textShadow: "0px 0px 20px rgba(255, 255, 255, 0.75)",
-				}}
-			>
-				Log In
+		<div className="min-h-screen flex flex-col items-center justify-center">
+			<h1 className="font-display text-3xl md:text-5xl mb-20">
+				Login to Portal
```
taesungh, line 12: Note: the application page redirects here for new users, so it may be confusing to see it saying this goes to the Portal when users are trying to apply.

PR #423, opened 2024-12-05
Restructure first/last name in user records

### LINKED ISSUE #422: Simplify how basic personal information is stored
Another inconvenience from last year was having basic profile data (first name and last name) stored differently for hackers than for non-hackers. This required juggling around the data fields in the participant manager for [`get_hackers`](https://github.com/HackAtUCI/irvinehacks-site/blob/76c767c1e549527dc8620edc27417dbbb0985267/apps/api/src/admin/participant_manager.py#L39) and [`get_non_hackers`](https://github.com/HackAtUCI/irvinehacks-site/blob/76c767c1e549527dc8620edc27417dbbb0985267/apps/api/src/admin/participant_manager.py#L63).

To simplify the data queries, it would be better to store the data in the same way for all participants. This can be modeled as part of [`UserRecord`](https://github.com/HackAtUCI/irvinehacks-site/blob/76c767c1e549527dc8620edc27417dbbb0985267/apps/api/src/utils/user_record.py#L34) by requiring anybody with a role to also have a name. The change should be mostly straightforward since `Applicant` and `Participant` model already inherit from `UserRecord`.

This change should ideally be made before applications are opened so the database records don't need to be retroactively modified.

taesungh:
Resolves #422. Can look into simplify actual participants query later on.

#### Changes
- Move `user_record.py` from `utils` to `models`
- Place first/last name on root of user record for all participants
  - Previously, applicants had their name stored in `application_data`
  - The new structure simplifies how all participants are queried
  - Anybody assigned a role should have a name
- Reorganize ApplicationData models and create new `BareApplicant` model
  - Constrain some database queries by using explicit fields
  - This simplifies some of the unit tests not needing application data
- Update data properties for applicants referenced in Admin site

#### Testing
If you have any old user records in your local database, you will need to update them with the new structure.
1. Impersonate an applicant and submit an application (adjust deadline if needed)
2. Create a user record with an admin role and impersonate that user
3. On the Admin site, observe the applicant is listed properly in the Applicants page and the details for the individual applicant also load properly
4. Manually update the applicant's status to `ATTENDING`
5. Observe the applicant name is listed properly in the Participants page on the Admin site

The waiver submission handler was not tested manually, but test coverage there is pretty good. Other admin-related functionality involving emails and personalizations was not tested (still need to add unit tests).


apps/api/src/admin/participant_manager.py
```diff
@@ -27,13 +27,20 @@ class Participant(UserRecord):
...
 
 
-PARTICIPANT_FIELDS = ["_id", "status", "role", "checkins", "badge_number"]
+PARTICIPANT_FIELDS = [
+    "_id",
+    "first_name",
+    "last_name",
+    "role",
+    "status",
+    "role",
```
waalbert, line 40: Issue: role defined twice

waalbert (changes requested):
Thanks for simplifying the `user_record` model for us Taesung! This will help us simplify user queries in the future. One thing to fix

PR #426, opened 2024-12-05
Update user/apply parameters, delete custom form_body decorator

pandyrew:
Updated user/apply parameters, deleted custom form_body decorator

*misspelled decorator


apps/api/src/routers/user.py
```diff
@@ -79,7 +79,7 @@ async def me(
 @router.post("/apply", status_code=status.HTTP_201_CREATED)
 async def apply(
     user: Annotated[User, Depends(require_user_identity)],
-    raw_application_data: Annotated[RawApplicationData, Depends(RawApplicationData)],
+    raw_application_data: Annotated[RawApplicationData, Form()],
     resume: Optional[UploadFile] = None,
```
taesungh, line 83: Seems this doesn't actually quite work, think the file upload is causing a conflict. Will look into this.

IanWearsHat (approved):

taesungh (approved):

Issue #386, opened 2024-11-24
Release Hacker, Mentor, and Volunteer Applications


waalbert:
All applications will open December 6th.
- Mentor and volunteer applications are moving from Google Forms to our site, so our backend will need to be modified accordingly.
- On the Home Page, a section representing hacker, mentor, or volunteer applications called "Choose your character" needs to be added. 
- The hacker application form can be updated from the last application form. 
  - route `/apply` can stay the same
- Mentor and Volunteer application forms will need to be created. 
  - routes `/mentor` and `/volunteer` need to be created
- On your local dev environment, there's a dev route `/api/dev/impersonate/<uid>` where you can sign in as `<uid>` and simulate this user's identity. 
  - To test application submission, disable `upload_resume` and `send_application_confirmation_email` in the backend api route `/apply` so you can submit applications using this user. 
  - This user will be on your local instance of MongoDB `localhost:8081` under database `irvinehacks-prod` and collection `users`
  - login credentials is username - admin, password - pass

---
General
- [x] #397
- [x] #405
- [x] #410
- [x] #441
- [x] #475 

Other API restructuring
- [x] #421
- [x] #422

Application Flow
- [x] #463
- [x] #459

Portal
- [x] #398

Hacker applications
- [x] #400
- [x] #461
- [x] #467

Mentor and volunteer applications
- [x] #395
- [x] #396
- [x] #470
- [x] #448
- [x] #450
- [x] #452

Unplanned
- [x] #482

To handle later
- #460
- #462

IanWearsHat:
- [ ] #397
- [ ] #396 
- [ ] #395 
- [ ] #398

waalbert:
Closing as all tasks have been completed

PR #429, opened 2024-12-06
Restructure user record to have multiple roles

### LINKED ISSUE #421: Allow users to have multiple roles
While general participants (both hackers and non-hackers) should usually have only one role, one inconvenient construct last year was that users could have only one role. This made defining some of the admin authorization more complicated since we had to consider multiple roles that had a common trait.

For example, the [`isAdminRole`](https://github.com/HackAtUCI/irvinehacks-site/blob/76c767c1e549527dc8620edc27417dbbb0985267/apps/site/src/lib/admin/authorization.ts#L16) function needs to check an array of admin roles and also another for organizer roles. A similar thing happens with [`require_checkin_associate`](https://github.com/HackAtUCI/irvinehacks-site/blob/76c767c1e549527dc8620edc27417dbbb0985267/apps/api/src/routers/admin.py#L29) on the admin API. Both of these could be simplified to anybody with the organizer role, and additional roles can be granted to those who need it.

Allowing multiple roles will make it easier to handle different combinations of roles and also allow for quickly adding roles on the fly (e.g. roles per department if say logistics organizers need special permissions). This will require changes across several places including the `UserRecord` model, the `Participant` interface, the user identity, and subsequent usages of the attribute/property.

One possible way to reduce some of the work could be keeping the role field for the user identity as singular but storing a separate property for additional organizer-related roles. This would align with the concept that organizer is equivalent to admin user, but this is just segmenting the problem and may not allow for full flexibility in the future (e.g. if a mentor is also a judge).

Allowing multiple roles could also help simplify managing the applicants collectively given the integration of mentor and volunteer applications to the site (#386): an applicant could have both roles (e.g. applicant and hacker) so checking if somebody is an applicant in general doesn't require tracking the three possible types (or perhaps more in the future).

This change should ideally be made before applications are opened so the database records don't need to be retroactively modified.

taesungh:
Potentially controversial change to how user roles are managed. Expecting this will allow for more flexibility for new admin features and also given the need to support other types of applicants in #386.

Resolves #421. See description of changes for notable modifications besides renaming the field.

#### Changes
- With the end goal of having more flexibility in assigning user roles, make `UserRecord` store multiple possible roles
  - Field is renamed to plural and modeled as a tuple or set
- Provide empty array in default user identity instead of `null`
  - Semantically `null` could be better to indicate unknown rather than explicitly no roles, but functionally, an empty array is simpler
- Constrain database query for `check_in_participant`
  - Still need unit tests for this
- Update role and status check for `confirm_attendance_non_hacker`
- Use `AfterValidator` for `Applicant.roles` to include `Role.APPLICANT`
- Reorganize admin authorization roles to reuse dependencies
  - Basic admin access is provided to anybody with the "Organizer" role
  - Directors will need to have both "Organizer" and "Director" roles
- Update logic for checking if user has already applied
- Fix logical bug in `ParticipantAction`: function was used like boolean
- Rename role strings to be the same as their display labels

#### Testing
Note any older records in the local database will need to be updated.
1. Impersonate a new user and submit an application
2. Observe in the database that the applicant has the correct default role: array with single entry of "Applicant"
3. Add another sample user record with the "Organizer" role
4. Impersonate said user and confirm the admin dashboard is accessible
5. Add an application manager role, e.g. "Reviewer" to said user
6. Observe the Applicants page of the admin site is accessible and the previously submitted application is included

### DIFF
--- a/apps/api/src/routers/user.py
+++ b/apps/api/src/routers/user.py
@@ -35,7 +35,7 @@ DEADLINE = datetime(2025, 1, 11, 8, 1, tzinfo=timezone.utc)
 class IdentityResponse(BaseModel):
     uid: Union[str, None] = None
     status: Union[str, None] = None
-    role: Union[Role, None] = None
+    roles: list[Role] = []
 
 
 def _is_past_deadline(now: datetime) -> bool:
@@ -67,7 +67,7 @@ async def me(
     if not user:
         return IdentityResponse()
     user_record = await mongodb_handler.retrieve_one(
-        Collection.USERS, {"_id": user.uid}, ["role", "status"]
+        Collection.USERS, {"_id": user.uid}, ["roles", "status"]
     )
 
     if not user_record:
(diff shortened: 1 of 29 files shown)


waalbert (approved):
Thanks for this important user record change Taesung! This will allow us to have a lot more flexibility in giving roles.

apps/api/src/routers/user.py
```diff
@@ -89,13 +89,17 @@ async def apply(
         raise HTTPException(status.HTTP_403_FORBIDDEN, "Applications have closed.")
 
     # check if email is already in database
-    EXISTING_RECORD = await mongodb_handler.retrieve_one(
-        Collection.USERS, {"_id": user.uid}
+    existing_record = await mongodb_handler.retrieve_one(
+        Collection.USERS, {"_id": user.uid, "roles": {"exists": True}}, ["roles"]
```
taesungh, line 93: Note: I made a mistake here, the query should use `"$exists"`, not just `"exists"`

Issue #396, opened 2024-11-26
Volunteer Applications


waalbert:
For IH 2025, we're moving volunteer applications to our site, so a volunteer application page component will need to be created.
- route to this component should be `/volunteer`
- use the [questions](https://docs.google.com/document/d/1WbAwd7Mvry1UvxssD9F7qFLX0sBeVokqf7ftRGfmkVc/edit?usp=sharing) sent from the volunteers subcommittee
- have volunteer application component contain its corresponding question groups and any volunteer app specific questions in components after #405 is done

Issue #395, opened 2024-11-26
Mentor Applications


waalbert:
Since mentor applications are being added to our site for IH 2025, a mentor application page component will need to be created.
- route to mentor application page should be `/mentor`
- use [questions](https://docs.google.com/forms/d/1IN3JXCYbz_-hS84ihM8rS6gmIipxQNziGCfjayE-Sfc/edit) sent from the mentors subcommittee
- have mentor application component contain its corresponding question groups and any mentor app specific questions in components after #405 is done

PR #431, opened 2024-12-06
Generalized ApplyConfirm to be used for each application form

IanWearsHat:
Intention is to use this in every application page following the structure of `app/(main)/apply/page.tsx`
for #395 and #396 

Changes:
- Added `continueHREF` prop
  - Intention is to set this to `/mentor` and `/volunteer` for their respective forms so their confirmations continue to those hrefs
- Added `isHacker` prop to display respective confirmation text
  - Added relevant text for each role


waalbert (approved):
Thanks for updating the application confirmation messages Ian! Waiting to confirm these messages are good before merging into main

PR #433, opened 2024-12-07
fix/compress-images

noahk004:
Compressing image, reducing image sizes by 70-80%. This improves load times while making negligible difference to the background image quality.


apps/site/src/assets/images/castle_island.png
IanWearsHat, line 1: This is an amazing idea for optimization. The only things that stand out are the bands around the castle. I wonder what others think about this.

image

apps/site/src/assets/images/lantern.png
IanWearsHat, line 1: I don't believe this file will be used, so it should be safe to delete it

apps/site/src/assets/images/castle_island.png
noahk004: Good catch, just added a better quality image. Now that you mention it tho, there is a bigger difference than I had anticipated. Does this look ok or should I improve the quality in each of the images a little bit?
Screenshot 2024-12-12 011958
Screenshot 2024-12-12 011946

IanWearsHat (approved):

PR #435, opened 2024-12-07
Volunteer Applications

Bl20052005:


apps/site/src/app/(main)/apply/sections/Form/ExtraQuestions.tsx
taesungh, line 1: Chore [repeating]: these components are located under the apply page but are used only by the volunteer page, so could they be relocated to there?

apps/site/src/app/(main)/apply/sections/Form/VolunteerForm.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+import styles from "./Form.module.scss";
+import hasDeadlinePassed from "@/lib/utils/hasDeadlinePassed";
+import VolunteerFRQ from "./volunteer/VolunteerFRQ";
+import ExtraQuestions from "./ExtraQuestions";
+
+const APPLY_PATH = "/api/user/apply";
+const FIELDS_WITH_OTHER = ["pronouns", "ethnicity", "school", "major"];
+
+export default function VolunteerForm() {
```
taesungh, line 21: Looks like all of the form logic is the same/very similar to the hacker application. Could we extract that into a reusable component that accepts the form fields as children?

apps/site/src/app/(main)/volunteer/Volunteer.tsx
```diff
@@ -0,0 +1,19 @@
+import VolunteerForm from "../apply/sections/Form/VolunteerForm";
+
+const Volunteer = () => {
+	return (
+		<div className="w-full pb-10 flex flex-col items-center gap-12 mt-32">
+			<div className="font-display text-5xl font-bold">
+				Apply as a Volunteer
+			</div>
```
taesungh, line 8: Chore: could we make this a proper heading element, h1 in this case?

apps/site/src/app/(main)/volunteer/page.tsx
```diff
@@ -0,0 +1,9 @@
+import { Metadata } from "next";
+
+export const revalidate = 60;
```
taesungh, line 3: Thought: not sure if specifying a revalidation time is necessary since the page content doesn't fetch dynamic data.

taesungh (changes requested):

apps/site/src/app/(main)/volunteer/VolunteerForm.tsx
```diff
@@ -0,0 +1,20 @@
+import BasicInformation from "../apply/sections/Form/BasicInformation";
+import AgeInformation from "../apply/sections/Form/AgeInformation";
+import SchoolInformation from "../apply/sections/Form/SchoolInformation";
+import ShiftAvailability from "./components/ShiftAvailability";
+import VolunteerFRQ from "./components/VolunteerFRQ";
+import ExtraQuestions from "../apply/sections/Form/ExtraQuestions";
+import Form from "../apply/sections/Form/Form";
```
taesungh, line 7: Chore: it would make more sense for the components for any form sections used by multiple application types to be placed in a common directory elsewhere rather than being part of the Apply page.

apps/site/src/app/(main)/apply/sections/Form/Form.tsx
```diff
@@ -5,19 +5,17 @@ import axios from "axios";
...
 
 const APPLY_PATH = "/api/user/apply";
 const FIELDS_WITH_OTHER = ["pronouns", "ethnicity", "school", "major"];
 
-export default function Form() {
+interface FormProps {
+	children: React.ReactNode;
+}
+
+export default function Form({ children }: FormProps) {
```
taesungh, line 18: Suggestion: you can use React's `PropsWithChildren` to specify this without defining a new interface, e.g.
function Form({ children}: PropsWithChildren).

apps/site/src/app/(main)/apply/sections/Form/Form.tsx
```diff
@@ -93,11 +91,7 @@ export default function Form() {
 			encType="multipart/form-data"
 			onSubmit={handleSubmit}
 		>
-			<BasicInformation />
-			<SchoolInformation />
-			<ProfileInformation />
-			<ResumeInformation />
-			<AgeInformation />
+			{children}
```
taesungh, line 94: Issue: the main hacker application no longer has the form fields, so this component needs to be moved elsewhere, and a new component needs to be made for the hacker application. Would suggest using a different name (e.g. `apply/components/HackerApplicationForm.tsx`) so git can detect a rename of this file.

taesungh (changes requested):

apps/site/src/app/(main)/volunteer/VolunteerForm.tsx
Bl20052005: Refactored common forms into a dedicated forms folder

waalbert (approved):
Thanks for taking on and completing this huge task Bo and Ian!

PR #438, opened 2024-12-07
Guest Login Color Fix

Bl20052005:


IanWearsHat (approved):
Good catch! It was definitely something that I missed initially.

waalbert (approved):
Thanks for the quick fix Bo!

PR #440, opened 2024-12-07
Feat: Mentor Application

IanWearsHat:
Added mentor application at `/mentor` route

To see application, comment out lines 33-39 in `lib/components/shared/ApplicationFlow.tsx` and navigate to `/mentor?prefaceAccepted=true`

Also navigate to `/volunteer?prefaceAccepted=true` to ensure those shared components work as expected

Changes:
---

- Added questions from the IrvineHacks 2025 Mentor Application google form (can be found in the Google Drive)
- Created Slider component in `lib/components/forms/Slider.tsx`
- Created MultipleSelect component in `lib/components/forms/MultipleSelect.tsx`
  - Replaced RadioSelect with MultipleSelect
- Refactored form into BaseForm
- Refactored mentor page.tsx into ApplicationFlow.tsx
  - Volunteer form also has a preface now
- Changed all apps to use these shared components

Testing:
---
- Uncomment lines 45-53 in BaseForm.
- Fill out and submit the application. Then observe console prints correct keys and values have been added to the form.


apps/site/src/app/(main)/mentor/Form/Form.tsx
```diff
@@ -0,0 +1,167 @@
...
+		}
+
+		// Compile all technologies the mentor is comfortable with into one array
+		const experienceList: Array<string> = [];
+		for (const fieldName of Array.from(formData.keys())) {
+			if (SELECT_OPTIONS.includes(fieldName)) {
+				experienceList.push(fieldName);
+				formData.delete(fieldName);
+			}
+		}
```
taesungh, line 82: Thought: I think if we give all of the checkbox inputs the same name (e.g. `mentor_experienced_technologies` as suggested above) and use explicit values (as we currently are), the name/value pairs will automatically be parsed as an array by FastAPI according to the HTTP specification, (see [MDN Docs for Handling multiple checkboxes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#handling_multiple_checkboxes)). In theory, this means we don't need to do any special handling for the multi-select (except for the "other" option).

apps/site/src/app/(main)/mentor/Form/ProfileInformation.tsx
```diff
@@ -0,0 +1,41 @@
...
+				inputClass={styles.input}
+				containerClass="flex flex-col w-full"
+				isRequired={false}
+				type="url"
+				placeholder="https://"
+			/>
+			<TextInput
+				name="github"
+				labelClass={styles.label}
+				labelText="Github"
```
taesungh, line 22: Chore: please stylize as "GitHub", maybe also include the word profile (same for LinkedIn).
Question: will the hacker application also ask for GitHub profile and portfolio separately?

apps/site/src/app/(main)/mentor/Form/ShortAnswers.tsx
```diff
@@ -0,0 +1,36 @@
...
+export default function ShortAnswers() {
+	return (
+		<div className="flex flex-col gap-5 w-11/12">
+			<p className="text-4xl m-0 font-bold max-[700px]:text-3xl">
+				Profile Information
+			</p>
+			<Textfield
+				name="mentor_previous_experience"
+				labelClass={styles.label}
+				labelText="Have you participated or mentored at a hackathon before? If so, please list which ones. e.g. Hack at UCI 2024 (Hacker), ZotHacks 2024 (Mentor)"
```
taesungh, line 13: Chore: for 2024, our hackathon name was IrvineHacks.

apps/site/src/app/(main)/mentor/page.tsx
taesungh, line 1: Chore [non-blocking, handle later if short on time]: Looks like we have a lot of duplication here. Could we extract the Apply page and this one to a common component, e.g. `ApplicationFlow`

apps/site/src/lib/components/forms/MultipleSelect.tsx
```diff
@@ -0,0 +1,100 @@
...
+	}, [isOtherChecked]);
+
+	return (
+		<div className={containerClass}>
+			<p className="m-0 text-lg mb-4">
+				{labelText} <RequiredAsterisk />
+			</p>
+			<div className="w-10/12 flex flex-col gap-2">
+				{values.map((item, i) => {
+					const keyAndId = `${identifierId}-${i}`;
```
taesungh, line 59: Question: is the item name not sufficient as the key here?

apps/site/src/lib/components/forms/Slider.tsx
```diff
@@ -0,0 +1,63 @@
...
+				{labelText} <RequiredAsterisk />
+			</p>
+			<div className="flex gap-4 w-full">
+				<p className="hidden md:block text-center m-0">{pretext}</p>
+				<div className="relative flex items-center min-h-[50px] w-full">
+					<SliderTick leftPercentage="0.7%" label="1" />
+					<SliderTick leftPercentage="25.2%" label="2" />
+					<SliderTick leftPercentage="49.7%" label="3" />
+					<SliderTick leftPercentage="74%" label="4" />
+					<SliderTick leftPercentage="98.7%" label="5" />
```
taesungh, line 42: Question: is there a particular reason these are not aligned to every 25%?

taesungh (changes requested):
Nice start on this, some notes about reducing duplicate code, a thought on how to handle the multi-select values, and another about asking for email addresses.

apps/site/src/app/(main)/mentor/Form/Form.tsx
```diff
@@ -0,0 +1,167 @@
...
+			return;
+		}
+
+		setSubmitting(true);
+		setSessionExpired(false);
+
+		const formData = new FormData(event.currentTarget);
+
+		// Use other values when selected
+		for (const field of FIELDS_WITH_OTHER) {
```
Bl20052005, line 67: Suggestion: if we want to use the common Form without refactoring further, we can move this logic into the component in which it originates and update the input attributes there when other is selected instead of at submission time

apps/site/src/app/(main)/mentor/Form/Form.tsx
taesungh: Fun fact: looks like two years ago, multiple pronouns could be selected: HackAtUCI/HackUCI-Site#27

apps/site/src/app/(main)/mentor/Form/ProfileInformation.tsx
IanWearsHat: There is one line for Github/Portfolio in the  hacker application.

apps/site/src/lib/components/forms/Slider.tsx
IanWearsHat: When testing, the ticks did not line up with the slider thumb at exactly 25% intervals, at least on Chrome. 

I felt like it should have lined up as the slider is set to full width and the slider ticks are also in that full width div.

I also have no way to test how the thumb behaves on Webkit, so it might be different there too.

apps/site/src/lib/components/forms/MultipleSelect.tsx
IanWearsHat: I don't know why this was actually done in [RadioSelect](https://github.com/HackAtUCI/irvinehacks-site/blob/main/apps/site/src/lib/components/forms/RadioSelect.tsx), but I changed it

apps/site/src/app/(main)/mentor/Form/MentorForm.tsx
```diff
@@ -0,0 +1,51 @@
+"use client";
+
+// const formEntries = Object.fromEntries(formData.entries());
+// 		console.log(formEntries);
+
+// 		if (SELECT_FIELDS) {
+// 			for (const field of SELECT_FIELDS) {
+// 				console.log(field, formData.getAll(field));
+// 			}
+// 		}
```
taesungh, line 10: Chore [also in `BaseForm.tsx`]: please remove unused commented snippet

apps/site/src/app/(main)/mentor/Mentor.tsx
```diff
@@ -0,0 +1,23 @@
+import ApplicationFlow from "@/lib/components/forms/shared/ApplicationFlow";
+import MentorForm from "./Form/MentorForm";
+
+export const revalidate = 60;
+
+export default async function Apply({
```
taesungh, line 6: Chore: component name should match file, i.e. `Mentor`

apps/site/src/lib/components/forms/MultipleSelect.tsx
```diff
@@ -54,28 +58,27 @@ export default function RadioSelect({
...
 					horizontal ? "flex-wrap gap-10" : "flex-col gap-2"
 				}`}
 			>
 				{values.map((item, i) => {
-					const keyandId = `${IdentifierId}-${i}`;
+					const inputId = `${name}-${i}`;
 					if (item.value === "other") {
 						return (
-							<div key={keyandId} className="flex gap-2">
+							<div key={i} className="flex gap-2">
```
taesungh, line 72: Thought: using `item.value` as the key should be sufficient and will also provide a warning if we mistakenly provided different options with the same value.

apps/site/src/lib/components/forms/shared/ApplicationFlow.tsx
```diff
@@ -0,0 +1,60 @@
...
+import ApplyConfirm from "./ApplyConfirmation/ApplyConfirm";
+import Title from "./Title/Title";
+import getUserIdentity from "@/lib/utils/getUserIdentity";
+
+export const revalidate = 60;
+
+interface ApplicationFlowProps {
+	searchParams?: {
+		prefaceAccepted?: string;
+	};
```
taesungh, line 17: Thought: do you think the flow will need to know about other search params besides `prefaceAccepted`? There is some work going around in #446 on restoring URL path after logging in, so maybe keep for now and take a mental note to simplify later if not needed.

taesungh (commented):
Thanks for helping reduce the duplication among the different application pages. Some additional minor notes.

taesungh (approved):

apps/site/src/lib/components/forms/shared/ApplicationFlow.tsx
taesungh: Tracking in #480.

Issue #439, opened 2024-12-07
Explore options for avoiding redeployments of unchanged apps


taesungh:
Derived from HackAtUCI/zothacks-site#209:

Because of our quota in Vercel of 100 builds per day, it would be nice if we could avoid redeploying the different projects (Site and Sanity) both in previews and production if nothing has changed in that specific app since its last deployment (and vice versa for the site although less often there).

Vercel has a mechanism for [skipping unaffected projects in monorepos](https://vercel.com/docs/monorepos#skipping-unaffected-projects), but that requires connecting the repository which we cannot do on the Hobby team.
> This feature is only available for projects connected to GitHub repositories.

We might be able to manually replicate that by setting up [ignoring the build step](https://vercel.com/docs/monorepos#ignoring-the-build-step) based on `turbo-ignore`. However, skipping the build still counts as a deployment.
> Canceled builds initiated using the ignore build step do count towards your deployment and concurrent build limits

Therefore, we would need to somehow replicate the feature from the Actions workflows.

IanWearsHat:
Could we run `git diff --quiet HEAD^ HEAD ./apps/specific-project` at the top of the action once for site and once for sanity, check its output, and then see if we should run a deployment or not? I might be oversimplifying a lot here.

taesungh:
My initial thought was we can run `npx turbo-ignore <package>` as one step of the workflow which does something similar using `--filter="{package}...[HEAD^]`, but unfortunately, it seems that doesn't account for the task inputs: vercel/turborepo#4678. This means non-functional changes such as updating a README file or a unit test will trigger a redeployment.

We might have to implement our own hash check to replicate the desired behavior.

taesungh:
Been creating a new reusable action in HackAtUCI/super-turbo-ignore#1 that can also be used by HackAtUCI/zothacks-site in the future. Once that is ready, will update the workflows in this repository to use that action as a  job step and also reduce duplication in the workflows by [using a matrix strategy](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#using-a-matrix-strategy).

PR #411, opened 2024-12-02
Add sponsors and prep FAQs

pandyrew:
Added sponsors section

image

Also, fixes FAQs item button vertical alignment

Prepares FAQs and Empty Sponsors section for prod. The empty sponsors section is for us to show potential sponsors that they can sponsor us and how to contact us if they want to 

image

PR #443, opened 2024-12-07
Upgrade to Turborepo v2

taesungh:
As part of #439 and also fixing an issue noticed in #411.

#### Changes
- Use explicit version instead of "latest" as suggested by codemod
- Change workflows to use pnpm version specified under `packageManager`
- Replace `pipeline` key with `tasks` in `turbo.json` files
- Remove `globalDependencies` since we do not build with `.env` files
- Remove package-specific outputs from root configuration

#### Testing
Verify that preview deployment works

### DIFF
--- a/.github/workflows/run-checks.yml
+++ b/.github/workflows/run-checks.yml
@@ -91,9 +91,8 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Set up pnpm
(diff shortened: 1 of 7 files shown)


waalbert (approved):
Thanks for the upgrades Taesung!

Issue #171, opened 2023-12-30
Guest Login should redirect authenticated users to Portal


taesungh:
Similar to the Login page, if a user is already authenticated, the Guest Login page should redirect the user to `/portal`.
This is very unlikely but still possible if a user uses the browser's Back button/command.

Partly related to #170 but not exactly.

PR #446, opened 2024-12-08
Apply Login Flow

### LINKED ISSUE #445: Pre apply confirmation page redirects to login then redirects back to correct confirmation page
Currently, if the user logs in from any `ConfirmationDetails` page (hacker, mentor, or volunteer), they will be redirected to `/portal` and always to `/apply` if they haven't submitted an application yet. However, due to mentor and volunteer applications being on the site, we need to redirect the user after logging in with this specified user flow back to the correct route (`/apply`, `/mentor`, or `/volunteer`) accordingly.

waalbert:
Continuation after #449 for #445
## Changes
- Updates site and backend to use a query parameter called `return_to` to store the application route (`/apply`, `/mentor`, or `/volunteer`) the user is initially on before logging in. 
- Assuming the user hasn't submitted an app, the user should be redirected back to the route stored in `return_to` or homepage if `return_to` doesn't exist after successful login

## Testing

- Using [staging.irvinehacks.com](https://staging.irvinehacks.com), go to any of the application pages without logging in and click "Login in to apply." Log in with your UCI email and observe that you are returned back to the correct application page.
- To test guest login, repeat the steps above with a non UCI email
- Logging in without going to an application page should redirect you back to the homepage if you haven't submitted an app with that user
- Attempting to go to `/portal` without logging in redirects you to `/#apply` (Choose your character section)

Closes #445


apps/api/src/routers/dev.py
```diff
@@ -8,16 +9,21 @@
...
     user = NativeUser(
         ucinetid=ucinetid,
         display_name="Local Dev",
         email="dev@irvinehacks.com",
         affiliations=[],
     )
+    if application is not None:
+        res = RedirectResponse(f"/portal?application={application}", status_code=303)
+    else:
+        res = RedirectResponse("/portal", status_code=303)
```
taesungh, line 25: Suggestion/thought: maintaining this behavior for the dev route seems a bit excessive since this is designed for internal usage, but if it helps to have around, sure.

apps/site/src/app/(main)/apply/sections/ApplyConfirmation/ConfirmationDetails.tsx
```diff
@@ -29,7 +29,11 @@ export default async function ConfirmationDetails({
 			<Button
 				className="text-2xl"
 				text={isLoggedIn ? "Proceed to Application" : "Log in to Apply"}
-				href={isLoggedIn ? `${continueHREF}?prefaceAccepted=true` : "/login"}
+				href={
+					isLoggedIn
+						? `${continueHREF}?prefaceAccepted=true`
+						: `/login?application=${continueHREF.substring(1)}`
```
taesungh, line 35: Nitpick: might be good to rename `continueHREF` to a less complicated name, e.g. application URL

taesungh (changes requested):
Noticed this works for the dev impersonation, but I think the API also needs to be updated to pass around the continuation URL, notably the guest login and SSO login. For SSO, SAML supports a RelayState via a `return_to` address. This would help support a more generic approach where the state is not tied to a specific type of application but any URL path.

Let me look into how that works with SAML. Maybe for now you could look into modifying API /guest/login to forward a query parameter called `return_to`.

apps/api/src/routers/dev.py
waalbert: I find it useful for quickly logging into different users who have or haven't submitted an application to test to see if the user login and apply flow is correct.

apps/site/src/lib/components/forms/shared/ApplyConfirmation/ApplyConfirm.tsx
```diff
@@ -2,12 +2,12 @@ import getUserIdentity from "@/lib/utils/getUserIdentity";
 import ConfirmationDetails from "./ConfirmationDetails";
 
 interface ApplyConfirmInterface {
```
taesungh, line 4: Nitpick: interface should be named as "Props", not "Interface"

apps/site/src/lib/components/forms/shared/ApplicationFlow.tsx
```diff
@@ -16,14 +16,14 @@ interface ApplicationFlowProps {
 		prefaceAccepted?: string;
 	};
 	applicationType: "Hacker" | "Mentor" | "Volunteer";
-	continueHREF: string;
+	applicationURL: string;
```
taesungh, line 19: Suggestion: might be good to enforce the possible values as a union
```suggestion
	applicationURL: "/apply" | "/mentor" | "/volunteer";
```

apps/site/src/lib/components/forms/shared/ApplyConfirmation/ConfirmationDetails.tsx
```diff
@@ -29,7 +29,11 @@ export default async function ConfirmationDetails({
 			<Button
 				className="text-2xl"
 				text={isLoggedIn ? "Proceed to Application" : "Log in to Apply"}
-				href={isLoggedIn ? `${continueHREF}?prefaceAccepted=true` : "/login"}
+				href={
+					isLoggedIn
+						? `${applicationURL}?prefaceAccepted=true`
+						: `/login?return_to=%2F${applicationURL.substring(1)}`
```
taesungh, line 35: Nitpick: using URLSearchParams would be safer and more understandable, e.g.
```suggestion
						: `/login?${new URLSearchParams({ return_to: applicationUrl })}`
```

apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
```diff
@@ -21,12 +21,27 @@ export const enum PortalStatus {
...
+			redirect("/");
+		} else if (
+			searchParams?.return_to !== undefined &&
+			searchParams.return_to !== "/portal"
+		) {
+			redirect(searchParams.return_to);
+		} else {
+			redirect("/");
+		}
 	}
```
taesungh, line 45: Question: This condition is for when a user has no status meaning they haven't applied. In this case, I think the proper behavior is redirecting to the apply page or the home page?

apps/site/src/app/(main)/login/Login.tsx
```diff
@@ -2,16 +2,27 @@ import { redirect } from "next/navigation";
...
+	};
+}) {
 	const identity = await getUserIdentity();
-	if (identity.uid !== null) redirect("/portal");
+	if (identity.uid !== null) {
+		if (searchParams?.return_to === undefined) {
+			redirect("/portal");
+		}
+		redirect(`/portal?return_to=${searchParams.return_to}`);
+	}
```
taesungh, line 18: Question: same note as with guest login, if a user is already logged in, I don't think it's possible to navigate to `/login`, especially with a `return_to` value provided by us.

apps/api/src/routers/guest.py
```diff
@@ -72,6 +76,9 @@ async def verify_guest(
     log.info("%s authenticated as guest.", email)
     guest = guest_auth.acquire_guest_identity(email)
 
-    res = RedirectResponse("/portal", status_code=status.HTTP_303_SEE_OTHER)
+    query = urlencode({"return_to": return_to})
+    res = RedirectResponse(
+        URL(path="/portal", query=query), status_code=status.HTTP_303_SEE_OTHER
+    )
```
taesungh, line 82: Suggestion: Since this is the step when authentication completes, it would be simpler to just redirect to the `return_to` value (and verify it starts with `/`) rather than redirecting again from `Portal.tsx` (similar to how in SAML ACS, the result is `RedirectResponse(relay_state, 303)`).

taesungh (changes requested):
Nice work on this so far, some interesting things to think about with edge cases when already logged in.

apps/site/src/app/(main)/guest-login/GuestLogin.tsx
```diff
@@ -3,9 +3,22 @@ import { redirect } from "next/navigation";
...
 	const identity = await getUserIdentity();
-	if (identity.uid !== null) redirect("/portal");
+	if (identity.uid !== null) {
+		if (searchParams?.return_to === undefined) {
+			redirect("/portal");
+		} else if (searchParams.return_to === "/portal") {
+			redirect(searchParams.return_to);
+		}
+		redirect(`/portal?return_to=${searchParams.return_to}`);
+	}
```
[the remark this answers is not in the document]
taesungh: Oh actually, reading #171 again, this is possible if the user navigates back in the browser. In that case, this can be redirect to the value or portal by default
```tsx
redirect(searchParams?.return_to ?? "/portal")
```

apps/site/src/app/(main)/login/Login.tsx
taesungh: Based on the other note, if the user spams the browser's navigates backwards (even past the SSO process), it is technically possible to end up here, so same logic:
```tsx
redirect(searchParams?.return_to ?? "/portal")
```

apps/api/src/routers/guest.py
```diff
@@ -62,6 +65,7 @@ async def verify_guest(
     email: Annotated[EmailStr, Depends(guest_email)],
     passphrase: Annotated[str, Form()],
     guest_confirmation: Annotated[str, Cookie()],
+    return_to: str = "/portal",
```
taesungh, line 68: Thought: since there's already a default value provided from the initiation (the `guest_login`) function, I don't think we need to specify a default value again.

apps/api/src/routers/guest.py
```diff
@@ -72,6 +76,6 @@ async def verify_guest(
     log.info("%s authenticated as guest.", email)
     guest = guest_auth.acquire_guest_identity(email)
 
-    res = RedirectResponse("/portal", status_code=status.HTTP_303_SEE_OTHER)
+    res = RedirectResponse(return_to, status_code=status.HTTP_303_SEE_OTHER)
```
taesungh, line 79: Note: for security purposes, we'll want to confirm this destination is on our own server. The simple check is making sure it starts with `/`.

apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
```diff
@@ -21,12 +21,18 @@ export const enum PortalStatus {
...
 }
 
-async function Portal() {
+async function Portal({
+	searchParams,
+}: {
+	searchParams?: {
+		return_to?: string;
+	};
+}) {
```
taesungh, line 30: Thought: I don't think there's any situation where a user will be sent to `/portal` with a `return_to` parameter?

apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
```diff
@@ -21,12 +21,18 @@ export const enum PortalStatus {
...
 	const identity = await getUserIdentity();
 	const status = identity.status;
 
 	if (status === null) {
-		redirect("/apply");
+		redirect(searchParams?.return_to ?? "/");
```
taesungh, line 35: Note for later: we should consider having the Choose your Character section anchored with an ID so we can navigate to something like `/#apply` in situations like these.

apps/site/src/app/(main)/(home)/sections/ChooseCharacter/ChooseCharacter.tsx
```diff
@@ -14,7 +14,7 @@ import styles from "./CharacterBox.module.scss";
 
 const ChooseCharacter = () => {
 	return (
-		<section className="flex flex-col justify-center my-14">
+		<section className="flex flex-col justify-center my-14" id="apply">
```
taesungh, line 17: Note for later: didn't think you would add it now, but I think this can be used now instead of what #437 did.

apps/api/tests/test_guest.py
```diff
@@ -123,15 +123,38 @@ def test_invalid_guest_verification_is_unauthorized(
...
+    mock_issue_user_identity: AsyncMock,
+    mock_remove_guest_key: AsyncMock,
+) -> None:
+    """Test a guest successfully verifying guest credentials."""
+    mock_get_existing_key.return_value = guest_auth._generate_key(
+        "some-confirmation", SAMPLE_PASSPHRASE
+    )
+
+    res = client.post(
+        "/verify?return_to=MIT",
```
taesungh, line 153: Nice easter egg :)

taesungh (approved):
Nicely done

PR #449, opened 2024-12-08
Implement RelayState for SAML authentication

taesungh:
As part of #445, related to #446:

Implement SAML RelayState for our UCI SSO implementation. See [python3-saml README](https://github.com/SAML-Toolkits/python3-saml/blob/master/README.md) for further explanation about the SAML response to the ACS. In practice, the RelayState could be any URL (even on a different host), but to simplify things, we will enforce that only a pathname is provided, e.g. (`/portal`) without a host or protocol.

Once this is merged, the main login endpoint (`/user/login`) can forward a `return_to` query parameter to the SAML login endpoint, and a similar thing will need to be added for the guest login endpoint.

#### Changes
- Login endpoint takes in `return_to` parameter which is provided back together with the SAML Response as `RelayState`
- Simplify `_get_saml_auth` to remove unused request data
- Use explicit form fields for `SAMLResponse` and `RelayState` in ACS
- Improve unit tests

#### Testing
1. Will try to manually assign staging domain name to preview deployment (if Vercel cooperates)
2. Once ready, people can try navigating to `staging.irvinehacks.com/api/saml/login?return_to=%2Fmy-path`
3. After the UCI SSO, you should be redirected to `/my-path`

### DIFF
--- a/apps/api/src/routers/saml.py
+++ b/apps/api/src/routers/saml.py
@@ -3,14 +3,15 @@ import os
 from functools import lru_cache
 from logging import getLogger
 from pathlib import Path
-from typing import Any
+from typing import Annotated, Any, Mapping
 
-from fastapi import APIRouter, HTTPException, Request
+from fastapi import APIRouter, Form, HTTPException, Request, status
+from fastapi.datastructures import URL
 from fastapi.responses import RedirectResponse, Response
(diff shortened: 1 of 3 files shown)


waalbert (approved):
Thanks Taesung for adding RelayState for UCI SSO! This will greatly benefit the user flow

PR #451, opened 2024-12-09
BaseForm filters out other field correctly

### LINKED ISSUE #450: Allow "other" values to be sent with more values on FormData
Right now on `MultipleSelect` checkbox form entries, if the user checks some checkbox values and selects "other" as well, our Forms don't properly replace `other_<field>: <other_value>` with `<field>: <other_value>`. Examples include `pronouns` and `experienced_technologies`

What needs to be done?
- Add `experienced_technologies` to `FIELDS_WITH_OTHER` in `BaseForm.tsx`
- Fix lines `37-43` in `BaseForm.tsx` to correctly find `<field>: "other"` even when there's multiple `<field>` entries and replace "other" with the value from `other_<field>`

IanWearsHat:
Closes #450 

- Added `experienced_technologies` to `FIELDS_WITH_OTHER`
- Changed logic in `handle_submit` to remove other field and add back its value under the correct field name

### DIFF
--- a/apps/site/src/app/(main)/mentor/Form/MentorForm.tsx
+++ b/apps/site/src/app/(main)/mentor/Form/MentorForm.tsx
@@ -32,7 +32,7 @@ export default function MentorForm() {
 			/>
 			<ProfileInformation />
 			<Textfield
-				name="_other_questions"
+				name="other_questions"
 				labelText="Questions/comments/concerns?"
 				containerClass="flex flex-col w-11/12"
 				isRequired={false}
--- a/apps/site/src/app/(main)/mentor/Form/MentorForm.tsx
+++ b/apps/site/src/app/(main)/mentor/Form/MentorForm.tsx
@@ -32,7 +32,7 @@ export default function MentorForm() {
 			/>
 			<ProfileInformation />
 			<Textfield
-				name="other_questions"
+				name="_other_questions"
 				labelText="Questions/comments/concerns?"
 				containerClass="flex flex-col w-11/12"
 				isRequired={false}
--- a/apps/site/src/lib/components/forms/shared/BaseForm.tsx
+++ b/apps/site/src/lib/components/forms/shared/BaseForm.tsx
@@ -46,10 +46,8 @@ export default function BaseForm({ children }: PropsWithChildren) {
 
 			formData.delete(otherField);
 
-			const entriesWithoutOther = Array.from(
-				formData
-					.entries()
-					.filter(([key, value]) => key === field && value !== "other"),
+			const entriesWithoutOther = Array.from(formData.entries()).filter(
+				([key, value]) => key === field && value !== "other",
 			);
 
 			formData.delete(field);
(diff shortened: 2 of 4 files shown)


apps/site/src/lib/components/forms/shared/BaseForm.tsx
```diff
@@ -36,12 +42,25 @@ export default function BaseForm({ children }: PropsWithChildren) {
 		// Use other values when selected
 		for (const field of FIELDS_WITH_OTHER) {
 			const otherField = `other_${field}`;
-			if (formData.get(field) === "other") {
-				formData.set(field, formData.get(otherField)!);
-				formData.delete(otherField);
-			}
+			const otherFieldValue = formData.get(otherField);
+
+			formData.delete(otherField);
```
taesungh, line 47: Issue: one of the mentor form fields is currently named `other_questions` and becomes removed here, so the field would need to be renamed.

apps/site/src/lib/components/forms/shared/BaseForm.tsx
```diff
@@ -36,12 +42,25 @@ export default function BaseForm({ children }: PropsWithChildren) {
...
+
+			const entriesWithoutOther = Array.from(formData.entries()).filter(
+				([key, value]) => key === field && value !== "other",
```
taesungh, line 50: Thought: I think you can simplify the filter by using `formData.getAll(field)`

apps/site/src/lib/components/forms/shared/BaseForm.tsx
IanWearsHat: I don't see the field being deleted on my end when I print all the form data out. The loop only takes fields that are in the `FIELDS_WITH_OTHER` array and "questions" is not one of them.

image

taesungh: Oh my bad, seems I somehow couldn't find it in the output. Perhaps still a good idea to rename it to reduce chance of collision, or better yet, use a different prefix for the actual other fields, e.g. `_other_pronouns`.

taesungh (dismissed):

apps/site/src/lib/components/forms/shared/BaseForm.tsx
```diff
@@ -36,12 +42,24 @@ export default function BaseForm({ children }: PropsWithChildren) {
...
+
+			formData.delete(field);
+
+			for (const value of valuesWithoutOther) formData.append(field, value);
+
+			if (otherFieldValue) formData.append(field, otherFieldValue);
 		}
 
+		const entries = Array.from(formData.entries());
+		for (const e of entries) console.log(e);
```
Bl20052005, line 61: should probably delete this before merging

Bl20052005 (changes requested):

Bl20052005 (dismissed):

apps/site/src/app/(main)/mentor/Form/MentorForm.tsx
```diff
@@ -32,7 +32,7 @@ export default function MentorForm() {
 			/>
 			<ProfileInformation />
 			<Textfield
-				name="other_questions"
+				name="_other_questions"
```
taesungh, line 35: Could you flip the usage of the private notation (update `OtherPopup` and `MultipleSelect`) since `other_questions` is being used in #453?

taesungh (approved):

Bl20052005 (approved):

PR #453, opened 2024-12-09
Added generalized flow for applications

### LINKED ISSUE #452: Restructure API routes for other application types
image
Workaround is to make other routes with specified RawApplicationData types.

IanWearsHat:
Created generalized flow for applications.

Intention is to use `apply_flow` function with each application route.

### DIFF
--- a/apps/api/src/routers/user.py
+++ b/apps/api/src/routers/user.py
@@ -9,7 +9,10 @@ from pydantic import BaseModel, EmailStr
 from auth import user_identity
 from auth.authorization import require_accepted_applicant
 from auth.user_identity import User, require_user_identity, use_user_identity
-from models.ApplicationData import ProcessedApplicationData, RawApplicationData
+from models.ApplicationData import (
+    ProcessedApplicationData,
+    RawApplicationData,
+)
 from models.user_record import Applicant, BareApplicant, Role, Status
 from services import docusign_handler, mongodb_handler
 from services.docusign_handler import WebhookPayload
@@ -67,14 +70,11 @@ async def me(
     return IdentityResponse(uid=user.uid, **user_record)
 
 
-@router.post("/apply", status_code=status.HTTP_201_CREATED)
-async def apply(
-    user: Annotated[User, Depends(require_user_identity)],
-    # media type should be automatically detected but seems like a bug as of now
-    raw_application_data: Annotated[
-        RawApplicationData, Form(media_type="multipart/form-data")
-    ],
-) -> str:
+async def apply_flow(
+    user: User,
+    raw_application_data: Union[RawApplicationData],
+    processed_application_type: Union[RawApplicationData],
+):
     if raw_application_data.application_type not in Role.__members__:
         raise HTTPException(
             status.HTTP_422_UNPROCESSABLE_ENTITY,
@@ -131,7 +131,7 @@ async def apply(
     else:
         resume_url = None
 
-    processed_application_data = ProcessedApplicationData(
+    processed_application_data = processed_application_type(
         **raw_app_data_dump,
         resume_url=resume_url,
         submission_time=now,
@@ -171,6 +171,20 @@ async def apply(
     return (
         "Thank you for submitting an application to IrvineHacks 2024! Please "
         + "visit https://irvinehacks.com/portal to see your application status."
+    )
+
+
+@router.post("/apply", status_code=status.HTTP_201_CREATED)
+async def apply(
+    user: Annotated[User, Depends(require_user_identity)],
+    # media type should be automatically detected but seems like a bug as of now
+    raw_application_data: Annotated[
+        RawApplicationData,
+        Form(media_type="multipart/form-data"),
+    ],
+) -> str:
+    return await apply_flow(
+        user, raw_application_data, processed_application_type=ProcessedApplicationData
     )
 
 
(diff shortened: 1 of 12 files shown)


Bl20052005 (dismissed):

apps/api/src/routers/user.py
```diff
@@ -67,13 +70,10 @@ async def me(
...
-@router.post("/apply", status_code=status.HTTP_201_CREATED)
-async def apply(
-    user: Annotated[User, Depends(require_user_identity)],
-    # media type should be automatically detected but seems like a bug as of now
-    raw_application_data: Annotated[
-        RawApplicationData, Form(media_type="multipart/form-data")
-    ],
+async def apply_flow(
+    user: User,
+    raw_application_data: Union[RawApplicationData],
```
taesungh, line 75: Question: Does the `raw_application_data` parameter need to be typed as a union? Is the plan to expand this in the future with the other application types, and if so, would that actually be necessary?

waalbert: I'm testing it taking other application types which I'm guessing is what it's supposed to do (correct me if I'm wrong Ian), and I'm running into the same issue as #452.

IanWearsHat: Yes, the intention is to expand the union to include other application types, but using a discriminated union as you explained should be much better.

taesungh: Separate routes should be fine if they each have a concrete form model. Since this parameter isn't really used below before being dumped into `ProcessedApplicationData` where validation will happen a second time (assuming each route does a first time), I think this could just be a `dict` so you don't have to worry about the union.

taesungh: New thought: if `ProcessedApplicationData` is a [Type Adapter](https://docs.pydantic.dev/latest/concepts/type_adapter/) for a discriminated union, then it's okay to have `raw_application_data` typed as `Union[HackerApplicationData, MentorApplicationData, VolunteerApplicationData]` which allows any common pre-processing for special fields (e.g. resume), and then you can remove the second parameter for which type of application it is.

apps/api/src/routers/user.py
```diff
@@ -67,14 +72,11 @@ async def me(
...
-    user: Annotated[User, Depends(require_user_identity)],
-    # media type should be automatically detected but seems like a bug as of now
-    raw_application_data: Annotated[
-        RawApplicationData, Form(media_type="multipart/form-data")
-    ],
+async def apply_flow(
+    user: User,
+    raw_application_data: Union[RawHackerApplicationData, RawMentorApplicationData],
 ) -> str:
+    print("apply_flow called")
```
taesungh, line 79: Chore: Please remove stray print statement

apps/api/src/routers/user.py
```diff
@@ -103,7 +105,7 @@ async def apply(
     raw_app_data_dump = raw_application_data.model_dump()
 
     for field in ["pronouns", "ethnicity", "school", "major"]:
-        if raw_app_data_dump[field] == "other":
+        if field in raw_app_data_dump and raw_app_data_dump[field] == "other":
```
taesungh, line 108: Suggestion: can simplify checking if field exists by using `dict.get`

apps/site/src/lib/components/forms/shared/BaseForm.tsx
```diff
@@ -10,7 +10,12 @@ import hasDeadlinePassed from "@/lib/utils/hasDeadlinePassed";
 const APPLY_PATH = "/api/user/apply";
 const FIELDS_WITH_OTHER = ["pronouns", "ethnicity", "school", "major"];
 
-export default function BaseForm({ children }: PropsWithChildren) {
+export default function BaseForm({
+	applicationType,
+	children,
+}: {
+	applicationType: string;
+} & PropsWithChildren) {
```
taesungh, line 18: Suggestion: `PropsWithChildren` is a generic type, so another type/interface can be specified, e.g. `PropsWithChildren<MyProps>`

apps/site/src/lib/components/forms/shared/BaseForm.tsx
```diff
@@ -42,6 +47,9 @@ export default function BaseForm({ children }: PropsWithChildren) {
 			}
 		}
 
+		// attach application type to formData
+		formData.append("application_type", applicationType);
```
taesungh, line 51: Issue [marginal]: for any users with JavaScript disabled (should be very rare), the additional processing won't run, so this field will be missed. A more versatile solution is having a hidden input field in the form, see `GuestLoginVerificationForm` for an example.

taesungh (approved):
Minor notes, can address later if in a rush.

apps/site/src/app/(main)/apply/Form/ProfileInformation.tsx
```diff
@@ -1,5 +1,8 @@
+import Textfield from "@/lib/components/forms/Textfield";
 import TextInput from "@/lib/components/forms/TextInput";
 
+const FRQ_MAX_LENGTH = 2000;
```
Bl20052005, line 4: Just realized that max length is technically 2048 in the api, but 2000 here

Bl20052005 (approved):

apps/site/src/app/(main)/apply/Form/ProfileInformation.tsx
taesungh: Different browsers count newline characters slightly differently, so the limit in the frontend was originally made to be slightly lower than in the API (see description in #152). Not sure if this is still the case, but might be good to explain this as a comment in the code.

PR #457, opened 2024-12-09
Added mentor app tests, addressed follow-up notes

### LINKED ISSUE #455: Follow-ups to new flow with apply routes in API
Leftover notes from #444 and #453:
- Use `Literal["Hacker", "Mentor", "Volunteer"]` type for `application_type`
  - This removes the need for manual validation in the application flow
  - In the specific application data models, the type should be just a single literal value
  - [When constructing the `Applicant` object](https://github.com/HackAtUCI/irvinehacks-site/blob/60e9e11fab4f98b29256b4d5543fc7872d20b91e/apps/api/src/routers/user.py#L148-L155), the value can be used directly rather than indexing the `Role` enum
  - The value in the input field in the application form will need to be updated to Titlecase
- Simplify checking other fields using `dict.get`
- New note: from `ApplicationData.py` declare ProcessedApplicationData as a [discriminated union](https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions) so the consumers don't have to write out the union each time
- New note: Remove the default value for `roles` in `UserRecord.Applicant`

### LINKED ISSUE #456: Follow-ups to application form modifications
Leftover notes from #453:
- Use hidden input field for application type
    > For any users with JavaScript disabled (should be very rare), the additional processing won't run, so this field will be missed. A more versatile solution is having a hidden input field in the form, [see `GuestLoginVerificationForm` for an example](https://github.com/HackAtUCI/irvinehacks-site/blob/60e9e11fab4f98b29256b4d5543fc7872d20b91e/apps/site/src/app/(main)/guest-login/components/GuestLoginVerificationForm.tsx#L26).

- Include comment explaining why FRQ length limit is lower on the site than in the API
  - This may not be an issue in modern versions of the browsers, so might need to verify again
- Simplify usage of `PropsWithChildren`

IanWearsHat:
Added tests for mentor apps.

Also addresses notes to close #455 and close #456

---

From #455:
- In `ApplicationData.py` declared `ProcessedApplicationData` as a discriminated union
  - Removed manual type checking in `_apply_flow`
- Used Literal type for `application_type` in `user_record.Applicant`
- Removed manual check for `application_type` in `_apply_flow`
- Simplified checking other fields, but added checking for inclusion in list (for multiple pronouns and experienced_technolgoies)
- Removed default value for roles in `UserRecord.Applicant`

From #456 
- Used hidden input field for application type
- Included comment explaining why FRQ length limit is lower on the site than in the API
- Simplify usage of `PropsWithChildren` in `BaseForm`

### DIFF
--- a/apps/api/src/models/ApplicationData.py
+++ b/apps/api/src/models/ApplicationData.py
@@ -58,15 +58,15 @@ class BaseMentorApplicationData(BaseModel):
     school: str
     major: str
     education_level: str
-    is_18_older: str
+    is_18_older: bool
     git_experience: str
     github: NullableHttpUrl = None
     portfolio: NullableHttpUrl = None
     linkedin: NullableHttpUrl = None
     mentor_prev_experience_saq1: Union[str, None] = Field(None, max_length=2048)
-    mentor_interest_saq2: Union[str, None] = Field(None, max_length=2048)
-    mentor_team_help_saq3: Union[str, None] = Field(None, max_length=2048)
-    mentor_team_help_saq4: Union[str, None] = Field(None, max_length=2048)
+    mentor_interest_saq2: str = Field(max_length=2048)
+    mentor_team_help_saq3: str = Field(max_length=2048)
+    mentor_team_help_saq4: str = Field(max_length=2048)
     other_questions: Union[str, None] = Field(None, max_length=2048)
 
 
@@ -105,7 +105,7 @@ class ProcessedMentorApplicationData(BaseMentorApplicationData):
     submission_time: datetime
     reviews: list[Review] = []
 
-    @field_serializer("linkedin", "portfolio", "resume_url")
+    @field_serializer("linkedin", "github", "portfolio", "resume_url")
     def url2str(self, val: Union[HttpUrl, None]) -> Union[str, None]:
         if val is not None:
             return str(val)
--- a/apps/api/src/models/ApplicationData.py
+++ b/apps/api/src/models/ApplicationData.py
@@ -47,15 +47,15 @@ class BaseApplicationData(BaseModel):
     is_first_hackathon: bool
     linkedin: NullableHttpUrl = None
     portfolio: NullableHttpUrl = None
-    frq_change: Union[str, None] = Field(None, max_length=2048)
+    frq_change: str = Field(max_length=2048)
     frq_video_game: str = Field(max_length=2048)
 
 
 class BaseMentorApplicationData(BaseModel):
     model_config = ConfigDict(str_strip_whitespace=True, str_max_length=254)
 
-    experienced_technologies: Optional[List[str]] = []
-    pronouns: Optional[List[str]] = []
+    experienced_technologies: List[str] = []
+    pronouns: List[str] = []
 
     school: str
     major: str
@@ -78,7 +78,7 @@ class RawHackerApplicationData(BaseApplicationData):
     first_name: str
     last_name: str
     resume: Union[UploadFile, None] = None
-    application_type: Literal["HACKER"]
+    application_type: Literal["Hacker"]
 
 
 class RawMentorApplicationData(BaseMentorApplicationData):
@@ -86,8 +86,8 @@ class RawMentorApplicationData(BaseMentorApplicationData):
 
     first_name: str
     last_name: str
-    resume: Union[UploadFile, None] = None
-    application_type: Literal["MENTOR"]
+    resume: UploadFile
+    application_type: Literal["Mentor"]
 
 
 class ProcessedHackerApplicationData(BaseApplicationData):
--- a/apps/api/src/models/ApplicationData.py
+++ b/apps/api/src/models/ApplicationData.py
@@ -1,14 +1,16 @@
 from datetime import datetime
 from enum import Enum
-from typing import Annotated, List, Optional, Union
+from typing import Annotated, Any, List, Literal, Optional, Union
 
 from fastapi import UploadFile
 from pydantic import (
     BaseModel,
     BeforeValidator,
     ConfigDict,
+    Discriminator,
     Field,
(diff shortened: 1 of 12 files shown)


apps/api/src/models/ApplicationData.py
```diff
@@ -52,21 +54,21 @@ class BaseApplicationData(BaseModel):
 class BaseMentorApplicationData(BaseModel):
     model_config = ConfigDict(str_strip_whitespace=True, str_max_length=254)
 
-    experienced_technologies: str
-    pronouns: str
+    experienced_technologies: Optional[List[str]] = []
+    pronouns: Optional[List[str]] = []
```
taesungh, line 58: Suggestion: If we are specifying a default value of an empty list, these fields do not need to be possibly `None`

apps/api/src/models/ApplicationData.py
```diff
@@ -85,7 +87,7 @@ class RawMentorApplicationData(BaseMentorApplicationData):
     first_name: str
     last_name: str
     resume: Union[UploadFile, None] = None
```
taesungh, line 89: Question: for mentors, I see the resume is currently specified as required. If that is the case, then I think this field should not be possibly `None`.

apps/api/src/models/user_record.py
```diff
@@ -75,7 +74,5 @@ class Applicant(BareApplicant):
     """Applicant with application data."""
 
     # Note validators not run on default values
```
taesungh, line 76: Chore: comment is no longer relevant and should be removed

apps/api/src/routers/user.py
```diff
@@ -105,7 +97,8 @@ async def apply_flow(
     raw_app_data_dump = raw_application_data.model_dump()
 
     for field in ["pronouns", "ethnicity", "school", "major"]:
-        if field in raw_app_data_dump and raw_app_data_dump[field] == "other":
+        value = raw_app_data_dump.get(field)
+        if any([value == "other", isinstance(value, list) and "other" in value]):
```
taesungh, line 101: Suggestion: the usage of any when this is not a dynamic iterable seems confusing. Consider replacing with a disjunction: `if cond1 or cond2:`.

apps/site/src/lib/components/forms/shared/BaseForm.tsx
```diff
@@ -16,12 +15,16 @@ const FIELDS_WITH_OTHER = [
 	"experienced_technologies",
 ];
 
+interface BaseFormProps {
+	applicationType: "HACKER" | "MENTOR" | "VOLUNTEER";
```
taesungh, line 19: Suggestion: if you make these Titlecase, they will match the expected `user_record.Role` label values, so the value can be used like a Role at `user.py:144` rather than indexing the enum.

taesungh (commented):
Thanks for helping clean things up. Some additional minor notes.

apps/site/src/lib/components/forms/shared/BaseForm.tsx
```diff
@@ -102,10 +102,17 @@ export default function BaseForm({
...
 			method="post"
 			className="bg-black border-[5px] border-white text-[var(--color-white)] w-8/12 flex flex-col items-center py-12 gap-14 z-1 max-[800px]:w-9/12 max-[400px]:w-11/12 drop-shadow-[25px_33px_0px_rgba(255,255,255,1)]"
-			action="/api/user/apply"
+			action={applyPath}
 			encType="multipart/form-data"
 			onSubmit={handleSubmit}
 		>
+			<input
+				type="text"
+				name="application_type"
```
taesungh, line 111: Chore: now that we have the hidden input field, we no longer need to manually add the value in `handleSubmit`.

apps/api/src/routers/user.py
```diff
@@ -141,7 +141,7 @@ async def _apply_flow(
         uid=user.uid,
         first_name=raw_application_data.first_name,
         last_name=raw_application_data.last_name,
-        roles=(Role.APPLICANT, Role[raw_application_data.application_type]),
+        roles=(Role.APPLICANT, raw_application_data.application_type),  # type: ignore
```
taesungh, line 144: I think you can construct the enum object instead as `Role(application_type)` to satisfy the typing, although that assumes the key is valid (which it should be).

apps/site/src/app/(main)/apply/Form/HackerBasicInformation.tsx
```diff
@@ -21,8 +21,8 @@ const ethnicity = [
 ];
 
 const yesNoOptions = [
-	{ value: "hack-yes", text: "Yes" },
-	{ value: "hack-no", text: "No" },
+	{ value: "Yes", text: "Yes" },
+	{ value: "No", text: "No" },
```
taesungh, line 25: Thank you for fixing this, also noticed while testing other stuff.

apps/api/src/models/ApplicationData.py
```diff
@@ -47,15 +47,15 @@ class BaseApplicationData(BaseModel):
...
     frq_video_game: str = Field(max_length=2048)
 
 
 class BaseMentorApplicationData(BaseModel):
     model_config = ConfigDict(str_strip_whitespace=True, str_max_length=254)
 
-    experienced_technologies: Optional[List[str]] = []
-    pronouns: Optional[List[str]] = []
+    experienced_technologies: List[str] = []
+    pronouns: List[str] = []
```
taesungh, line 58: Chore: since we are on Python 3.9+, we can directly use generics of built-in containers, e.g. `list[str]` instead of using `typing.List`.

taesungh (dismissed):

waalbert (approved):
Thanks for all the work allowing mentor apps to be submitted and fixing hacker apps Ian!

Issue #420, opened 2024-12-05
Use new FastAPI CLI for development


taesungh:
Following from #384, [FastAPI 0.111.0](https://fastapi.tiangolo.com/release-notes/#01110) includes a new [FastAPI CLI](https://fastapi.tiangolo.com/fastapi-cli/) to make running the development server easier. Sample usage has already been integrated into the [HackAtUCI/starter-pack-full-stack-web-app](https://github.com/HackAtUCI/starter-pack-full-stack-web-app), and we can use it for this project as well.
- Update the requirements files to include [fastapi-cli](https://pypi.org/project/fastapi-cli/) as a dev dependency
  - Note: do not use `fastapi[standard]` since the additional dependencies do not need to be bundled in the Vercel Serverless Function deployment (see explanation in HackAtUCI/starter-pack-full-stack-web-app#19)
- Update the API README file and Dockerfile with the new CLI commands
- Remove the [`dev.py` script](https://github.com/HackAtUCI/irvinehacks-site/blob/76c767c1e549527dc8620edc27417dbbb0985267/apps/api/src/dev.py)

PR #465, opened 2024-12-10
Mock external services during local development

### LINKED ISSUE #447: Explore ways to patch external services in local development
For the API to function properly, it needs to connect to external services including SendGrid, the Google Drive API, MongoDB, and UCI SSO. The ability to use a local database via a Docker container was added in #72, and the ability to simulate a user identity was added in #193, but it would help to be able to patch the other services so parts of the `/user/apply` endpoint can be tested in full: notably, the act of sending emails and uploading files to Google Drive. This would also help with testing the guest login system (relevant for #446) which involves sending the passphrase by email.

The interactions with these external services could be simulated with a log message describing such. Ideally, the actual service handlers (e.g. `sendgrid_handler.py` and `gdrive_handler.py`) should not be modified so that their intended functionality is clear. One alternative is constructing layers of dependency injection, but this can also get messy with going through FastAPI and just complicates things for actual production usage. Thankfully, Python provides the ability to dynamically patch things on the fly with `unittest.mock` (not just for unit testing!) which will allow us to override things from the top level without modifying the internals of the service handlers.

Ideally, this change should happen in `app.py` since `dev.py` is planned to be removed in #420. We should be able to attach [Lifespan Events](https://fastapi.tiangolo.com/advanced/events/) to the FastAPI app where we can start and stop a patch session through a context manager.

taesungh:
Resolves #447: Adding mocks for external services to simplify local development by patching service handlers during the [app lifespan](https://fastapi.tiangolo.com/advanced/events/#lifespan).

Later want to replace checking the `DEPLOYMENT` environment variable with checking for entering from the FastAPI CLI while working on #420.

#### Changes
- When testing the API server, patch usage of external services
  - Mock uploading file to Google Drive and sending emails with SendGrid
  - Remove `Secure` attribute from cookies on `RedirectResponse`
    - Also simplifies issuing user identity in `dev.impersonate`
- Add additional log messages for guest authentication flow

#### Testing
1. Launch the API dev server with the MongoDB instance: `npm run dev` from `apps/api`
   - Make sure you have a document with the `_id` field equal `word_list` and a field called `word` of an array of strings inside the collection `irvinehacks-prod.settings`
2. Complete the guest authentication process: the passphrase should be shown as a mocked email in the logs
3. Submit an application with a resume file
4. Observe the file upload is mocked, and a fake URL is generated
5. Observe the confirmation email is also mocked
6. Verify that impersonating a dev user still works

### DIFF
--- a/apps/api/src/utils/dev_mock.py
+++ b/apps/api/src/utils/dev_mock.py
@@ -1,0 +1,71 @@
+import base64
+import hashlib
+import random
+from contextlib import asynccontextmanager
+from logging import getLogger
+from typing import Any, AsyncIterator, Iterable, Union
+from unittest.mock import patch
+
+from fastapi import FastAPI, Response
+from fastapi.responses import RedirectResponse
+
+from services.sendgrid_handler import Template
+
+log = getLogger(__name__)
+
+SAMPLE_RESUMES_FOLDER_ID = "1a_Hacker_Resumes"
+
+
+def mock_set_cookie(self: RedirectResponse, *args: Any, **kwargs: Any) -> None:
+    """Allow cookies on non-https for local development."""
+    kwargs["secure"] = False
+    return Response.set_cookie(self, *args, **kwargs)
+
+
+async def mock_send_email(
+    template_id: Template,
+    sender_email: tuple[str, str],
+    receiver_data: Union[dict[str, str], Iterable[dict[str, str]]],
+    send_to_multiple: bool = False,
+) -> None:
+    """Mock sending of email through SendGrid."""
+    if send_to_multiple:
+        log.info(
+            "Simulating sending %s email to %s", template_id.name, list(receiver_data)
+        )
+    else:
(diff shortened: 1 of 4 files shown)


apps/api/src/utils/dev_mock.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@asynccontextmanager
+async def lifespan(app: FastAPI) -> AsyncIterator[None]:
+    """Patch usage of external services for local development."""
+
+    # Important: these must be imported as `from services import handler`
+    # rather than `from services.handler import function` when used to be patched
+    patch(
+        "utils.resume_handler.RESUMES_FOLDER_ID", new=SAMPLE_RESUMES_FOLDER_ID
```
taesungh, line 63: Note: this will need to be updated because of #468.

waalbert: I will update when I get the chance or you can right now if you want.

waalbert (approved):
Thanks Taesung! This will greatly help our local development

PR #468, opened 2024-12-10
Separate hacker and mentor resumes

### LINKED ISSUE #467: Separate hacker and mentor resumes
Right now, hacker and mentor resumes are being sent to the same GDrive folder. Separating hacker and mentor resume submissions into different folders would make life easier for the other committees if they need access to the resumes.

waalbert:
Closes #467

### DIFF
--- a/apps/api/src/utils/resume_handler.py
+++ b/apps/api/src/utils/resume_handler.py
@@ -1,7 +1,7 @@
 import hashlib
 import os
 from logging import getLogger
-from typing import Protocol, Literal
+from typing import Protocol, Literal, Union
 
 from aiogoogle import HTTPError
 from fastapi import UploadFile
@@ -20,7 +20,7 @@ ACCEPTED_TYPES = ("application/pdf",)
 class Person(Protocol):
     first_name: str
     last_name: str
-    application_type: Literal["Hacker"] | Literal["Mentor"]
+    application_type: Union[Literal["Hacker"], Literal["Mentor"]]
 
 
 async def upload_resume(person: Person, resume_upload: UploadFile) -> HttpUrl:
--- a/apps/api/src/utils/resume_handler.py
+++ b/apps/api/src/utils/resume_handler.py
@@ -1,7 +1,7 @@
 import hashlib
 import os
 from logging import getLogger
-from typing import Protocol
+from typing import Protocol, Literal
 
 from aiogoogle import HTTPError
 from fastapi import UploadFile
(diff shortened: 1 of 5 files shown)


apps/api/src/utils/resume_handler.py
```diff
@@ -11,20 +11,29 @@
...
+HACKER_RESUMES_FOLDER_ID = os.getenv("HACKER_RESUMES_FOLDER_ID")
+MENTOR_RESUMES_FOLDER_ID = os.getenv("MENTOR_RESUMES_FOLDER_ID")
 SIZE_LIMIT = 500_000
 ACCEPTED_TYPES = ("application/pdf",)
 
 
 class Person(Protocol):
     first_name: str
     last_name: str
+    application_type: Union[Literal["Hacker"], Literal["Mentor"]]
```
taesungh, line 23: Note: Literals can accept multiple values, e.g. `Literal["Hacker", "Mentor"]`

apps/api/src/utils/resume_handler.py
```diff
@@ -11,20 +11,29 @@
...
 async def upload_resume(person: Person, resume_upload: UploadFile) -> HttpUrl:
     """Upload resume file to Google Drive and provide url to uploaded file.
     Reject files larger than size limit"""
-    if not RESUMES_FOLDER_ID:
+
+    RESUME_FOLDER_ID = None
+    if person.application_type == "Hacker":
+        RESUME_FOLDER_ID = HACKER_RESUMES_FOLDER_ID
+    elif person.application_type == "Mentor":
+        RESUME_FOLDER_ID = MENTOR_RESUMES_FOLDER_ID
```
taesungh, line 34: Suggestion: to ensure we don't miss a literal value if the input type is updated in the future, it would help to do [exhaustiveness checking](https://mypy.readthedocs.io/en/stable/literal_types.html#exhaustiveness-checking) with `assert_never`, e.g.
```python
else:
    assert_never(person.application_type)
```

apps/api/src/utils/resume_handler.py
```diff
@@ -11,20 +11,29 @@
...
+
+    if not RESUME_FOLDER_ID:
         raise RuntimeError("RESUMES_FOLDER_ID is not defined")
```
taesungh, line 37: Note for later: we probably want to move a check like this to `apps/api/index.py` and add an integration test so we ensure these environment variables are always set on a deployment.

apps/api/src/utils/resume_handler.py
```diff
@@ -11,20 +11,29 @@
 
 log = getLogger(__name__)
 
-RESUMES_FOLDER_ID = os.getenv("RESUMES_FOLDER_ID")
+HACKER_RESUMES_FOLDER_ID = os.getenv("HACKER_RESUMES_FOLDER_ID")
+MENTOR_RESUMES_FOLDER_ID = os.getenv("MENTOR_RESUMES_FOLDER_ID")
```
taesungh, line 15: Do you think you could make a similar update to #465 when you have the chance, thanks.

taesungh (commented):
See also check failure in `apps/api/src/routers/user.py:129` (Unchanged files with check annotations), will look into how this can be enforced from the workflow.

taesungh (approved):
Feel like an adapter shouldn't be needed there, but oh well.

waalbert:
@taesungh Without the adapter, mypy didn't like `raw_application_data` being passed to `upload_resume` without validation. There might be another way to do this, which we can take a look later.

IanWearsHat (approved):

Issue #461, opened 2024-12-10
Allow multiple selections for pronouns


waalbert:
Currently, only mentor applications allow users to select multiple pronouns with the use of checkboxes. Hacker and volunteer applications need to be updated to allow for multiple pronouns.

PR #469, opened 2024-12-10
Changed Pronouns option to multiple select

IanWearsHat:
Part of #461 

Changes:
---
- Changed hacker pronouns to be multiple select
- Simplified SimpleRadio
- Changed unit test to include list for pronouns


waalbert (approved):
Thanks for the quick work Ian!

PR #471, opened 2024-12-10
Add role to application preface

noahk004:
Application preface now includes the type of the application that the user is filling out.

Screenshot 2024-12-10 152525
Screenshot 2024-12-10 152535
Screenshot 2024-12-10 152545


waalbert (approved):
Thanks for the quick work Noah!

PR #472, opened 2024-12-10
Added app descriptions to mentor and volunteer

### LINKED ISSUE #463: Add mentor and volunteer descriptions
When mentor and volunteer applications were on Google forms in the past, there were brief descriptions at the top illustrating roles and responsibilities. Because these applications were moved to our site this year, we need to add these brief descriptions to the application preface.

IanWearsHat:
Closes #463

### DIFF
--- a/apps/site/src/lib/components/forms/shared/ApplyConfirmation/ApplyConfirm.tsx
+++ b/apps/site/src/lib/components/forms/shared/ApplyConfirmation/ApplyConfirm.tsx
@@ -3,25 +3,32 @@ import ConfirmationDetails from "./ConfirmationDetails";
 
 interface ApplyConfirmProps {
 	applicationURL: "/apply" | "/mentor" | "/volunteer";
-	applicationType: "Hacker" | "Mentor" | "Volunteer";
+	role: "Hacker" | "Mentor" | "Volunteer";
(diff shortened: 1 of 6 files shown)


waalbert (approved):
Thanks for the quick work Ian!

apps/site/src/lib/components/forms/shared/ApplyConfirmation/ApplyConfirm.tsx
```diff
@@ -2,27 +2,43 @@ import getUserIdentity from "@/lib/utils/getUserIdentity";
...
-		: "In addition, I understand that I must show up for my scheduled shifts.";
+
+	const roleText = role === "Hacker" ? hackerClause : nonHackerClause;
+
+	const appDescription =
+		role === "Mentor"
+			? mentorDetails
+			: role === "Volunteer"
+			  ? volunteerDetails // eslint-disable-line
+			  : ""; // eslint-disable-line
```
taesungh, line 32: Suggestion: given the complexity of this condition, it might be better to have a component slot in `ApplyConfirmProps` so the consumer can specify more of a free-form component rather than trying to shape the usage within here and `ConfirmationDetails.

PR #476, opened 2024-12-11
Update/include email in processed application data

### LINKED ISSUE #475: Include email from user identity in processed application data
From #448 

>Please also include the email from the user identity in the processed application data to assist with https://github.com/HackAtUCI/irvinehacks-site/issues/454 and to save some trouble when the time comes to release decisions. This doesn't need to be an actual form field (see note in https://github.com/HackAtUCI/irvinehacks-site/pull/440), just from `user.email`.

Ensure this is done for all 3 application types and update tests accordingly

noahk004:
Add "email" field with value as user.email in processed_application_data in `apps/api/src/routers/user.py`

Closes #475

### DIFF
--- a/apps/api/src/routers/user.py
+++ b/apps/api/src/routers/user.py
@@ -223,7 +223,7 @@ async def _apply_flow(
     elif raw_application_data.application_type == "Volunteer":
         processed_application_data = (
             ProcessedApplicationDataUnionAdapter.validate_python(
-                _to_processed_application_data_volunteers(raw_app_data_dump, now)
+                _to_processed_application_data_volunteers(raw_app_data_dump, user, now)
             )
         )
     else:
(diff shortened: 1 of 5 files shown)


Bl20052005:
Issue: #474 is already merged into main, which conflicts with the change in user.py

apps/api/src/routers/user.py
```diff
@@ -150,6 +150,7 @@ async def _apply_flow(
     processed_application_data = ProcessedApplicationDataUnionAdapter.validate_python(
         {
             **raw_app_data_dump,
+            "email": user.email,
```
Bl20052005, line 153: when merged with main, this should be added in both hackers/mentors and volunteer handlers

Bl20052005 (dismissed):

waalbert:
Noting here for myself because I already mentioned this to you: The `email` field will also need to be added to `ProcessedHackerApplicationData`, `ProcessedMentorApplicationData`, and `ProcessedVolunteerData`

Issue #484, opened 2024-12-11
Integrate log handlers for logging to Axiom


taesungh:
Back in May 2024, Vercel announced that [Log Drains are now generally available](https://vercel.com/changelog/log-drains-are-now-generally-available) and restricted the usage on Hobby plans
> Log Drains are only available on Pro and Enterprise plans. Existing Hobby customers may continue to use Log Drains as configured, but no further usage or configuration is available.

This means we cannot create a new log drain for the IrvineHacks 2025 project on Vercel. We could try and swap with last year's project, but there's no guarantee that will work cleanly with renaming the projects, and that doesn't seem like a sustainable long-term solution (maybe Vercel forcefully disables the existing log drains in the future).

We previously used [Axiom](https://axiom.co) which has a Personal plan providing log retention for 30 days along with nice monitoring features such as streaming, querying, and alerting. Perhaps it might be good to invest in another solution with more permanent log retention, but Axiom was decently convenient and free to use. To continue using this with new Vercel projects, we'll need to manually connect Axiom's integrations to the Next.js app and the Python serverless function for the API.

For the Next.js app
- Install [next-axiom](https://github.com/axiomhq/next-axiom) and add the necessary middleware
- Include the `AxiomWebVitals` component in the root layout
- Update the server-side [`console.error` message in `getUserIdentity.ts`](https://github.com/HackAtUCI/irvinehacks-site/blob/00975fad7cea19dbd515acdf3aa957bfd0be2366/apps/site/src/lib/utils/getUserIdentity.ts#L17) to use a `Logger` from next-axiom

For the Python API
- Install [axiom-py](https://github.com/axiomhq/axiom-py)
- Add an instance of `AxiomHandler` as a handler to the root logger in [`index.py`](https://github.com/HackAtUCI/irvinehacks-site/blob/main/apps/api/index.py)

taesungh:
Looks like next-axiom has a peer dependency of `next>=14.0`, but we haven't completed #403 yet. Will see how far I can get with the current v13.5.6.

taesungh:
Update: considering just temporarily paying for Vercel Pro tier for two months since the interim approach from #487 needs some more work to get nicer contextual data.

taesungh:
Decided for now we'll just swap back the two projects (new and last year's) since we still have the previously created log drain (for now, who knows for how much longer). For next year, would be convenient to just consider paying for good logs during the two months we need them.

PR #487, opened 2024-12-11
Set up Axiom integration for logging

taesungh:
As part of #484, manually setting up a logging integration with Axiom since Vercel no longer allows creating new Log Drains on the Hobby tier.

Need to investigate further on improving the fields of the log records.

#### Changes
- Install and configure [next-axiom](https://github.com/axiomhq/next-axiom)
  - Wrap Next config and configure integration in `src/middleware.ts`
    - Include body in request details, possibly to be adjusted later
  - Include `AxiomWebVitals` in root layout
  - Use `Logger` object in `getUserIdentity`
- Add [axiom-py](https://github.com/axiomhq/axiom-py) as logging handler to API deployment
  - In `index.py`, add `AxiomHandler` to root logger and set level
  - Use axiom token and dataset from environment variables

#### Testing
- Created testing project in Vercel and connected temporary dataset
- Observed log messages from API are visible in the stream
  - Unfortunately, all of the contextual data from the request and Vercel is not available right now, only the Python log record from the Lambda function
- Data from requests to Next.js site are also included via the middleware
  - Unfortunately don't have `request.statusCode` right now unlike native integration

#### Notes
Before merging, will need to add new environment variables for the Axiom token and dataset name.


taesungh:
Closing as not planned, see notes in corresponding issue.

PR #501, opened 2024-12-15
ApplicantPortal shows the application that was submitted

IanWearsHat:
To test, add the following lines at the top of the `/me` route in `api/src/routers/user.py`, replacing "Role.ROLE" with the Hacker, Mentor, and Volunteer roles.

```
return IdentityResponse(
        uid="edu",
        status=Status.PENDING_REVIEW,
        roles=[Role.APPLICANT, Role.ROLE],
    )
```

Then navigate to `/portal`.


waalbert (approved):
Thanks for this quick enhancement Ian!

PR #502, opened 2024-12-17
Links to other apps

### LINKED ISSUE #460: Add links to other applications in application preface
In the case the user wants to apply for another role, they should easily be able to switch to the other applications. Links should be added at the bottom of `ConfirmationDetails` with something along the lines of "Interested in being a `Mentor` or `Volunteer` instead?"

cyriljoby:
Resolves #460 . Added links to other 2 applications at the bottom of each application page

### DIFF
--- a/apps/site/src/lib/components/forms/shared/ApplyConfirmation/ConfirmationDetails.tsx
+++ b/apps/site/src/lib/components/forms/shared/ApplyConfirmation/ConfirmationDetails.tsx
@@ -15,6 +15,9 @@ export default async function ConfirmationDetails({
 	roleText,
 	role,
 }: ConfirmationDetailsProps) {
+	const otherRoles = ["Hacker", "Mentor", "Volunteer"].filter(
+		(r) => r !== role,
+	);
 	return (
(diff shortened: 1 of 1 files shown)


apps/site/src/lib/components/forms/shared/ApplyConfirmation/ConfirmationDetails.tsx
```diff
@@ -61,6 +64,22 @@ export default async function ConfirmationDetails({
...
 			/>
+
+			<p>
+				Interested in being a{" "}
+				{otherRoles.map((role, index) => (
+					<span key={role}>
+						<a
+							href={role === "Hacker" ? `/apply` : `/${role.toLowerCase()}`}
+							className="text-[#FBA80A] underline"
+						>
```
waalbert, line 75: Suggestion: can you use `<Link>` from `next/link` rather than the standard `<a>` tag so the page doesn't reload when you click the links?

waalbert (approved):
Thanks for the work Cyril!

PR #506, opened 2024-12-17
deleted Demo.tsx and its router demo.py

### LINKED ISSUE #406: Remove Demo page and API router
Similar to #56/#57 and following from #41 and #72, we've verified the API works properly even with data fetching in server components, so we should remove the demo page in [`Demo.tsx`](https://github.com/HackAtUCI/irvinehacks-site/blob/7e8175001f074a0e715ab6af0a8734066691173f/apps/site/src/app/demo/Demo.tsx) and the corresponding API router in [`demo.py`](https://github.com/HackAtUCI/irvinehacks-site/blob/7e8175001f074a0e715ab6af0a8734066691173f/apps/api/src/routers/demo.py). As written, the `add-user` endpoint will break the expected database schema for the Users collection.

cyriljoby:
Resolves #406 by deleting the Demo.tsx , its corresponding page.tsx, and the Demo.py files.


waalbert (approved):
Thanks for the quick work Cyril!

PR #509, opened 2024-12-20
removed ref on input and useEffect to clear it

### LINKED ISSUE #479: Remove unnecessary `useEffect` in `ResumeInformation`
Noticed while working on #426 that there's a `ref` and a `useEffect` in [`ResumeInformation.tsx`](https://github.com/HackAtUCI/irvinehacks-site/blob/00975fad7cea19dbd515acdf3aa957bfd0be2366/apps/site/src/lib/components/forms/shared/ResumeInformation.tsx) for the input field originally added in #103. The `useEffect` sets the current value of the input to an empty string, but I think the browser already does that anyway which is why we account for the empty file scenario in the `/user/apply` endpoint, so I don't think we actually need the ref or effect.

cyriljoby:
Resolves #479 by removing ref attached to an input and useEffect to clear it

### DIFF
--- a/apps/site/src/lib/components/forms/shared/ResumeInformation.tsx
+++ b/apps/site/src/lib/components/forms/shared/ResumeInformation.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { ChangeEvent, useEffect, useState, useRef } from "react";
+import { ChangeEvent, useState } from "react";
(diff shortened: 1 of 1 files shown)


waalbert (approved):
Thanks for the cleanup Cyril!

PR #514, opened 2024-12-21
[Admin] Hacker Applicants Overview

IanWearsHat:
Part of #510, will open 2 more PRs after this one is merged

- Insert document in MongoDB under `settings` collection similar to the following:
```
{
    _id: 'hacker_score_thresholds',
    accept: 8,
    waitlist: 5,
    reject: 3
}
```
- Insert document in MongoDB under `users` collection similar to the following:
```
{
    _id: 'edu.uci.app22',
    application_data: {
        pronouns: [
            'he',
            'she'
        ],
        ethnicity: 'Two-or-more',
        is_18_older: false,
        school: 'UC San Diego',
        education_level: 'third-year-undergrad',
        major: 'Software Engineering',
        is_first_hackathon: false,
        linkedin: null,
        portfolio: null,
        frq_change: 'dsad',
        frq_video_game: 'dsa',
        email: 'dev@irvinehacks.com',
        resume_url: null,
        submission_time: ISODate('2024-12-20T17:32:59.973Z'),
        reviews: [
            [
                ISODate('2024-12-20T17:41:22.634Z'),
                'edu.uci.ian4',
                4
            ],
            [
                ISODate('2024-12-20T17:42:01.098Z'),
                'edu.uci.ian5',
                10
            ]
        ]
    },
    first_name: 'dwa',
    last_name: 'sds',
    roles: [
        'Applicant',
        'Hacker'
    ],
    status: 'ACCEPTED'
}
```
- Navigate to `/admin/applicants/hackers`
- Ensure that cards show up


apps/api/src/admin/applicant_review_processor.py
```diff
@@ -0,0 +1,75 @@
...
+    applicant_record["num_reviewers"] = _get_num_unique_reviewers(applicant_record)
+
+
+def _include_avg_score(applicant_record: dict[str, Any]) -> None:
+    applicant_record["avg_score"] = _get_avg_score(
+        applicant_record["application_data"]["reviews"]
+    )
+
+
+def _delete_reviews(applicant_record: dict[str, Any]) -> None:
```
taesungh, line 39: Suggestion: naming this "delete" initially made me think the database record was being updated. A less scary name might be "exclude" or "omit".

apps/api/src/admin/applicant_review_processor.py
```diff
@@ -0,0 +1,75 @@
...
+    unique_reviewers = {t[1] for t in reviews}
+    if len(unique_reviewers) < 2:
+        return -1
+
+    last_score = _get_last_score(unique_reviewers.pop(), reviews)
+    last_score2 = _get_last_score(unique_reviewers.pop(), reviews)
+    return (last_score + last_score2) / 2
+
+
+def _set_decision_based_on_threshold(
```
taesungh, line 60: Suggestion: Similar to the previous note, naming this "set" might imply this updates the database record. Consider renaming as "include" similar to the other functions.

apps/api/src/admin/applicant_review_processor.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+def _delete_reviews(applicant_record: dict[str, Any]) -> None:
+    del applicant_record["application_data"]["reviews"]
+
+
+def _get_last_score(reviewer: str, reviews: list[tuple[str, str, float]]) -> float:
+    for i in range(len(reviews) - 1, -1, -1):
+        if reviews[i][1] == reviewer:
+            return reviews[i][2]
```
taesungh, line 46: Suggestion: using an iterable directly would be simpler than using an index, e.g.
```python
for review in reversed(reviews):
```

apps/api/src/admin/applicant_review_processor.py
```diff
@@ -0,0 +1,69 @@
...
+
+def include_hacker_app_fields(
+    applicant_record: dict[str, Any], accept_threshold: float, waitlist_threshold: float
+) -> None:
+    _include_decision_based_on_threshold(
+        applicant_record, accept_threshold, waitlist_threshold
+    )
+    _include_num_reviewers(applicant_record)
+    _include_avg_score(applicant_record)
+    _omit_reviews(applicant_record)
```
taesungh, line 14: Thought: since the resulting record is validated through the type adapter for HackerApplicantSummary, I don't think it's necessary to omit the reviews field since the model validator will ignore it.

apps/api/src/routers/admin.py
```diff
@@ -319,6 +364,21 @@ async def _process_batch(batch: tuple[dict[str, Any], ...], decision: Decision)
     )
 
 
+async def _update_batch_of_decisions(
+    batch: tuple[dict[str, Any], ...],
+    decision: Decision,
+    *,
+    update_field: Literal["status", "decision"],
```
taesungh, line 371: Thought: I'm not sure if y'all have worked out a plan for decision release, but we previously just set the status as the final decision value without tracking a separate value for decision (see discussion in #520). If the same strategy will be used, then the `_process_batch` function could be reverted to how it was before.

apps/api/tests/test_admin.py
```diff
@@ -119,7 +120,7 @@ def test_can_include_decision_from_reviews() -> None:
         },
     }
 
-    admin._include_review_decision(record)
+    applicant_review_processor.include_review_decision(record)
```
taesungh, line 123: Chore: since this test (and the one below) are now about functions in the new module, it would also make more sense to move it to the new test file.

apps/site/src/lib/admin/useHackerApplicants.ts
```diff
@@ -0,0 +1,34 @@
+import axios from "axios";
+import useSWR from "swr";
+
+import { Decision, Status } from "./useApplicant";
```
taesungh, line 4: Note: these are being moved in #519, depending on whichever is merged first, will need to update the latter.

apps/api/src/admin/applicant_review_processor.py
```diff
@@ -0,0 +1,69 @@
+from typing import Any
+
+from models.ApplicationData import Decision
+
+
+def include_hacker_app_fields(
+    applicant_record: dict[str, Any], accept_threshold: float, waitlist_threshold: float
```
taesungh, line 7: Thought [can handle later]: might be good to define another model for the pre-processed data so we can have stronger type checking rather than relying on `Any`.

apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
```diff
@@ -0,0 +1,134 @@
...
+			header={<Header counter={counter}>Applicants</Header>}
+		/>
+	);
+}
+
+const CardHeader = ({ _id, first_name, last_name }: HackerApplicantSummary) => {
+	const followWithNextLink = useFollowWithNextLink();
+	return (
+		<Link
+			href={`/admin/applicants/${_id}`}
```
taesungh, line 122: Note [for later]: we'll eventually need to create new subpages for each type of applicant and reorganize some of the shared components.

taesungh (approved):
Thanks for putting this together, think it'll definitely help to have more integrated review capabilities in the site. Some other minor notes, can handle now or later.

waalbert:
Thoughts on making the path `/applicants` into a route group `(applicants)` so that hacker apps can be accessed 
at `/admin/hackers` instead of `/admin/applicants/hackers` and in the future `/mentors` and `/volunteers` can also be at this higher path level?

IanWearsHat:
I think that makes sense especially if we have nothing to show at /admin/applicants.

taesungh:
I suggested before in #510 that the parent page could be used to show more summary charts about the different types of applicants and with links to the subpage (and the side navigation would include an expandable link group), but whatever y'all think makes more sense.

waalbert:
If we choose to have the summary charts at `/admin/dashboard,` then I agree that we can have `/admin/dashboard/hackers`. In that case, I will create a placeholder page for `/admin/dashboard`

apps/site/src/lib/admin/useHackerApplicants.ts
waalbert: #519 merged first, so this PR will need to be updated

apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
taesungh, line 1: Could you also delete `app/admin/applicants/Applicants.tsx` so Git can detect this was like a moved file?

taesungh (approved):

waalbert (approved):
Thanks for the work Ian!

PR #519, opened 2024-12-22
Refactor user roles from admin site to main site

### LINKED ISSUE #516: Reorganize user roles in the Admin site into an enum
Mentioned in #429, the Admin site currently has the user roles defined as strings in [`authorization.ts`](https://github.com/HackAtUCI/irvinehacks-site/blob/3752272e3bba64e62ded6441b3a79d095e11ed2d/apps/site/src/lib/admin/authorization.ts). There are a few places these are referenced for purposed beyond authentication, and there's some partial duplication in [`useParticipants.ts`](https://github.com/HackAtUCI/irvinehacks-site/blob/3752272e3bba64e62ded6441b3a79d095e11ed2d/apps/site/src/lib/admin/useParticipants.ts). I think such usages will also expand while working on #513, so it would be good to reorganize these into an enum to have named references rather than comparing strings.

taesungh:
Using enums to represent the different possible user roles to have stronger named references.
Resolves #516. Will create a new issue about reorganizing how `Status` is used in the Applicant Portal.

#### Changes
- Create new enums for user roles for both the admin site and main site
- Move status enums from `useApplicant` to the new `userRecord.ts`
- Move `isNonHacker` check to `ParticipantAction`
- The Portal still needs to be updated to use the new common enums

### DIFF
--- a/apps/api/src/models/user_record.py
+++ b/apps/api/src/models/user_record.py
@@ -11,7 +11,7 @@ from services.mongodb_handler import BaseRecord
 class Role(str, Enum):
     """
     Possible roles of organizers and participants.
-    The values are the display labels to be shown on the Admin site.
+    The values are the display labels to be shown throughout the site.
     """
 
     APPLICANT = "Applicant"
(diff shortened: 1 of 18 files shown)


IanWearsHat (approved):

waalbert (approved):
Thanks Taesung!

PR #525, opened 2024-12-23
Enable request memoization for `getUserIdentity`

### LINKED ISSUE #168: Next.js build complains about dynamic server usage

### LINKED ISSUE #524: Fix request memoization for `getUserIdentity`
Relating to #523, the [`getUserIdentity`](https://github.com/HackAtUCI/irvinehacks-site/blob/3752272e3bba64e62ded6441b3a79d095e11ed2d/apps/site/src/lib/utils/getUserIdentity.ts) function is currently requested by different components in the same render which each make their own request to the API. Normally, because of [Request Memoization](https://nextjs.org/docs/app/building-your-application/caching#request-memoization), React would automatically cache `fetch`  requests, but we currently go through Axios which uses XHR by default.

Axios v1.7.0 added a new [fetch adapter](https://github.com/axios/axios#-fetch-adapter), so we could try switching to that or manually use [React's `cache` function](https://nextjs.org/docs/app/building-your-application/caching#react-cache-function) on `getUserIdentity`. Another option is replacing Axios with `fetch` because of #168, but Axios does provide some convenience features (although not widely needed throughout the site).

Note this memoization happens when React renders the component tree per request, not across multiple requests.

taesungh:
Resolves #524 and fixes #168.

#### Changes
- Wrap function in [RSC `cache`](https://react.dev/reference/react/cache) so the API request is made only once each time the page is rendered (for each server request)
  - Previously, the request might be unnecessarily made multiple times
- Call `cookies()` inside the function to indicate to Next.js that pages using said function must be [dynamically rendered](https://nextjs.org/docs/app/building-your-application/rendering/server-components#dynamic-rendering)
  - In the future, [`dynamicIO` in Next.js 15](https://nextjs.org/docs/app/api-reference/config/next-config-js/dynamicIO) might work better instead
- Remove stray comment in `app/(main)/layout.tsx`

#### Testing
1. Start the local development servers
2. Navigate to `/apply` in a web browser
3. Observe in the FastAPI logs that the `/user/me` endpoint is accessed only one time rather than three times
4. Reload the same page
5. Observe the `/user/me` endpoint is accessed again, demonstrating that the cache is per request

### DIFF
--- a/apps/site/src/app/(main)/layout.tsx
+++ b/apps/site/src/app/(main)/layout.tsx
@@ -20,7 +20,6 @@ export default function Layout({ children }: PropsWithChildren) {
 			style={{ backgroundImage: `url(${stars.src})` }}
 			className="overflow-x-hidden bg-top bg-repeat-y bg-[length:100%]"
 		>
-			{/* reference: https://github.com/pmndrs/react-three-next */}
 			<NavbarParent />
 			{children}
 			<Footer />
--- a/apps/site/src/lib/utils/getUserIdentity.ts
+++ b/apps/site/src/lib/utils/getUserIdentity.ts
@@ -1,3 +1,6 @@
+import { cache } from "react";
+import { cookies } from "next/headers";
+
 import axios from "axios";
 
 import { Role, Uid } from "@/lib/userRecord";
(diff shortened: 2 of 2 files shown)


waalbert (approved):
Thanks for caching `getUserIdentity` and fixing #168 Taesung!

PR #526, opened 2024-12-23
[Admin] New hacker review logic and indicator

IanWearsHat:
Part of #510

- Insert document in MongoDB under `settings` collection similar to the following:
```
{
    _id: 'hacker_score_thresholds',
    accept: 8,
    waitlist: 5
}
```
- Insert document in MongoDB under `users` collection similar to the following:
```
{
    _id: 'edu.uci.app22',
    application_data: {
        pronouns: [
            'he',
            'she'
        ],
        ethnicity: 'Two-or-more',
        is_18_older: false,
        school: 'UC San Diego',
        education_level: 'third-year-undergrad',
        major: 'Software Engineering',
        is_first_hackathon: false,
        linkedin: null,
        portfolio: null,
        frq_change: 'dsad',
        frq_video_game: 'dsa',
        email: 'dev@irvinehacks.com',
        resume_url: null,
        submission_time: ISODate('2024-12-20T17:32:59.973Z'),
        reviews: [
            [
                ISODate('2024-12-20T17:41:22.634Z'),
                'edu.uci.ian4',
                4
            ],
            [
                ISODate('2024-12-20T17:42:01.098Z'),
                'edu.uci.ian5',
                10
            ]
        ]
    },
    first_name: 'dwa',
    last_name: 'sds',
    roles: [
        'Applicant',
        'Hacker'
    ],
    status: 'ACCEPTED'
}
```

## All applicants overview
- Navigate to `/admin/applicants/hackers`
- Ensure that reviewer indicator works as expected.
  - Add 0, 1, and 2 reviewers in the `application_data.reviews` array and ensure that reviewer indicator correctly displays number of reviewers

## Hacker applicant overview
- Navigate to `/admin/applicants/hackers`
- Click on the applicant card to navigate to that specific application
- Submit score in top right input box
- Ensure review log underneath updates correctly

## New score encoding
- Navigate to the specific application page `/admin/applicants/(uid)`
- Change the "Hacker" string in the roles array in the mongo doc to "Mentor" or "Volunteer"
- Ensure that the review action and the review log changes
- Submit a review
- Ensure that the correct review score is created in the mongo doc


apps/site/src/lib/admin/useApplicant.ts
taesungh, line 1: Thought: I'm not sure how different the review logic will be for mentors/volunteers, but we'll need to reorganize this at some point since right now there's a lot of hacker-specific logic. Thinking it could possibly be a higher-order function to reduce some duplication or perhaps a generic function. Alternatively, we just make three copies since the `ApplicationData` for each will be different anyway.

If having separate versions makes more sense, could you move this to be within `app/admin/applicants/hackers/`?

apps/api/src/routers/admin.py
```diff
@@ -172,6 +172,47 @@ async def submit_review(
...
+        ["_id", "application_data.reviews"],
+    )
+    if not applicant_record:
+        log.error("Could not retrieve applicant after submitting review")
+        raise HTTPException(status.HTTP_500_INTERNAL_SERVER_ERROR)
+
+    num_reviewers = applicant_review_processor.get_num_unique_reviewers(
+        applicant_record
+    )
+    if num_reviewers > 1:
```
taesungh, line 208: Suggestion: could you include a comment above this condition explaining that the status is set when there are at least two reviewers and/or use `>= 2` instead of `> 1`? Took me a moment to understand why this was moved out.

apps/api/src/routers/admin.py
```diff
@@ -172,6 +172,47 @@ async def submit_review(
         raise HTTPException(status.HTTP_500_INTERNAL_SERVER_ERROR)
 
 
+@router.post("/review/hacker")
+async def submit_hacker_review(
+    applicant: str = Body(),
+    score: float = Body(),
```
taesungh, line 178: Suggestion: I know this is copied from above, but I think it would be better to put these fields in a model that describes the [request body](https://fastapi.tiangolo.com/tutorial/body/).

apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
```diff
@@ -77,6 +78,11 @@ function HackerApplicants() {
 						header: "Status",
 						content: ApplicantStatus,
 					},
+					{
+						id: "reviewers",
+						header: "",
```
taesungh, line 83: Question: is there a particular reason you left the header empty?

apps/site/src/app/admin/applicants/components/ApplicantReviewerIndicator.tsx
```diff
@@ -0,0 +1,58 @@
+import {
+	Box,
+	SpaceBetween,
+	StatusIndicator,
+} from "@cloudscape-design/components";
```
taesungh, line 5: Note: I'm not sure if it actually matters with the transpiler in Next.js, but it might be safer to import these directly, see guide on [Using Cloudscape components](https://cloudscape.design/get-started/integration/using-cloudscape-components/).

apps/site/src/app/admin/applicants/components/ApplicantReviewerIndicator.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const IndicatorContainer = ({
+	reviewerNumber,
+	type,
+}: IndicatorContainerProps) => {
+	return (
+		<>
+			<Box variant="awsui-key-label">Reviewer {reviewerNumber}</Box>
+			{type === "notReviewed" ? (
+				<StatusIndicator type="error">Not Reviewed</StatusIndicator>
```
taesungh, line 20: Thought: the Pending or In Progress types might make more sense here than Error.

apps/site/src/app/admin/applicants/components/ApplicantReviewerIndicator.tsx
```diff
@@ -0,0 +1,58 @@
...
+					<IndicatorContainer reviewerNumber="2" type="notReviewed" />
+				</>
+			) : num_reviewers === 2 ? (
+				<>
+					<IndicatorContainer reviewerNumber="1" type="reviewed" />
+					<IndicatorContainer reviewerNumber="2" type="reviewed" />
+				</>
+			) : (
+				<></>
+			)}
```
taesungh, line 53: Suggestion: the duplication here could be reduced by using a condition based on an index, e.g.
```tsx
{[1, 2].map((n) => <IndicatorContainer number={n} reviewed={n >= num_reviewers} />}
```

apps/api/src/routers/admin.py
```diff
@@ -184,14 +225,33 @@ async def release_decisions() -> None:
...
+    """Update hacker applicant status based on decision and send decision emails."""
+    records = await mongodb_handler.retrieve(
+        Collection.USERS,
+        {"status": Status.REVIEWED},
+        ["_id", "application_data.reviews", "first_name"],
+    )
+
+    thresholds = await mongodb_handler.retrieve_one(
+        Collection.SETTINGS, {"_id": "hacker_score_thresholds"}, ["accept", "waitlist"]
+    )
```
taesungh, line 242: Note: I think this query is also written inside the review function, might help to extract it to another private function to reduce duplication, or perhaps integrate it inside `applicant_review_processor`.

apps/api/src/routers/admin.py
```diff
@@ -172,6 +172,47 @@ async def submit_review(
         raise HTTPException(status.HTTP_500_INTERNAL_SERVER_ERROR)
 
 
+@router.post("/review/hacker")
+async def submit_hacker_review(
+    applicant: str = Body(),
+    score: float = Body(),
+    reviewer: User = Depends(require_role({Role.REVIEWER})),
+) -> None:
+    """Submit a review decision from the reviewer for the given hacker applicant."""
```
taesungh, line 181: Thought: do you know what the review process for the mentors/volunteers will look like? Maybe instead of creating separate endpoints, we could combine them and use dynamic logic based on the applicant type inside for the reviewer count. If the plan for mentors/volunteers is to use discrete discussions, these could be internally encoded as numeric scores (e.g. accept = 100, reject = 0) while the UI shows the decision names.

taesungh (commented):
Nice work so far, have an interesting thought about what the review system for the other types of applicants will look like. Thinking we could maybe combine some things instead of creating separate versions.

apps/api/src/routers/admin.py
IanWearsHat: After discussion with @waalbert , the review process for mentors/volunteers should be the exact same as last year. So 1 reviewer with the same review logic as before. The only thing that would change is separating them into their own tabs in the portal. This would keep us from having to rewrite all of the review logic.

apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
IanWearsHat: I didn't want to add a header because it didn't looked redundant and I couldn't find a text style that looked good with the labels that I added.

image

apps/api/src/routers/admin.py
waalbert: Yep, in this case we can have all apps share the same endpoint and based on the type of app, assign the scores accordingly.

apps/site/src/app/admin/applicants/[uid]/components/HackerApplicantActions.tsx
```diff
@@ -0,0 +1,47 @@
...
+	return (
+		<SpaceBetween direction="horizontal" size="xs">
+			<Input
+				onChange={({ detail }) => setValue(detail.value)}
+				value={value}
+				type="number"
+				inputMode="decimal"
+				placeholder="Applicant score"
+				step={0.1}
+			/>
```
Comment (waalbert) [inline apps/site/src/app/admin/applicants/[uid]/components/HackerApplicantActions.tsx:34]: Note: according to the finalized rubric created by Applications, the only possible values for scores are 0, 0.5, 1, 1.5, 2, and so on, so changing the `step` value to `0.5` would make more sense

apps/site/src/app/admin/applicants/components/ApplicantReviewerIndicator.tsx
```diff
@@ -0,0 +1,46 @@
...
+	hasReviewed,
+}: IndicatorContainerProps) => {
+	return (
+		<>
+			<Box variant="awsui-key-label">Reviewer {number}</Box>
+			{hasReviewed ? (
+				<StatusIndicator>Reviewed</StatusIndicator>
+			) : (
+				<StatusIndicator type="pending">Not Reviewed</StatusIndicator>
+			)}
```
waalbert, line 21: Note [nice to have]: next to "Reviewer 1" and "Reviewer 2," I think it will be helpful to show the UID of the reviewer next to them

apps/site/src/app/admin/applicants/[uid]/components/HackerApplicantActions.tsx
```diff
@@ -1,28 +1,43 @@
...
 	applicant,
+	reviews,
 	submitReview,
 }: ApplicantActionsProps) {
-	const { roles } = useContext(UserContext);
+	const { uid, roles } = useContext(UserContext);
 	const [value, setValue] = useState("");
+	// const canReview = either there are less than 2 reviewers or uid is in current reviews
+	const uniqueReviewers = new Set(reviews.map((review) => review[1]));
+	const canReview = uid
```
Comment (waalbert) [inline apps/site/src/app/admin/applicants/[uid]/components/HackerApplicantActions.tsx:32]: Nitpick: move comment to line 31

apps/site/src/app/admin/applicants/[uid]/components/HackerApplicantActions.tsx
```diff
@@ -31,16 +46,23 @@ function HackerApplicantActions({
...
+				disabled={!canReview}
 			>
 				Submit
 			</Button>
 		</SpaceBetween>
+	) : (
+		<Box variant="awsui-key-label" color="text-status-warning">
+			Two reviewers have already submitted reviews. Log in as one of them to
+			submit a review.
+		</Box>
```
Comment (waalbert) [inline apps/site/src/app/admin/applicants/[uid]/components/HackerApplicantActions.tsx:65]: Suggestion: since this component already can see the reviews, I think it'll be helpful to add the reviewers' UIDs to the warning text.

apps/api/src/routers/admin.py
```diff
@@ -149,28 +152,49 @@ async def applicant_summary() -> dict[ApplicantStatus, int]:
...
+    applicant_review: ReviewRequestModel,
     reviewer: User = Depends(require_role({Role.REVIEWER})),
 ) -> None:
-    """Submit a review decision from the reviewer for the given applicant."""
-    log.info("%s reviewed applicant %s", reviewer, applicant)
+    """Submit a review decision from the reviewer for the given hacker applicant."""
+    log.info("%s reviewed hacker %s", reviewer, applicant_review.applicant)
 
-    review: OtherReview = (utc_now(), reviewer.uid, decision)
+    review: Review = (utc_now(), reviewer.uid, applicant_review.score)
```
waalbert, line 162: Suggestion: even though the frontend hides the ability to add reviews once there have been 2 reviewers, it's still important to double check the number of unique reviewers in the backend before attempting to add a review to the DB. In practice, since only Hack orgs have access to the admin portal and its API endpoints, this shouldn't matter too much but for the sake of robustness and extra security in the case of a malicious user trying to submit a review straight to this endpoint.

apps/api/src/routers/admin.py
```diff
@@ -149,28 +152,49 @@ async def applicant_summary() -> dict[ApplicantStatus, int]:
...
+        ["_id", "application_data.reviews", "roles"],
+    )
+    if not applicant_record:
+        log.error("Could not retrieve applicant after submitting review")
         raise HTTPException(status.HTTP_500_INTERNAL_SERVER_ERROR)
 
+    if Role.HACKER in applicant_record["roles"]:
+        num_reviewers = applicant_review_processor.get_num_unique_reviewers(
+            applicant_record
+        )
```
taesungh, line 181: Thought: I'm not entirely sure how [MongoDB's casual consistency](https://www.mongodb.com/docs/manual/core/causal-consistency-read-write-concerns/) works, but there may be an edge case where the previously written data is not immediately available at the next read. Will need to look into this more.

apps/api/tests/test_admin.py
```diff
@@ -111,31 +112,80 @@ def test_can_retrieve_applicants(
...
+        },
+    }
 
-    mock_mongodb_handler_retrieve_one.return_value = REVIEWER_IDENTITY
+    mock_mongodb_handler_retrieve_one.side_effect = [
+        REVIEWER_IDENTITY,
+        returned_record,
+        REVIEWER_IDENTITY,
+        returned_record,
+    ]
```
taesungh, line 137: Note [for later]: this mocking is getting a bit ridiculous, I think it might be good to look into overriding the FastAPI dependency (think Sam and I did this in some other test file before).

apps/site/src/lib/decisionScores.ts
```diff
@@ -0,0 +1,18 @@
+import { Decision } from "./userRecord";
+
+/** For use in nonhacker applicant reviews.*/
+const acceptScore = "100";
+const waitlistScore = "-1";
+const rejectScore = "0";
```
taesungh, line 6: Question: is there a specific reason these are strings rather than numbers?

apps/site/src/app/admin/applicants/components/ApplicantReviewerIndicator.tsx
taesungh: And same in `HackerApplicantActions.tsx`

apps/site/src/app/admin/applicants/components/ApplicantReviewerIndicator.tsx
taesungh: Note: just pointing out for context that in previous years we omitted who submitted the review (unless it was yourself) to reduce potential inter-reviewer bias. Ultimately, up to you all to decide what to show.

apps/site/src/app/admin/applicants/components/ApplicantReviewerIndicator.tsx
```diff
@@ -0,0 +1,51 @@
...
+	reviewers,
+}: ApplicantReviewerIndicatorProps) {
+	return (
+		<SpaceBetween size="l">
+			{[0, 1].map((n) => (
+				<IndicatorContainer
+					key={n}
+					displayNumber={n + 1}
+					reviewer={reviewers[n] ? reviewers[n] : ""}
+					hasReviewed={reviewers.length >= n + 1}
```
taesungh, line 44: Suggestion: the `hasReviewed` prop is now redundant since that can be indicated by whether or not `reviewers[n]` is a string or `undefined`.

apps/site/src/app/admin/applicants/[uid]/components/HackerApplicantActions.tsx
```diff
@@ -0,0 +1,91 @@
...
+		</SpaceBetween>
+	) : (
+		<SpaceBetween size="xxs">
+			<Box variant="awsui-key-label" color="text-status-info">
+				<ColoredTextBox text={uniqueReviewers[0].split(".").at(-1)} /> and{" "}
+				<ColoredTextBox text={uniqueReviewers[1].split(".").at(-1)} /> already
+				submitted reviews.
+			</Box>
+			<Box variant="awsui-key-label" color="text-status-info">
+				Log in as one of them to submit a review.
```
Comment (taesungh) [inline apps/site/src/app/admin/applicants/[uid]/components/HackerApplicantActions.tsx:85]: Question: this direction seems rather puzzling since in practice, reviewers cannot impersonate each other. Is there a better way to indicate the course of action here?

apps/site/src/lib/admin/useApplicant.ts
taesungh: Can revisit later, but I think with the common review endpoint, this could be turned into a shared generic function for the different types of applicants.

apps/api/src/routers/admin.py
```diff
@@ -48,11 +48,16 @@ class HackerApplicantSummary(BaseRecord):
     last_name: str
     status: str
     decision: Optional[Decision] = None
-    num_reviewers: int
+    reviewers: list[str] = []
     avg_score: float
     application_data: ApplicationDataSummary
 
 
+class ReviewRequestModel(BaseModel):
```
taesungh, line 56: Nitpick: including "Model" in the name feels a bit redundant (like Hungarian notation).

taesungh (commented):
Looking pretty good, some additional minor notes on little things that could be further improved.

apps/api/src/routers/admin.py
waalbert: A reviewer will be allowed to change their score afterwards. To be more clear on what I meant previously by double checking the number of unique reviewers, if the user isn't already one of the 2 reviewers, then prevent a new review from being added, but if the user is one of the existing reviewers, then allow them to add another (updated) review.

apps/site/src/app/admin/applicants/components/ApplicantReviewerIndicator.tsx
waalbert: Don't think log/applications have decided if and how to assign reviewers, but if they decide to keep track of who's assigned to which app, then I think showing who submitted reviews would be fine because reviewers could still see who their "co-reviewer" would be. Showing reviewers would also help everyone keep track of all the apps, in my opinion.

apps/site/src/app/admin/applicants/[uid]/components/HackerApplicantActions.tsx
Comment (waalbert) [reply apps/site/src/app/admin/applicants/[uid]/components/HackerApplicantActions.tsx:85]: This can be changed to something like "Contact Rosalind, Nicole, and Albert if you think this is an error."

apps/site/src/lib/decisionScores.ts
IanWearsHat: The Input component for inputting scores in `HackerApplicantActions.tsx` returns a string when calling `onChange`, so the `submitReview` function in `useApplicant.ts` expects a string.

Because I'm using the same route for submitting hacker, mentor, and volunteer reviews, I made these specific scores also strings.

Something I could do is set the score parameter for `submitReview` to be of type `string | number` so that these scores can be numbers.

apps/api/src/routers/admin.py
```diff
@@ -176,23 +171,44 @@ async def submit_review(
...
+        # or reviewer is one of the reviewers
+        if len(unique_reviewers) < 2 or str(reviewer) in unique_reviewers:
+            update_query.update({"$push": {"application_data.reviews": review}})
+        else:
+            log.error(
+                "%s tried submitting a review, but %s already has two reviewers",
+                reviewer,
+                app,
+            )
+            raise HTTPException(status.HTTP_403_FORBIDDEN)
```
taesungh, line 190: Suggestion: since the else clause terminates the control flow, inverting the condition might be easier to follow, i.e.
```python
if len(unique_reviewers) >= 2 and reviewer.uid not in unique_reviewers:
    raise ...

...
```

taesungh (commented):
Sorry to keep dragging along this PR, maybe note for next time that keeping things small would help with faster turnaround.

apps/site/src/lib/decisionScores.ts
waalbert: I personally think rather than using type `string | number` it'd be better to leave these score types as `number` types and then converting the `value` argument string into a `number` before passing it into `submitReview` in `HackerApplicantActions`

apps/api/src/routers/admin.py
waalbert: Follow up to this, it seems that log/applications will be deciding how to assign apps to reviewers in their next meeting in a few days.

apps/site/src/app/admin/applicants/components/ApplicantReviewerIndicator.tsx
waalbert: Follow up to this, it seems that log/applications will be deciding how to assign apps to reviewers in their next meeting in a few days, so we can keep this. I'll keep everyone updated in case something happens.

taesungh (approved):

waalbert (approved):
Thanks for handling this big task Ian!

PR #528, opened 2024-12-24
Reorganize search params for server components

### LINKED ISSUE #521: Elevate search params used by `GuestLoginVerificationForm`
Noticed in #403:

The [`GuestLoginVerificationForm` calls `useSearchParams`](https://github.com/HackAtUCI/irvinehacks-site/blob/3752272e3bba64e62ded6441b3a79d095e11ed2d/apps/site/src/app/(main)/guest-login/components/GuestLoginVerificationForm.tsx#L15) for the user's email and the `return_to` location. At build time, Next.js currently thinks this page can be statically rendered because of #168, but [Static Rendering for `useSearchParams`](https://nextjs.org/docs/app/api-reference/functions/use-search-params#static-rendering) requires a suspense boundary to avoid de-opting to client-side rendering.

With Next.js 13, this wasn't surfaced, but when attempting to build with Next.js 14, this causes an error:
> site:build: ⨯ useSearchParams() should be wrapped in a suspense boundary at page "/guest-login". Read more: https://nextjs.org/docs/messages/missing-suspense-with-csr-bailout

To fix this, we can either force dynamic rendering on the page (or properly fix #168) or rely on the page's [`searchParams`](https://nextjs.org/docs/14/app/api-reference/file-conventions/page#searchparams-optional) and pass those as props to the verification form component. This will unblock upgrading to Next.js 14.

Another option is having the form component use [Streaming with Suspense](https://nextjs.org/docs/14/app/building-your-application/routing/loading-ui-and-streaming#streaming-with-suspense), but this Guest Login page is inherently dynamic and doesn't really need such optimization, especially when there are no database queries.

waalbert:
## Changes
- Removes the use of `useSearchParams` in `GuestLoginVerificationForm`
- Uses `searchParams` from `GuestLogin` and passes `searchParams` as props to `GuestLoginVerificationForm`
- These changes allow for `GuestLoginVerificationForm` to no longer need to be a client component
- Unmark `searchParams` as optional as Next.js always provides this to page components
- Unpacks `searchParams` before using query parameters

Closes #521 and unblocks #403 to allow upgrading to Next.js 14.

### DIFF
--- a/apps/site/src/app/(main)/guest-login/GuestLogin.tsx
+++ b/apps/site/src/app/(main)/guest-login/GuestLogin.tsx
@@ -7,6 +7,7 @@ async function GuestLogin({
 	searchParams,
 }: {
 	searchParams?: {
+		email?: string;
 		return_to?: string;
 	};
 }) {
@@ -20,7 +21,10 @@ async function GuestLogin({
 			<h1 className="font-display text-3xl md:text-5xl mb-20">
 				Enter Passphrase
 			</h1>
-			<GuestLoginVerificationForm />
+			<GuestLoginVerificationForm
+				email={searchParams?.email}
+				return_to={searchParams?.return_to}
+			/>
 		</div>
 	);
 }
--- a/apps/site/src/app/(main)/guest-login/GuestLogin.tsx
+++ b/apps/site/src/app/(main)/guest-login/GuestLogin.tsx
@@ -6,14 +6,16 @@ import getUserIdentity from "@/lib/utils/getUserIdentity";
 async function GuestLogin({
 	searchParams,
 }: {
-	searchParams?: {
+	searchParams: {
 		email?: string;
 		return_to?: string;
 	};
 }) {
+	const { email, return_to } = searchParams;
+
 	const identity = await getUserIdentity();
 	if (identity.uid !== null) {
-		redirect(searchParams?.return_to ?? "/portal");
+		redirect(return_to ?? "/portal");
 	}
 
 	return (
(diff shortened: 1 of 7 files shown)


IanWearsHat (approved):

apps/site/src/app/(main)/guest-login/GuestLogin.tsx
```diff
@@ -7,6 +7,7 @@ async function GuestLogin({
 	searchParams,
 }: {
 	searchParams?: {
```
taesungh, line 9: Thought: I think Next.js will always provide `searchParams` to server components, so I don't think this needs to be marked as optional.

apps/site/src/app/(main)/guest-login/GuestLogin.tsx
```diff
@@ -20,7 +21,10 @@ async function GuestLogin({
 			<h1 className="font-display text-3xl md:text-5xl mb-20">
 				Enter Passphrase
 			</h1>
-			<GuestLoginVerificationForm />
+			<GuestLoginVerificationForm
+				email={searchParams?.email}
+				return_to={searchParams?.return_to}
```
taesungh, line 26: Suggestion: since we have more of these now, might be simpler to unspread them at the beginning of the function and shorten the specification also on line 16, e.g.
```tsx
const { email, return_to } = searchParams;
```

taesungh (dismissed):
Thanks for updating this.

apps/site/src/app/(main)/guest-login/GuestLogin.tsx
waalbert: I will go through and make these changes elsewhere too.

taesungh (approved):

Issue #513, opened 2024-12-21
Separate apps into different tabs


waalbert:
To help reviewers, we need to separate the different apps into separate tabs: Hackers, Mentors, and Volunteers. To minimize potential tampering with apps, we will also create separate admin roles for each of the apps. For example, `Hacker Reviewer`, `Mentor Reviewer`, and `Volunteer Reviewer` rather than having a general `Reviewer` role.

PR #533, opened 2024-12-31
Setup admin app review pages

### LINKED ISSUE #512: Update hacker applicants
Because the FRQ's on the hacker apps have changed this year, the admin dashboard models representing these questions need to be updated accordingly.

waalbert:
Closes #512 and sets up #513 

## Changes
- Updates title `IrvineHacks 2024` to `IrvineHacks 2025`
- Updates names of hacker app FRQs to properly show FRQs on hacker applicants review page
- Creates placeholder pages that shows all mentor and volunteer apps
- Adds these pages to the sidebar using [link groups](https://cloudscape.design/components/side-navigation/?tabId=playground&example=with-link-groups)
- Changes individual application page route (previously `/admin/applicants/<uid>`) to `/admin/applicants/hackers/<uid>`, `/admin/applicants/mentors/<uid>`, and `/admin/applicants/volunteers/<uid>` based on the app type
- Refactors component `Application` into 3 separate components: `HackerApplication`, `MentorApplication`, `Volunteer Application`
- Refactors component `ApplicationSection` into 3 separate components: `HackerApplicationSection`, `MentorApplicationSection`, `VolunteerApplicationSection`
- Refactors component `Applicant` to handle all 3 different app types with a prop `applicationType`
  - `Applicant` is used in `HackerApplicant`, `MentorApplicant`, and `VolunteerApplicant`
- Moves `Applicant`, `ApplicantActions`, `ApplicantOverview`, `ApplicationReviews`, and `HackerApplicantActions` to `/admin/applicants/components/` folder
- Updates `admin.py` by replacing route `/applicant/{uid}` with `/applicant/hacker/{uid}`, `/applicant/mentor/{uid}`, and `/applicant/volunteer/{uid}`
  - this decision was made so each route could be protected by user authorization roles (e.g. `Hacker Reviewer`, `Mentor Reviewer`) which will be done in a separate PR

## Testing

1. Submit multiple hacker, mentor, and volunteer apps
2. Verify that each individual application can be seen at route `/admin/applicants/<app_type>/<uid>`


apps/site/src/app/admin/layout/AdminSidebar.tsx
```diff
@@ -33,10 +33,37 @@ function AdminSidebar() {
...
+	}
+
+	// TODO: change access to role "Hacker Reviewer"
+	if (isApplicationManager(roles)) {
+		navigationItems.splice(2, 0, {
+			type: "link",
+			text: "Hacker Applications",
+			href: "/admin/applicants/hackers",
+		});
+	}
```
taesungh, line 61: Suggestion: not sure how tricky this will be with the more specific authorization, but given the path structure we previously decided, I think it would make more sense to nest these within Applicants, see example for [expandable link groups in Side navigation](https://cloudscape.design/components/side-navigation/?tabId=playground&example=with-expandable-link-groups).

apps/site/src/app/admin/applicants/[uid]/components/Application.tsx
Comment (taesungh) [inline apps/site/src/app/admin/applicants/[uid]/components/Application.tsx:1]: Question: currently this page will be at `/admin/applicants/<uid>`, do you think it would make more sense to group these by type of applicant, e.g. `/admin/applicants/hackers/<uid>` and `/admin/applicants/mentors/<uid>`? We'll probably also want to extract this into a common component and parameterize the sections to show.

Comment (waalbert) [reply apps/site/src/app/admin/applicants/[uid]/components/Application.tsx:1]: Yes, I think that'll make more sense. I was planning to do this in another PR, but since this PR isn't big, I'll refactor and move files around in this PR. Also, I'll extract the components into common components if possible.

apps/site/src/app/admin/applicants/hackers/components/HackerApplication.tsx
taesungh, line 1: Question: are the different applications to be displayed significantly differently? Unsure if it this component (and the sections component) could be parameterized and shared or if we really need distinct versions.

apps/site/src/lib/admin/useApplicant.ts
```diff
@@ -0,0 +1,135 @@
...
+export type Review = [string, Uid, Score];
+
+// The application responses submitted by an applicant
+export interface HackerApplicationData {
+	email: string;
+	pronouns: string[];
+	ethnicity: string;
+	is_18_older: boolean;
+	school: string;
+	education_level: string;
```
taesungh, line 15: Thought: since interfaces can be extended, we could combine some of the shared fields into a common interface to reduce the duplication.

apps/site/src/lib/admin/useApplicant.ts
```diff
@@ -0,0 +1,135 @@
...
+	mentor_team_help_saq4: string;
+	resume_share_to_sponsors: boolean;
+	other_questions: string | null;
+	resume_url: string;
+	submission_time: string;
+	reviews: Review[];
+}
+
+export interface VolunteerApplicationData {
+	kind: "mentor_prev_experience_saq1";
```
taesungh, line 51: Chore: stray value?

apps/site/src/lib/admin/useApplicant.ts
```diff
@@ -0,0 +1,135 @@
...
+		return null;
+	}
+	const res = await axios.get<Applicant>(api + uid);
+	return res.data;
+};
+
+function useApplicant(
+	uid: Uid,
+	application_type: "hacker" | "mentor" | "volunteer",
+) {
```
taesungh, line 112: Thought: with [TypeScript's Function Overloads](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads), I think we can correlate the return type with the value of `application_type`, possibly with a [generic interface](https://www.typescriptlang.org/docs/handbook/2/objects.html#generic-object-types) for the return type.

Chore: let's make `application_type` written in camelCase since this is in TypeScript.

apps/site/src/lib/admin/useApplicant.ts
```diff
@@ -0,0 +1,135 @@
...
+	const { data, error, isLoading, mutate } = useSWR<
+		Applicant | null,
+		unknown,
+		[string, Uid]
+	>([`/api/admin/applicant/${application_type}/`, uid], fetcher);
```
taesungh, line 117: Suggestion: it might make more sense to have the application type as a second item of the array and let the `fetcher` function combine the three values into the path.

apps/site/src/lib/admin/useApplicant.ts
taesungh, line 1: Guess we should have kept this file where it was rather than moving it in #526 😂.

taesungh (approved):
Main question on reducing duplication among the individual application views, but looking good otherwise.

apps/site/src/lib/admin/useApplicant.ts
waalbert: I tried doing this, but then there ended up being a TS overload error where `applicationType` couldn't match one of the overloaded signatures (`Argument of type '"hacker" | "mentor" | "volunteer"' is not assignable to parameter of type '"hacker"'`) in `Applicant` where `useApplicant` is called

apps/site/src/app/admin/applicants/hackers/components/HackerApplication.tsx
waalbert: I initially separated the components to deal with the different application interfaces separately rather than creating a single type of a union of the 3 application interfaces. I found that using union types creates more overhead and unnecessary complexity because of the need for type guarding. 

Separating the components will also allow for better isolation, in my opinion, in case the different questions need to be displayed differently (e.g. volunteer shifts) and would prevent `ApplicationResponse` from getting too large. Maybe the shared questions can be put into a shared component, but I think it'll be better if that's taken a look at in a different PR.

waalbert:
Will merge now for the sake of time. We can take a look at some of the comments later if needed

PR #535, opened 2025-01-01
Feat/organizer carousel

pandyrew:
Fun Organizer Carousel!

I put all the images and names and departments into sanity so I hope it works. I added localhost:3000 to the CORS origins in the sanity dashboard because i was running into CORS issues. LMK if there are better solutions. 

Also I added Images to be optimized by next in the next.config. Only from Sanity. 

Pictures that aren't added (as in if they weren't in the Zothacks headshots) are replaced with a placeholder: volunteer_sprite.png. Maybe we can pick a different one.

Screenshot 2024-12-31 at 5 18 30 PM


waalbert:
Regarding your comment about needing to add `localhost:3000` to CORS, the reason you have to do that is because you're using `next/image` and added `cdn.sanity.io` to `next.config.js` to optimize the images. We didn't have to worry about CORS in the past because we have our Sanity dataset exposed to the public, so we're able to use [Sanity Client](https://www.sanity.io/docs/js-client) along with regular `` tags. Our `Sanity Client` is configured to use Sanity's [API CDN](https://www.sanity.io/help/js-client-cdn-configuration), so using Next to optimize images might not be needed as I'm not sure if Next can provide meaningful performance improvements.

waalbert (approved):
Merging to unblock

PR #542, opened 2025-01-07
fix: remove applied before question from volunteer applications

### LINKED ISSUE #541: Remove hacker question on volunteer app
On the volunteer app, there is a question "Did you apply to be a hacker for IrvineHacks 2025?" that is redundant and can be removed because users can only submit one app. Not only does the frontend form need to reflect this change, the backend will need to be updated accordingly along with the admin portal.

jotalis:
Summary
- Removes everything on the frontend and backend related to the "applied_before" question for the volunteer application

Test Plan
- Create a user, submit a volunteer app, ensure data is transferred and saved properly

Closes #541

### DIFF
--- a/apps/api/src/models/ApplicationData.py
+++ b/apps/api/src/models/ApplicationData.py
@@ -97,7 +97,6 @@ class BaseVolunteerApplicationData(BaseModel):
     school: str
     education_level: str
     major: str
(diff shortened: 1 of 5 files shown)


waalbert (approved):
Thanks for the quick fix Jay!

PR #546, opened 2025-01-10
Changed Text for Login

### LINKED ISSUE #522: [Portal] Additional Login UI notes
From #418 

Regarding [this line](https://github.com/HackAtUCI/irvinehacks-site/blob/9aecc9830d936aee077e3442da424025607449ee/apps/site/src/lib/components/ValidatingForm/ValidatingForm.tsx#L43): 
> "log in" is being used as a verb, so it should be two words. Personally think having it say "Continue" made more sense since the login process has multiple steps.
---
Regarding [this line](https://github.com/HackAtUCI/irvinehacks-site/blob/9aecc9830d936aee077e3442da424025607449ee/apps/site/src/app/(main)/login/Login.tsx#L12), change text from "Login to Portal" to "Log in":
> the application page redirects here for new users, so it may be confusing to see it saying this goes to the Portal when users are trying to apply.

ofjuneskies:
Fixes #522

### DIFF
--- a/apps/site/src/app/(main)/login/Login.tsx
+++ b/apps/site/src/app/(main)/login/Login.tsx
@@ -18,9 +18,7 @@ async function Login({
 
 	return (
 		<div className="min-h-screen flex flex-col items-center justify-center">
-			<h1 className="font-display text-3xl md:text-5xl mb-20">
-				Log in
-			</h1>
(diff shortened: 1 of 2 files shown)


waalbert (approved):
Thanks for the quick work Jenny!

PR #547, opened 2025-01-11
Send apply reminder emails

waalbert:
## Changes

- Allow directors to send reminder to apply emails to user that have logged in but never submitted an app
  - Confirmation Modals and success and error Flashbars included for better UI experience
- New page called "Email Sender" that will be added to later for other email sending abilities
- Emails will not be sent more than once to each user
  -  Every time emails are sent out, the document with `_id: "apply_reminder"` under collection `emails` will be updated with every sender's uid and recipient uids

## Testing

1. Grant yourself the `Director` role
2. Ensure you can send emails to all users with a `last_login` field but no `roles` field


taesungh (approved):
Approving to unblock, will leave some more suggestions later. Trusting you were able to test this well and that the email message text is flexible enough regarding deadline/time perspective (although don't think this will be run multiple times this year). Perhaps next year, this could be run on schedule throughout the time applications are open.

waalbert:
> Approving to unblock, will leave some more suggestions later. Trusting you were able to test this well and that the email message text is flexible enough regarding deadline/time perspective (although don't think this will be run multiple times this year). Perhaps next year, this could be run on schedule throughout the time applications are open.

We might push back apps close, so I will update the dates on the email template if needed.

PR #549, opened 2025-01-12
added numbers of apps above accepted and waitlisted threshold

### LINKED ISSUE #537: Display number of apps above accept and waitlist threshold
The applications committee brought up that it would be easier for them if they could easily see how many applications are above the accept and waitlist threshold, and also the number of rejected apps.

ofjuneskies:
fixes #537

### DIFF
--- a/apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
+++ b/apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
@@ -1,9 +1,7 @@
 "use client";
 
 import { useRouter } from "next/navigation";
-
-import { useContext, useState } from "react";
-
+import { useContext, useState, useEffect } from "react";
 import Box from "@cloudscape-design/components/box";
 import Cards from "@cloudscape-design/components/cards";
 import Header from "@cloudscape-design/components/header";
@@ -24,9 +22,10 @@ import { isHackerReviewer } from "@/lib/admin/authorization";
 import HackerThresholdInputs from "../components/HackerThresholdInputs";
 import ApplicantReviewerIndicator from "../components/ApplicantReviewerIndicator";
 
+import useHackerThresholds from "@/lib/admin/useHackerThresholds";
+
 function HackerApplicants() {
 	const router = useRouter();
-
 	const { roles } = useContext(UserContext);
 
 	if (!isHackerReviewer(roles)) {
@@ -40,6 +39,13 @@ function HackerApplicants() {
 	const selectedStatusValues = selectedStatuses.map(({ value }) => value);
 	const selectedDecisionValues = selectedDecisions.map(({ value }) => value);
 
+	const [acceptedCount, setAcceptedCount] = useState(0);
+	const [waitlistedCount, setWaitlistedCount] = useState(0);
+
+	const { thresholds } = useHackerThresholds();
+	const acceptThreshold = thresholds?.accept;
+	const waitlistThreshold = thresholds?.waitlist;
+
 	const filteredApplicants = applicantList.filter(
 		(applicant) =>
 			(selectedStatuses.length === 0 ||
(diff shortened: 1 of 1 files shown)


apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
```diff
@@ -120,8 +143,21 @@ function HackerApplicants() {
 			}
 			empty={emptyContent}
 			header={
-				<Header counter={counter} actions={<HackerThresholdInputs />}>
-					Hacker Applicants
+				<Header actions={<HackerThresholdInputs />}>
+					<div style={{ display: "flex", alignItems: "baseline", gap: "8px" }}>
+						<div>Hacker Applicants</div>
+						<div>({counter})</div>
```
waalbert, line 149: Suggestion: `Header` can take `counter` as a prop, so the `<div>({counter})</div>` isn't needed.

apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
```diff
@@ -120,8 +143,21 @@ function HackerApplicants() {
...
+					<div
+						style={{ fontSize: "0.875rem", color: "#5f6b7a", marginTop: "4px" }}
+					>
+						{acceptedCount} applicants above acceptance threshold
+					</div>
+					<div
+						style={{ fontSize: "0.875rem", color: "#5f6b7a", marginTop: "4px" }}
+					>
+						{waitlistedCount} applicants above waitlist threshold
+					</div>
```
waalbert, line 160: Can you also add the number of rejects applicants too?

apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
```diff
@@ -48,12 +54,30 @@ function HackerApplicants() {
...
+	}, [applicantList, acceptThreshold, waitlistThreshold]);
+
 	const items = filteredApplicants;
 
 	const counter =
 		selectedStatuses.length > 0 || selectedDecisions.length > 0
-			? `(${items.length}/${applicantList.length})`
-			: `(${applicantList.length})`;
+			? `${items.length}/${applicantList.length}`
+			: `${applicantList.length}`;
```
waalbert, line 81: Relating to my other comment about the `counter`, this change isn't needed if you leave the `counter` as it was written before your commits

apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
```diff
@@ -48,12 +54,30 @@ function HackerApplicants() {
...
+	useEffect(() => {
+		const accepted = acceptThreshold ? acceptThreshold : 0;
+		const waitlisted = waitlistThreshold ? waitlistThreshold : 0;
+		const waitlistedCount = applicantList.filter(
+			(applicant) =>
+				applicant.avg_score >= waitlisted && applicant.avg_score < accepted,
+		).length;
+		setWaitlistedCount(waitlistedCount);
+	}, [applicantList, acceptThreshold, waitlistThreshold]);
+
```
waalbert, line 74: Suggestion: these can be combined into one `useEffect` to call `setAcceptedCount` and `setWaitlistedCount`.

apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
waalbert: Since the wording "applicants above ... threshold" is confusing because it implies that "above waitlist threshold" includes both waitlisted and accepted applicants, something like "applicants with accepted status" or whatever you think is good would make more sense, in my opinion.

waalbert (changes requested):
Awesome start Jenny! Here are a few comments

waalbert (approved):
Thanks for the work Jenny! I will merge this in after apps close

Issue #531, opened 2024-12-25
Prevent reviewers from submitting invalid scores/thresholds


waalbert:
Currently, there isn't a check to see if a reviewer submits an invalid score or threshold (i.e. negative numbers or numbers greater than the max possible). Adding this check would help minimize accidental mistakes in reviewing apps.

PR #550, opened 2025-01-13
Threshold and Invalid Input Fix

### LINKED ISSUE #536: Restrict threshold modification to directors
Currently, anyone with the authorization role of `Reviewer` can change thresholds, but for improved security, we should restrict this ability to only directors.

Bl20052005:
Closes #536 and #531

### DIFF
--- a/apps/api/src/routers/admin.py
+++ b/apps/api/src/routers/admin.py
@@ -302,6 +302,33 @@ async def set_hacker_score_thresholds(
     Sets accepted and waitlisted score thresholds.
     Any score under waitlisted is considered rejected.
     """
+
+    record = None
+
+    try:
+        record = await mongodb_handler.retrieve_one(
+            Collection.SETTINGS,
+            {"_id": "hacker_score_thresholds"},
+        )
+    except RuntimeError:
+        log.error("%s could not retrieve thresholds")
+
+    if accept != -1 and record is not None:
+        record["accept"] = accept
+    if waitlist != -1 and record is not None:
+        record["waitlist"] = waitlist
+
+    if (
+        accept < -1
+        or accept > 10
+        or waitlist < -1
+        or waitlist > 10
+        or (accept != -1 and waitlist != -1 and waitlist > accept)
+        or (record and record["waitlist"] > record["accept"])
+    ):
+        log.error("Invalid threshold score submitted.")
+        raise HTTPException(status.HTTP_400_BAD_REQUEST)
+
     log.info("%s changed thresholds: Accept-%f | Waitlist-%f", user, accept, waitlist)
 
     # negative numbers should not be received, but -1 in this case
--- a/apps/api/src/routers/admin.py
+++ b/apps/api/src/routers/admin.py
@@ -229,7 +229,7 @@ async def submit_review(
 ) -> None:
     """Submit a review decision from the reviewer for the given hacker applicant."""
 
-    if applicant_review.score < 0 or applicant_review.score > 10:
+    if applicant_review.score < -2 or applicant_review.score > 100:
         log.error("Invalid review score submitted.")
         raise HTTPException(status.HTTP_400_BAD_REQUEST)
 
@@ -248,6 +248,11 @@ async def submit_review(
         raise HTTPException(status.HTTP_500_INTERNAL_SERVER_ERROR)
 
     if Role.HACKER in applicant_record["roles"]:
+
+        if applicant_review.score < 0 or applicant_review.score > 10:
+            log.error("Invalid review score submitted.")
+            raise HTTPException(status.HTTP_400_BAD_REQUEST)
+
         unique_reviewers = applicant_review_processor.get_unique_reviewers(
             applicant_record
         )
--- a/apps/api/src/routers/admin.py
+++ b/apps/api/src/routers/admin.py
@@ -303,20 +303,12 @@ async def set_hacker_score_thresholds(
     Any score under waitlisted is considered rejected.
     """
 
-    record = None
+    thresholds: Optional[dict[str, float]] = await _retrieve_thresholds()
 
-    try:
-        record = await mongodb_handler.retrieve_one(
-            Collection.SETTINGS,
-            {"_id": "hacker_score_thresholds"},
-        )
-    except RuntimeError:
-        log.error("%s could not retrieve thresholds")
-
-    if accept != -1 and record is not None:
-        record["accept"] = accept
-    if waitlist != -1 and record is not None:
-        record["waitlist"] = waitlist
+    if accept != -1 and thresholds is not None:
+        thresholds["accept"] = accept
+    if waitlist != -1 and thresholds is not None:
+        thresholds["waitlist"] = waitlist
 
     if (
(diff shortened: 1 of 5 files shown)


apps/api/src/routers/admin.py
```diff
@@ -228,6 +228,11 @@ async def submit_review(
     reviewer: User = Depends(require_reviewer),
 ) -> None:
     """Submit a review decision from the reviewer for the given hacker applicant."""
+
+    if applicant_review.score < 0 or applicant_review.score > 10:
+        log.error("Invalid review score submitted.")
+        raise HTTPException(status.HTTP_400_BAD_REQUEST)
+
```
waalbert, line 235: Issue: since mentor and volunteer app reviews use the set values of 0 (reject), -2 (waitlist), and 100 (accept) for the application score when a review is submitted (see [ApplicantActions.tsx](https://github.com/HackAtUCI/irvinehacks-site/blob/1b6028cb1c490d4d9c89619d2b9498b66a0a2479/apps/site/src/app/admin/applicants/components/ApplicantActions.tsx)), you would need to allow these scores as well.

waalbert (changes requested):
Awesome start Bo! 

Can you also add frontend form validation for submitting a review score and also frontend form and backend validation for setting the thresholds?

Note for your reference: we will be moving the setting thresholds form to the director's dashboard, so since this PR is small, can you also remove `HackerThresholdInputs` from `HackerApplicants`? We will add `HackerThresholdInputs` to the director's dashboard in another PR.

apps/api/src/routers/admin.py
```diff
@@ -243,6 +248,11 @@ async def submit_review(
         raise HTTPException(status.HTTP_500_INTERNAL_SERVER_ERROR)
 
     if Role.HACKER in applicant_record["roles"]:
+
+        if applicant_review.score < 0 or applicant_review.score > 10:
+            log.error("Invalid review score submitted.")
+            raise HTTPException(status.HTTP_400_BAD_REQUEST)
+
```
waalbert, line 255: Can you add a similar conditional to the API endpoint `/set-thresholds` too?

waalbert (changes requested):
Also, make sure to add form validation in `HackerThresholdsInput` for valid threshold values: 0 to 10 and make sure that the `accept` threshold value must be greater than or equal to the `waitlist` threshold value . 

Note: when the input field is empty, the value of -1 is sent to `/api/admin/set-thresholds` for when the accept or waitlist threshold value should stay the same.

apps/site/src/app/admin/applicants/components/HackerThresholdInputs.tsx
```diff
@@ -19,26 +19,54 @@ function HackerThresholdInputs() {
...
+		const sentAcceptValue = parseFloat(acceptValue);
+		const sentWaitlistValue = parseFloat(waitlistValue);
+
+		if (sentWaitlistValue < -1 || sentWaitlistValue > 10) return false;
+
+		if (sentAcceptValue < sentWaitlistValue) return false;
+
+		return true;
+	};
+
```
waalbert, line 69: Issue: in the case that one of `acceptValue` or `waitlistValue` is the empty string, then this doesn't check, for example, if the `waitlistValue` is 6, `acceptValue` is "" and `thresholds.accept` is 5, then there should be an error in this form instead of attempting to submit the POST request. 

The new threshold values will need to be compared against the old threshold values in `thresholds`.

apps/site/src/app/admin/applicants/hackers/HackerApplicants.tsx
```diff
@@ -120,9 +122,11 @@ function HackerApplicants() {
...
 			empty={emptyContent}
 			header={
-				<Header counter={counter} actions={<HackerThresholdInputs />}>
-					Hacker Applicants
-				</Header>
+				isUserDirector ? (
+					<Header counter={counter} actions={<HackerThresholdInputs />}>
+						Hacker Applicants
+					</Header>
+				) : null
```
waalbert, line 129: Can you show the `Header` with "Hacker Applicants" for everyone and conditionally render just `<HackerThresholdInputs />` for directors? At the moment, the title "Hacker Applicants" along with the number of applicants (`counter`) are hidden from non-directors.

waalbert (changes requested):

apps/site/src/app/admin/applicants/components/HackerThresholdInputs.tsx
```diff
@@ -19,26 +19,90 @@ function HackerThresholdInputs() {
...
+			return true;
+		}
+
+		const sentWaitlistValue = waitlistValue
+			? parseFloat(waitlistValue)
+			: thresholds?.waitlist;
+
+		if (sentAcceptValue < -1 || sentAcceptValue > 10) return false;
+
+		if (sentAcceptValue < sentWaitlistValue) return false;
```
waalbert, line 70: Nitpick: these can be combined into one conditional block like you have in lines 54-59.

apps/site/src/app/admin/applicants/components/HackerThresholdInputs.tsx
```diff
@@ -19,26 +19,90 @@ function HackerThresholdInputs() {
...
+			return true;
+		}
+
+		const sentAcceptValue = acceptValue
+			? parseFloat(acceptValue)
+			: thresholds?.accept;
+
+		if (sentWaitlistValue < -1 || sentWaitlistValue > 10) return false;
+
+		if (sentAcceptValue < sentWaitlistValue) return false;
```
waalbert, line 101: Nitpick: these can be combined into one conditional block like you have in lines 85-90.

apps/site/src/app/admin/applicants/components/HackerThresholdInputs.tsx
```diff
@@ -19,26 +19,90 @@ function HackerThresholdInputs() {
...
+				sentWaitlistValue > 10
+			)
+				return false;
+
+			return true;
+		}
+
+		const sentAcceptValue = acceptValue
+			? parseFloat(acceptValue)
+			: thresholds?.accept;
```
waalbert, line 97: Nitpick: since you've already checked to see if `thresholds` is undefined in line 80, you don't need to use the optional chaining operator here.

apps/site/src/app/admin/applicants/components/HackerThresholdInputs.tsx
```diff
@@ -19,26 +19,90 @@ function HackerThresholdInputs() {
...
+				sentAcceptValue > 10
+			)
+				return false;
+
+			return true;
+		}
+
+		const sentWaitlistValue = waitlistValue
+			? parseFloat(waitlistValue)
+			: thresholds?.waitlist;
```
waalbert, line 66: Nitpick: since you've already checked to see if `thresholds` is undefined in line 80, you don't need to use the optional chaining operator here.

waalbert (changes requested):
Almost done! A few more nitpicks

waalbert (approved):
Thanks for the work Bo!

PR #551, opened 2025-01-13
simplified ApplicationFlow's searcParam to only have prefaceAccepted

### LINKED ISSUE #480: Simplify accessing `prefaceAccepted` parameter in `ApplicationFlow`
Mentioned in #440: the [`ApplicationFlowProps` includes the entire `searchParams`](https://github.com/HackAtUCI/irvinehacks-site/blob/00975fad7cea19dbd515acdf3aa957bfd0be2366/apps/site/src/lib/components/forms/shared/ApplicationFlow.tsx#L14-L20) object when the component cares only about the `prefaceAccepted` parameter, so it would be simpler to have that as the prop and have the consumers select just that parameter to provide.

cyriljoby:
resolves #480

### DIFF
--- a/apps/site/src/app/(main)/apply/Hacker.tsx
+++ b/apps/site/src/app/(main)/apply/Hacker.tsx
@@ -12,7 +12,7 @@ export default async function Hacker({
 }) {
 	return (
 		<ApplicationFlow
-			searchParams={searchParams}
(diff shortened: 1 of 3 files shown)


IanWearsHat (approved):
LGTM

Issue #317, opened 2024-01-21
[API] Consider new statuses in waiver handler


taesungh:
Following from #276/#280, we need to update our `process_waiver_completion` function to make sure a user who signs the waiver again after having an `ATTENDING` status is not affected.

waalbert:
Closing as completed in #561

Issue #523, opened 2024-12-23
Use streaming for dynamic page content


taesungh:
All pages of the main site need to be dynamically rendered because [the Navbar depends on `getUserIdentity`](https://github.com/HackAtUCI/irvinehacks-site/blob/3752272e3bba64e62ded6441b3a79d095e11ed2d/apps/site/src/lib/components/Navbar/NavbarParent.tsx#L5) to show specific navigation items based on the user's status. While dynamic rendering itself is not a huge problem, the initial loading of the site can feel really slow if there is a cold start for the serverless function that's running the API. I noticed that it can take over 6 seconds just to get a page response:
```shell
16:25:28.351589 > GET / HTTP/2
16:25:28.351589 > Host: irvinehacks.com
16:25:28.351589 > User-Agent: curl/8.7.1
16:25:28.351589 > Accept: */*
16:25:28.351589 > 
16:25:28.351642 * Request completely sent off
16:25:34.527435 < HTTP/2 200 
```

This is a very poor score for [Time To First Byte](https://web.dev/articles/ttfb) which could be causing negative user sentiment. This can be improved by using [Streaming with Suspense](https://nextjs.org/docs/14/app/building-your-application/routing/loading-ui-and-streaming#streaming-with-suspense) for the navigation items that do need to by dynamically rendered. This will allow all parts of the UI not depending on the API to be rendered and streamed first so the primary page content can be visible faster.

One alternative is reverting to a [client-side fetch](https://nextjs.org/docs/app/building-your-application/data-fetching/fetching#fetching-data-on-the-client) for the user identity in the navigation, but fetching data on the server-side is just a lot more convenient (no need to deal with `useEffect`, SWR, or React Query).

Besides the Navbar, the Apply, Login, Guest Login, and Portal pages also rely on `getUserIdentity`. Most of the time, these should be less impacted since the API would likely be in a warm state, but it's possible for somebody to directly open such pages and encounter a cold start. However, in many cases, the user identity determines a [redirect](https://nextjs.org/docs/app/api-reference/functions/redirect), so it may semantically make more sense to not use streaming to preserve the HTTP status codes (although client-side redirects aren't terrible).

Next.js 14 also introduced [Partial Prerendering](https://nextjs.org/docs/app/building-your-application/rendering/partial-prerendering), but this is still considered experimental even in Next.js 15. PPR would allow mixing static and dynamic components so the non-dynamic parts can be cached and served from a CDN.

#### Main tasks
[...]

PR #561, opened 2025-01-15
Fixing waiver handler and signing process

### LINKED ISSUE #273: [Portal] Waiver signing process lacks clarity
A small but noticeable proportion of accepted applicants signed the waiver more than once, likely thinking the submission was not received since the DocuSign Connect webhook can have a slight delay. Additionally, several participants have signed the waiver but not confirmed their attendance.
- Open the waiver signing form in a new tab
- Include a message that it may take a minute or so for the waiver completion to be recognized
- Use a client-side hook with SWR to refetch the user's status when refocusing the window/tab with the Portal

Bl20052005:
Closes #273 and #317


waalbert (approved):
Looking awesome Bo!

apps/site/src/app/(main)/portal/@applicant/ApplicantPortal.tsx
```diff
@@ -25,15 +27,15 @@ export const enum PortalStatus {
...
+	const identity = useUserIdentity();
+	const status = identity?.status;
 
 	if (status === null) {
 		redirect("/#apply");
 	}
 
-	const roleToDisplay = identity.roles.find((role) =>
+	const roleToDisplay = identity?.roles.find((role) =>
 		rolesArray.includes(role),
```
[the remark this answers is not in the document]
taesungh: Oh I see, `identity`is initially undefined. In that case, it would be good to display a loading state or perhaps provide the initial value from the server (subject to downsides noted in #523).

apps/api/src/utils/waiver_handler.py
```diff
@@ -38,6 +38,9 @@ async def process_waiver_completion(uid: str, email: EmailStr) -> None:
                 f"User {uid} attempted to sign waiver but already signed it previously."
             )
             return
+        elif applicant_record.status == Status.ATTENDING:
+            log.warning(f"User {uid} has already signed the waiver and is attending.")
+            return
```
taesungh, line 43: Sorry I missed this before, but I think this condition could be combined with the one before on line 36. For the future, might be worth considering somehow making `Status` a partially ordered set.

Bl20052005: It definitely could, I just thought that the warning message here is different enough to the previous one

PR #563, opened 2025-01-15
Mentor and Volunteer decision emails

### LINKED ISSUE #559: Add mentor and volunteer decision emails
Right now, the site only sends out decision emails specific for hackers. However, mentor and volunteer emails have different content, so our backend will need to be updated accordingly to handle the different application types.

waalbert:
## Changes

- Participants are processed in batches based on their `application_type` (from `roles`)
- Adds SendGrid templates for `MENTOR_ACCEPTED_EMAIL`, `MENTOR_REJECTED_EMAIL`, `VOLUNTEER_ACCEPTED_EMAIL`, and `VOLUNTEER_REJECTED_EMAIL`
- Renames SendGrid templates for hackers to `HACKER_ACCEPTED_EMAIL`, `HACKER_WAITLISTED_EMAIL`, and `HACKER_REJECTED_EMAIL`

Closes #559


apps/api/src/routers/admin.py
```diff
@@ -558,14 +558,20 @@ async def subevent_checkin(
...
-        group = [record for record in records if record["decision"] == decision]
-        if not group:
-            continue
-        await asyncio.gather(
-            *(_process_batch(batch, decision) for batch in batched(group, 100))
-        )
+async def _process_records_in_batches(
+    records: list[dict[str, object]], application_types: list[Role]
+) -> None:
+    for application_type in application_types:
```
taesungh, line 564: Suggestion: rather than iterating here, the consumer could just call this function multiple times.

apps/api/src/utils/email_handler.py
```diff
@@ -54,15 +65,17 @@ async def send_guest_login_email(email: EmailStr, passphrase: str) -> None:
 
 
 async def send_decision_email(
-    applicant_batch: Iterable[tuple[str, EmailStr]], decision: Decision
+    applicant_batch: Iterable[tuple[str, EmailStr]],
+    decision: Decision,
+    application_type: Role,
```
taesungh, line 70: Suggestion: might help to constrain this type to require one of the three roles as a Literal

PR #578, opened 2025-01-17
Moved director only routes to director.py

### LINKED ISSUE #572: Move director only routes from admin to director
Before the new director routes were created for IH 25 (`/api/director/organizers` and `/api/director/apply-reminder`), all director only routes are stored in `/admin`. But due to the `admin` route getting to large, we should move the existing director only routes to `/director` (e.g. `/set-thresholds`, `/release/hackers`, etc.).

IanWearsHat:
Closes #572 

- Fixed test_admin and test_director accordingly

### DIFF
--- a/apps/api/src/routers/admin.py
+++ b/apps/api/src/routers/admin.py
@@ -1,10 +1,9 @@
-import asyncio
 from datetime import date, datetime
 from logging import getLogger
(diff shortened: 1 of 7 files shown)


waalbert (approved):
Looking good Ian!

PR #599, opened 2025-01-24
Combined get_hackers and get_non_hackers into one function

### LINKED ISSUE #595: Remove non-attending mentors and volunteers from participants table
Because mentors and volunteers are currently being fetched at `participant_manager.get_non_hackers`, all mentors and volunteers with any status including `VOID` and `REJECTED` get fetched, which is not what we want because we don't want them showing up during check-in. Also, the backend endpoint `/api/admin/participants` can be simplified by combining `participant_manager.get_hackers` and `participant_manager.get_non_hackers` into one, especially now with mentors and volunteers needing to be fetched separately from other non-hackers. 

- [ ] Change query for fetching only mentors and volunteers without `VOID` and `REJECTED`
- [ ] Simplify `/api/admin/participants` endpoint

### LINKED ISSUE #598: Remove duplicate attending mentors and volunteers
Mentors and volunteers that have a status of `ATTENDING` will be shown up twice on the participants table because they are getting fetched twice by both `participant_manager.get_hackers` and `participant_manager.get_non_hackers`. Following from #595, these two functions can be combined into one function.

IanWearsHat:
Closes #595 and closes #598

### DIFF
--- a/apps/api/src/admin/participant_manager.py
+++ b/apps/api/src/admin/participant_manager.py
@@ -42,13 +42,12 @@ PARTICIPANT_FIELDS = [
 ]
 
 
-async def get_hackers() -> list[Participant]:
+async def get_participants() -> list[Participant]:
     """Fetch all applicants who have a status of ATTENDING, WAIVER_SIGNED, CONFIRMED,
     or WAITLISTED."""
     records: list[dict[str, Any]] = await mongodb_handler.retrieve(
         Collection.USERS,
         {
-            "roles": Role.APPLICANT,
             "status": {
                 "$in": [
                     Status.ATTENDING,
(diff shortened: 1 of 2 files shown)


apps/api/src/admin/participant_manager.py
```diff
@@ -42,13 +42,12 @@ class Participant(UserRecord):
...
+async def get_participants() -> list[Participant]:
     """Fetch all applicants who have a status of ATTENDING, WAIVER_SIGNED, CONFIRMED,
     or WAITLISTED."""
     records: list[dict[str, Any]] = await mongodb_handler.retrieve(
         Collection.USERS,
         {
-            "roles": Role.APPLICANT,
             "status": {
                 "$in": [
                     Status.ATTENDING,
```
waalbert, line 53: Issue: all non-hackers, non-mentors, and non-volunteer participants won't have the status of `ATTENDING` until they're first checked in, so you would have to fetch all of these participants regardless of status. See `participant_manager.confirm_attendance_non_hacker`.

waalbert (approved):
Awesome thanks Ian!

PR #602, opened 2025-01-25
Fixed bug where empty doc with only status breaks get_applicants function

IanWearsHat:


waalbert (approved):
Thanks for the quick fix Ian!

Issue #623, opened 2025-09-12
Update resume_handler


IanWearsHat:
Currently, the folder IDs are stored in `resume_handler` globally. `gdrive_handler` shouldn't need to be touched.

We want to upload dynamically based on the hackathon type with some context logic similar to #621

PR #626, opened 2025-09-13
Created middleware and changed mongodb_handler to handle multiple databases

### LINKED ISSUE #619: App needs to handle different databases
The `mongodb_handler.py` currently:
1. Sets the database to use globally, so we can't change it based on any arguments
2. Only uses the databases "irvinehacks" and "irvinehacks-prod"

We want each request to determine which database to use (IH or ZH).

## Next.js side
We don't want to manually specify the db for every request.
We can do this by attaching a header to each request through middleware:
```typescript
import { NextResponse } from "next/server";
import type { NextRequest } from "next/server";

export function middleware(request: NextRequest) {
	const requestHeaders = new Headers(request.headers);
	requestHeaders.set("X-Hackathon-Name", "irvinehacks");

	return NextResponse.next({
		request: {
			headers: requestHeaders,
		},
	});
}

export const config = {
	matcher: ["/api/:path*"],
};
```

## FastAPI side
The plan:
1. For every request, we take the db header, verify it, then set that db name to a **global context**.
2. We then run the function for that endpoint (ex. fetching all applicants, uploading application data)
3. If we run a function involving Mongo, we set the database name based on the context.

To achieve each step:
1. Use FastAPI middleware to set a context. Something like:
```python
# mongodb_context.py
import contextvars

db_name_ctx = contextvars.ContextVar("db_name", default=None)
```
```python
from fastapi import Request
from mongodb_context import db_name_ctx

@app.middleware("http")
async def set_db_name_from_header(request: Request, call_next):
    db_name_ctx.set(request.headers.get("X-Database-Name"))
    return await call_next(request)
```
2. The function runs from `call_next`
3. In our `mongodb_handler.py`, for each of our functions, we have something like:
```python
MONGODB_CLIENT = AsyncIOMotorClient("mongodb://localhost:27017")

def get_db() -> AsyncIOMotorDatabase:
    db_name = db_name_ctx.get()
    if not db_name:
        raise RuntimeError("No database selected in request context")
    return MONGODB_CLIENT[db_name].with_options(
        codec_options=CodecOptions(tz_aware=True)
    )

async def insert(collection_name: str, data: dict, , db = Depends(get_db)):):
    collection = db[collection_name]
    result = await collection.insert_one(data)
    return result
```

### LINKED ISSUE #620: Attach headers to outgoing requests in Next.js
It should match our existing API call pattern: /api/something/something

It should attach a header to our request that specifies the name of the databse.

### LINKED ISSUE #621: Create context and middleware for Mongo in FastAPI
It should match and verify the header provided in #620. 

The context should then be given to each Mongo function in `mongodb_handler.py`

Test that irvinehacks is stable first.

IanWearsHat:
Closes #619 
Closes #620 
Closes #621 

## Changes
- FastAPI
  - `app.py`
    - Added middleware to FastAPI app
    - Sets context from `X-Hackathon-Name` header
  - `mongodb_handler.py`
    - Created `get_database` function to get database based on context value
    - For each function, called `get_database` before performing operations
  - `hackathon_context`
    - Only purpose is to provide context value
    - In the future, use this context value for things like #623 
- NextJS
  - `api.ts`
    - This is a server-side function, so header needs to be set here separately.
  - `middleware.ts`
    - This is for every client-side request that starts with `/api/`, so header needs to be set here separately.

## Testing
Add print statement to FastAPI middleware like:
```python
@app.middleware("http")
async def set_hackathon_name_context_from_header(request: Request, call_next):
    hackathon_name_ctx.set(request.headers.get("X-Hackathon-Name"))
    print("middleware works")
    print("hackathon_name_ctx.get())
    return await call_next(request)
```

- Server-side api call
  - Loading the site should run `getUserIdentity`, which should hit the middleware and print
- Client-side api call
 Add some way to call axios in a client component, which should hit the middleware and print. I created a button in `CharacterBox.tsx`  for this:
```typescript
<button type="button" onClick={() => {
	axios.get("/api/admin/set-hackathon-name");
}}>
	Set Hackathon Name
</button>
```

### DIFF
--- a/apps/api/.gitignore
+++ b/apps/api/.gitignore
@@ -2,3 +2,4 @@ __pycache__
 *.key
 .coverage
 lib
+.docker
--- a/apps/api/src/app.py
+++ b/apps/api/src/app.py
@@ -4,6 +4,7 @@ import os
 from fastapi import FastAPI, Request
 
 from routers import admin, director, guest, saml, user
+from utils.hackathon_context import hackathon_name_ctx
 
 logging.basicConfig(level=logging.INFO)
 
@@ -32,9 +33,6 @@ async def root() -> dict[str, str]:
 
 
 @app.middleware("http")
-async def set_db_name_from_header(request: Request, call_next):
-    # db_name_ctx.set(request.headers.get("X-Database-Name"))
-    print("Setting db name from header")
-    print(request.headers.get("X-Hackathon-Name"))
-    print("--------------------------------")
+async def set_hackathon_name_context_from_header(request: Request, call_next):
+    hackathon_name_ctx.set(request.headers.get("X-Hackathon-Name"))
     return await call_next(request)
--- a/apps/api/src/app.py
+++ b/apps/api/src/app.py
@@ -1,7 +1,8 @@
 import logging
 import os
+from typing import Awaitable, Callable
 
-from fastapi import FastAPI, Request
+from fastapi import FastAPI, Request, Response
 
 from routers import admin, director, guest, saml, user
 from utils.hackathon_context import hackathon_name_ctx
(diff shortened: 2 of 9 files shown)


YKawesome (approved):
Works on my end!

PR #683, opened 2025-10-24
feat: add resume review status filter to admin applicants page

### LINKED ISSUE #673: [Admin] Add filter for reviewed resumes
Following #672, we should add a filter for applicants who had their resumes reviewed.

VishokL:
Added a dropdown filter to the applicants page in the admin dashboard that filters applicants by their resume review status. This is calculated on the backend by checking if any dictionary in the "review_breakdown" of the "application_data" has a "resume" entry.

Closes #673

### DIFF
--- a/apps/api/src/admin/applicant_review_processor.py
+++ b/apps/api/src/admin/applicant_review_processor.py
@@ -90,9 +90,7 @@ def _include_avg_score(applicant_record: dict[str, Any]) -> None:
 
 
 def _include_resume_reviewed(applicant_record: dict[str, Any]) -> None:
-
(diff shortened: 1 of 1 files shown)


YKawesome:
I think the only two issues from the tests were (in `test_admin.py`):
- `test_hacker_applicants_returns_correct_applicants`: 150 (int instead of double I guess?)
- `test_can_retrieve_applicants`: 'resume_reviewed' isn't added as part of the expected records

(Unless I'm sorely mistaken and the values were incorrect, but I saw the 150 and I assume you just didn't add that resume_reviewed flag in the test since you added it).

PR #695, opened 2025-11-08
Feature/reviewer notes

RichieRish05:
Enhanced backend endpoint to support submission of detailed reviews with reviewer notes. Updated frontend to include a notes field and ensure it’s properly passed through the updated API. Added a new component to display previous reviewers’ notes and allow users to add their own.

### DIFF
--- a/apps/api/tests/test_admin.py
+++ b/apps/api/tests/test_admin.py
@@ -16,7 +16,6 @@ from routers.admin import (
     GlobalScores,
     DeleteNotesRequest,
     ZotHacksHackerDetailedScores,
-
 )
 from services.mongodb_handler import Collection
 from services.sendgrid_handler import Template


apps/site/src/app/admin/applicants/components/DetailedScoreApplicant.tsx
```diff
@@ -70,8 +69,9 @@ function DetailedScoreApplicant({
 								applicant={applicant._id}
 								reviews={application_data.reviews}
 								scores={scores}
-								submitDetailedReview={(Uid, scores) =>
-									submitDetailedReview(Uid, scores).then(() => {
+								notes={notes}
+								submitDetailedReview={(Uid, scores, notes) =>
+									submitDetailedReview(Uid, scores, notes).then(() => {
```
IanWearsHat: Issue: When I leave a comment and submit a review, the comment stays as expected. However, say I score something else and then submit a review. My comment from before is added again, leading to duplicate comments. 

The easiest solution would be to reset the `notes` state on this line and in the other instance of `ZotHacksHackerApplicantActions`. Actually, it would be best to extract those anonymous functions to a named function.

apps/site/src/app/admin/applicants/components/ReviewerNotes.tsx
```diff
@@ -0,0 +1,110 @@
...
+}: ReviewerNotesProps) {
+	/*
+	Filter out reviews with null notes, keeping the original index with the review.
+	This is because the backend endpoint needs the original index of the review in 
+	the mongodb document to delete the correct note.
+	*/
+	const [reviewsWithNotes, setReviewsWithNotes] = useState<
+		ReviewWithOriginalIdx[]
+	>(
+		(reviews ?? [])
```
IanWearsHat: Issue: When submitting a new note, the UI doesn't update with my new note. It's because this line doesn't actually change state when `reviews` changes. 

We can replace this with something like this. The notes will still update when reviews updates and we avoid using `useEffect`:
```typescript
const reviewsWithNotes: ReviewWithOriginalIdx[] = (reviews ?? [])
		.map((review, originalIdx) => ({ review, originalIdx }))
		.filter(({ review }) => review[3] !== null)
```

apps/site/src/lib/admin/useApplicant.ts
```diff
@@ -156,10 +169,29 @@ function useApplicant(
 		error,
 		submitReview,
 		submitDetailedReview,
+		deleteNotes,
 	};
+
+	async function deleteNotes(uid: Uid, reviewIndex: number) {
+		await axios.delete("/api/admin/delete-notes", {
+			data: {
```
IanWearsHat: Question: Is there a reason this `data` key has to be here? i.e. Why can't the parameter just be

```
{
	applicant: uid,
	review_index: reviewIndex,
}
```

RichieRish05: axios.delete() behaves a little differently compared to axios.post, axios.put, and axios.get. The url parameters must be wrapped in a data object. I've attached a Stack Overflow thread highlighting this. 

https://stackoverflow.com/questions/51069552/axios-delete-request-with-request-body-and-headers

PR #729, opened 2026-01-11
Setup/footer

### LINKED ISSUE #701: Footer
As per Figma, we need to update the font and colors for the footer. For mobile, we can make the elements stack vertically like Logos -> "Made with <3" -> "Learn more". I believe this is how the previous site behaved for mobile. 

Note: The footer background is transparent with a color.

Last year's site mobile:

Image

Figma:
Image

collinchenn:
Created the 2026 IrvineHacks Footer 
Added pink SVGs for Hack socials
Removed inline CSS into file for structure

Closes #701

### DIFF
--- a/apps/site/src/lib/components/Footer/Footer.module.scss
+++ b/apps/site/src/lib/components/Footer/Footer.module.scss
@@ -4,7 +4,7 @@
 	display: flex;
 	flex-direction: column;
 	gap: 10px;
-}	
+}
 
 .hackLogo {
 	display: flex;
@@ -12,10 +12,10 @@
 }
 
 @media (max-width: 600px) {
-  .footerText {
-    white-space: normal;
-    text-align: center;
-  }
+	.footerText {
+		white-space: normal;
+		text-align: center;
+	}
 }
 
 .bullet {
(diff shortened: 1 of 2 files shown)


apps/site/src/lib/components/Footer/Footer.tsx
```diff
@@ -1,18 +1,15 @@
...
-import redHeart from "@/assets/icons/red-heart.svg";
+import mail from "@/assets/logos/mail-pink.svg";
+import discord from "@/assets/logos/discord-pink.svg";
+import instagram from "@/assets/logos/instagram-pink.svg";
+import tiktok from "@/assets/logos/tiktok-pink.svg";
+import linkedin from "@/assets/logos/linkedin-pink.svg"
+import facebook from "@/assets/logos/facebook-pink.svg"
 import hack from "@/assets/logos/hack.svg";
+
+import Image from "next/image";
```
IanWearsHat: I didn't tell you this earlier, but going forward, let's try to order our imports in this way from top to bottom:
React/Third-party -> our assets and components -> stylesheet.

apps/site/src/lib/components/Footer/Footer.module.scss
```diff
@@ -1,6 +1,47 @@
 .footer {
+	background-color: var(--color-blue);
```
IanWearsHat: Issue: The Figma says the footer is a little transparent.

image

IanWearsHat (changes requested):
Thanks for getting started on this Collin! I left some comments about style and additional mobile behavior.

PR #734, opened 2026-01-14
Updates to the nav bar

### LINKED ISSUE #696: Navbar
As per Figma, we need to change the font and color of the navbar. We can comment out the "Devpost", "Report Incident", "Feedback Form", and "Log In" options.

Note: The background is transparent with a color.

Image

victorialee13:
Closes #696

Added the transparent (with color) to the nav bar. Changed the font, font size, and animations of the nav bar.

https://github.com/user-attachments/assets/f06d2d5a-9ba4-4242-8890-7523aab7479c


apps/site/src/lib/components/HackLogo/HackLogo.tsx
```diff
@@ -2,7 +2,7 @@ import Image from "next/image";
 
 // import hackAnteater from "@/assets/logos/hackuci-anteater.svg";
 // import hackGear from "@/assets/logos/hackuci-gear.svg";
-import irvineHacksLogo from "@/assets/logos/irvinehacks-logo.svg";
+import irvineHacksLogo from "@/assets/logos/irvinehacks-logo-yellow.svg";
```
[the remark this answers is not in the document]
IanWearsHat: On a separate note, if you open `irvinehacks-logo.svg` or `irvinehacks-logo-yellow.svg` in a text editor, you can see that all the svg really is is an embedded image. If the Figma has it like this, try to seek an svg that is made up of paths instead like the files above.

The reasoning for this is so that we can save some storage with path svgs as opposed to embedded image svgs. This also makes the image load faster.

apps/site/src/lib/components/Navbar/NavbarHelpers.tsx
```diff
@@ -12,7 +12,11 @@ const NavLinkItem = React.forwardRef<
 	return (
 		<NavMenu.Item>
 			<NavMenu.Link
-				className={clsx(styles.navMenuLink, className)}
+				className={clsx(
+					"font-display text-[#E5F200]",
```
IanWearsHat: Nitpick: Same as the utility class comment above.

apps/site/src/lib/components/Navbar/BaseNavbar.tsx
```diff
@@ -41,11 +41,11 @@ export default function BaseNavbar({ children }: PropsWithChildren) {
 
 	return (
 		<NavMenu.Root
-			className={`${hasScrolled ? "md:bg-opacity-50" : ""} ${
+			className={`${hasScrolled ? "md:bg-midnight/60" : "md:bg-midnight/40"} ${
 				hidden ? "max-md:h-[4rem] max-md:overflow-hidden" : ""
-			} transition-colors duration-0 md:duration-700 ease-out w-full z-50 flex flex-col fixed bg-black bg-opacity-0 md:flex-row md:items-center`}
+			} transition-colors duration-0 md:duration-700 ease-out w-full z-50 flex flex-col fixed bg-midnight/40 backdrop-blur md:flex-row md:items-center`}
```
IanWearsHat: Issue: I don't think the Figma has a blur applied for the navbar, but let me know what you think.

From Figma:
image

IanWearsHat (changes requested):

PR #736, opened 2026-01-15
Implemented Partners section

VishokL:


apps/site/src/app/(main)/(home)/sections/Partners/Partners.tsx
```diff
@@ -14,62 +14,36 @@ export default async function Partners() {
...
-				src={SmallStar}
-				alt=""
-				className="absolute right-24 bottom-48 hidden lg:block"
-				width={42}
-				height={42}
-			/>
-
-			<h2 className={`text-center text-4xl sm:text-[3rem] font-display mb-16`}>
+			<h2
+				className={`text-center text-4xl sm:text-[3rem] font-display mb-14 text-pink`}
```
IanWearsHat: Nitpick: If we don't have any templating here, we can just use a string literal without the curly braces.

IanWearsHat (approved):
Thanks Vishok!

Issue #566, opened 2025-01-16
Use new features from Python 3.12 throughout API code


taesungh:
Following from #388/#565, the serverless function for the API is now deployed with Python 3.12. Accordingly, there are a few new language features we could be using along with other things to update. These should ideally happen in phases rather than all at once.

General
- [ ] Update checks workflow and Dockerfile to use Python 3.12
- [ ] Check if overriding `AsyncIOMotorClient.get_io_loop` ([`mongodb_handler.py:23`](https://github.com/HackAtUCI/irvinehacks-site/blob/0e4addf1d873af142e7dc27b9ba110ed97f5cde0/apps/api/src/services/mongodb_handler.py#L23)) is still necessary

[New in Python 3.10](https://docs.python.org/3/whatsnew/3.10.html)
- [ ] Use new type union operator (e.g. `X | Y`) instead of `typing.Union`
- ~~Import `TypeAlias` from `typing` rather than `typing_extensions`~~

[New in Python 3.11](https://docs.python.org/3/whatsnew/3.11.html)
- [ ] Use the native [`enum.StrEnum`](https://docs.python.org/3/library/enum.html#enum.StrEnum)
  - Be careful to check there are no regressions (e.g. might need `.value` in some places or no longer need in others)
- [ ] Import `assert_never` from `typing` rather than `typing_extensions`

[New in Python 3.12](https://docs.python.org/3/whatsnew/3.12.html#pep-695-type-parameter-syntax)
- [ ] Replace `TypeAlias` with new `type` statements
- [ ] Use the new [`itertools.batched`](https://docs.python.org/3/library/itertools.html#itertools.batched) rather than our own copy in [`utils.batched`](https://github.com/HackAtUCI/irvinehacks-site/blob/0e4addf1d873af142e7dc27b9ba110ed97f5cde0/apps/api/src/utils/batched.py)

PR #778, opened 2026-02-03
Add role and status dropdown filters to admin dashboard

### LINKED ISSUE #767: [ADMIN] Add filter for Roles and Statuses for Pie Chart
On the `/admin/dashboard` route, we display a summary of every participant, including hackers, volunteers, mentors, judges.

We should add filters for roles and statuses in the pie chart. Ensure we have a filter for all roles and all statuses.

One way we can do this is create a dropdown filter for role and one for status, then send arguments to the `useApplicantSummary` hook. Clicking on an option will update the pie chart.

cyriljoby:
fixes #767

### DIFF
--- a/apps/api/src/routers/admin.py
+++ b/apps/api/src/routers/admin.py
@@ -251,7 +251,7 @@ async def applicant_summary(
     role: Optional[str] = None,
     status_filter: Optional[str] = None,
 ) -> dict[ApplicantStatus, int]:
-    """Provide summary of statuses of applicants, optionally filtered by role and status."""
+    """Provide summary of statuses of applicants"""
     # Convert string params to enums if provided
     role_enum: Optional[Role] = None
     status_enum: Optional[ApplicantStatus] = None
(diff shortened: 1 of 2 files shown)


apps/api/src/routers/admin.py
```diff
@@ -257,9 +247,37 @@ async def volunteer_applicant(
...
+async def applicant_summary(
+    role: Optional[str] = None,
+    status_filter: Optional[str] = None,
+) -> dict[ApplicantStatus, int]:
+    """Provide summary of statuses of applicants"""
+    # Convert string params to enums if provided
+    role_enum: Optional[Role] = None
+    status_enum: Optional[ApplicantStatus] = None
+
+    if role:
```
IanWearsHat: If we move to fix #566, the checks here could be shortened to `if role not in Role:`. But that would mean moving us all to 3.12 first.

IanWearsHat (approved):
Looks clean and works for me!

Issue #618, opened 2025-06-29
Create queue page


connieyyy:
- [ ] #614
- [ ] #615
- [ ] #616
- [ ] #617

<img width="743" alt="Image" src="https://github.com/user-attachments/assets/951eb189-6dca-44b4-906e-08ab4d453f7f" />

Issue #617, opened 2025-06-29
add get next batch dropdown option


connieyyy:
[ACCEPTED] -> [WAITLIST]
- Move participants who didn't come in person to check in on time to waitlist
[WAITLIST] -> [ACCEPTED] 
- Move [WAITLIST] participants to accepted

Issue #615, opened 2025-06-29
add promote button


connieyyy:
- [QUEUED] -> [CONFIRMED]

PR #799, opened 2026-02-12
Feature/queue hacker

### LINKED ISSUE #614: Add end accepted check in dropdown option
[CONFIRMED] participants who didn't check in (in person) status are moved to [WAITLIST]

connieyyy:
### Description
- Fixed check in modal directions
- Updated waitlist transfer and added queuing routes
- Added checkin-leads page

### Testing
- Updated statuses through local dashboard and was able to see accurate changes inside database
- Will test more next week in person with check in subcommittee

### Issues
Closes #614 #615 #616 #617 #618

### DIFF
--- a/apps/api/src/routers/checkin_leads.py
+++ b/apps/api/src/routers/checkin_leads.py
@@ -82,10 +82,8 @@ async def queue_participants() -> None:
     num_spots = HACKER_WAITLIST_MAX - len(await participant_manager.get_participants())
     if len(settings["users_queue"]) == 0:
         raise HTTPException(status_code=400, detail="QUEUE EMPTY")
-    
     if num_spots <= 0:
         raise HTTPException(status_code=400, detail="VENUE FULL")
-    
     num_queued = min(len(settings["users_queue"]), num_spots)
     uids_to_promote = settings["users_queue"][:num_queued]
 


apps/api/src/routers/checkin-leads.py
```diff
@@ -0,0 +1,156 @@
...
+    Template,
+)
+from utils.email_handler import IH_SENDER, recover_email_from_uid
+from utils.batched import batched
+
+log = getLogger(__name__)
+
+router = APIRouter()
+
+HACKER_COUNT = 400
```
IanWearsHat: Suggestion: We can rename this to `HACKER_WAITLIST_MAX` or something similar. I was a little confused initially on why we were using the number of hackers.

apps/api/src/routers/checkin-leads.py
```diff
@@ -0,0 +1,156 @@
...
+        log.error("Queue settings or users_queue field is missing.")
+        return
+    uids_to_promote = settings["users_queue"][:num_queued]
+
+    await mongodb_handler.update_one(
+        Collection.SETTINGS, {}, {"$pull": {"users_queue": {"$in": uids_to_promote}}}
+    )
+
+    records = await mongodb_handler.retrieve(
+        Collection.USERS, {"_id": {"$in": uids_to_promote}}, ["_id", "first_name"]
```
IanWearsHat: Issue: We should validate the return of this `retrieve` in case some user doesn't have a `first_name`, although it's unlikely. See [here ](https://github.com/HackAtUCI/irvinehacks-site/blob/main/apps/api/src/routers/director.py#L194)for an example.

apps/api/src/routers/checkin-leads.py
```diff
@@ -0,0 +1,156 @@
...
+                first_name=record["first_name"],
+            )
+        )
+
+    log.info(f"Sending emails to {len(records)} hackers that we are at max capacity.")
+
+    if len(records) > 0:
+        await sendgrid_handler.send_email(
+            # TODO: Fix email type.
+            Template.WAITLIST_TRANSFER_EMAIL,
```
IanWearsHat: Note just to update this variable.

When I added the emails in #800, I used the name `WAITLIST_CLOSED_EMAIL` from [here](https://docs.google.com/document/d/1fRcnjMMmmVmoSJ1APrMWBjJAd5QtcFft4Z2GXVMzRzA/edit?tab=t.673om1magaxk).

apps/site/src/app/admin/checkin-leads/components/useCheckInAction.tsx
```diff
@@ -0,0 +1,49 @@
...
+		if (!selectedAction) return;
+
+		setLoading(true);
+		setMessage(null);
+
+		try {
+			let endpoint = "";
+			const actionValue = selectedAction.value as string;
+
+			if (actionValue === "end-accepted-checkin") {
```
IanWearsHat: Suggestion: instead of these if checks, we can use a js object like:
```typescript
const optionEndpoints: Record<string, string> = {
    "end-accepted-checkin": "/api/admin/checkin-leads/queue-removal",
    "get-next-batch": "/api/admin/checkin-leads/queue-participants",
    "notify-venue-full": "/api/admin/checkin-leads/close-walkins"
}

// and later:
const endpoint = optionEndpoints[selectedAction.value as string];
const method = ...
```

apps/site/src/app/admin/checkin-leads/components/useCheckInAction.tsx
```diff
@@ -0,0 +1,49 @@
...
+				endpoint = "/api/admin/checkin-leads/queue-removal";
+			} else if (actionValue === "get-next-batch") {
+				endpoint = "/api/admin/checkin-leads/queue-participants";
+			} else if (actionValue === "notify-venue-full") {
+				endpoint = "/api/admin/checkin-leads/close-walkins";
+			}
+
+			if (endpoint) {
```
IanWearsHat: Question: Why do we check for `endpoint` here? What's supposed to happen if there's no endpoint?

apps/site/src/app/admin/checkin-leads/components/StatusUpdateContainer.tsx
```diff
@@ -0,0 +1,46 @@
...
+	return (
+		<Container header={<Box variant="h2">Update Statuses</Box>}>
+			<SpaceBetween size="l">
+				<ActionSelector
+					selectedAction={selectedAction}
+					onActionChange={onActionChange}
+					onUpdate={onUpdate}
+					loading={loading}
+				/>
+				{message && (
```
IanWearsHat: (ignore if no time)

We can consider using [`NotificationContext`](https://github.com/HackAtUCI/irvinehacks-site/blob/main/apps/site/src/lib/admin/NotificationContext.ts) for notifications like this, but if you have a reason for using `Alert` or don't think it's worth the time right now, you can ignore this.

apps/api/src/routers/checkin-leads.py
connieyyy: Updated!

apps/api/src/routers/checkin-leads.py
```diff
@@ -0,0 +1,156 @@
...
+@router.post(
+    "/queue-participants",
+    dependencies=[Depends(require_role({Role.DIRECTOR, Role.CHECKIN_LEAD}))],
+)
+async def queue_participants() -> None:
+    """Remove QUEUED participants from queue and send them email notification."""
+    records = []
+    num_queued = HACKER_COUNT - len(await participant_manager.get_participants())
+    settings = await mongodb_handler.retrieve_one(
+        Collection.SETTINGS, {}, ["users_queue"]
```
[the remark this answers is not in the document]
connieyyy: Updated in revision

apps/api/src/routers/checkin-leads.py
```diff
@@ -0,0 +1,156 @@
...
+    )
+    if not settings or "users_queue" not in settings:
+        log.error("Queue settings or users_queue field is missing.")
+        return
+    uids_to_promote = settings["users_queue"][:num_queued]
+
+    await mongodb_handler.update_one(
+        Collection.SETTINGS, {}, {"$pull": {"users_queue": {"$in": uids_to_promote}}}
```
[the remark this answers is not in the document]
connieyyy: Updated in revision

apps/api/src/routers/checkin-leads.py
connieyyy: Updated in revision

apps/site/src/app/admin/checkin-leads/components/useCheckInAction.tsx
connieyyy: Not needed and removed in revision

apps/api/src/routers/checkin-leads.py
connieyyy: Updated in revision

PR #832, opened 2026-02-20
Applicant Table added

### LINKED ISSUE #630: [Admin] Add more stats
Corporate sometimes needs stats about hackathons for their outreach.

We should consider more flexible and granular stats to a [table](https://cloudscape.design/components/table/?tabId=playground). Include a filter by Role (Hacker, Mentor, Volunteer), a filter by Status, then another filter for:
- school
- major
- year

For example, maybe we first set role to Hacker, then status to "Accepted", then filter by school:
| Answer | Number of responses|
|---|---|
| UCI | 345 |
| UCLA | 50 |

**Ensure we have a filter for all roles and all statuses.**

See #778 for example on adding filters. Similar to that, one way we can add filters is create a dropdown filter for role and one for status, then send arguments to the `useApplicantSummary` hook. Clicking on an option will update the table.

victorialee13:
Created a flexible granular table that shows # results filtered through by role, status, (school/major/year). Auto updates upon changing the filters of the table. 

image

btw I tried modifying useApplicantSummary so that I can use the same endpoint but since it doesn't support the required structure as it returns the count after the statuses are filtered and doesn't allow grouping by school/major/year i added a separate endpoint /summary/applicants/table and the useApplicantTable hook for the table.

Also w the school/major/year filter, defaulted to school and user can click between the three to change thru.
Closes #630

### DIFF
--- a/apps/site/src/app/admin/dashboard/components/ApplicantTable.tsx
+++ b/apps/site/src/app/admin/dashboard/components/ApplicantTable.tsx
@@ -2,6 +2,7 @@
 
 import { useMemo, useState } from "react";
 import Box from "@cloudscape-design/components/box";
+import Button from "@cloudscape-design/components/button";
 import Container from "@cloudscape-design/components/container";
 import Select, { SelectProps } from "@cloudscape-design/components/select";
 import SpaceBetween from "@cloudscape-design/components/space-between";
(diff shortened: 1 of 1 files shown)


NoNathan17 (approved):
lgtm

PR #849, opened 2026-03-04
Feature: Applicant Navigation Buttons

### LINKED ISSUE #668: [Admin][QoL] "Next applicant" button
Adding a button to go to the next applicant after submitting would be nice so we don't have to go back and see where we left off.

collinchenn:
Implemented applicant navigation buttons (ApplicantNavigationButtons.tsx), which include:

1 ) Previous button
2) Next button
3) [Needs Reviewing | Reviewed] button

These buttons are included alongside the submit button in each applicant page, both at the top and the bottom, allowing reviewers to navigate and review applications more efficiently

There are two "pools" of applicants defined by the navigation buttons: (1) needs reviewing and (2) already reviewed. Any application returned by `useHackerApplicants()` with less than two reviewers is thrown into pool 1. Any application with two reviewers is thrown into pool 2. 

When a reviewer opens an application that belongs in pool 1, then the previous and next buttons will only cycle through applicants in pool 1. The same logic applies for pool 2.

A reviewer can adjust which pool they're in with the [Needs Reviewing | Reviewed] button, which is automatically disabled when one of the pools are empty. 

Ex:
Pool 1 (needs reviewing): app_3 <-> app_6 <-> app_1 <-> app_5 <-> app_7
Pool 2 (reviewed): app_10 <-> app_12 <-> app_11

Click on app_1: currently pointing at app_1, previous button refers to app_6, and next button refers to app_5
Click on reviewed button: currently pointing at app_10, previous button refers to null, and next button refers to app_12

Closes #668

### DIFF
--- a/apps/site/src/app/admin/applicants/components/ApplicantNavigationButtons.tsx
+++ b/apps/site/src/app/admin/applicants/components/ApplicantNavigationButtons.tsx
@@ -3,7 +3,6 @@
 import { useRouter } from "next/navigation";
 import Button from "@cloudscape-design/components/button";
 import SpaceBetween from "@cloudscape-design/components/space-between";
-import Popover from "@cloudscape-design/components/popover";
 
 import useHackerApplicants from "@/lib/admin/useHackerApplicants";
 


NoNathan17 (approved):
thanks for this collin! i just refactored a bit of stuff but overall good work

PR #871, opened 2026-04-10
Check in Modal Confirmation and Warning

### LINKED ISSUE #868: Check-In Modal Confirmation
### Summary and Context

Currently, after clicking on the modal you can exit without checking in the hacker. This can lead to confusion so we want there to be a confirmation after checking in the user but also an "You haven't checked in this user yet." if you exit the modal or try to click off.

victorialee13:
Closes #868 

I know that the issue only targets the warning message for when a user tries to close out without checking out, but I thought doubling down and also giving the successful check in message was helpful in letting users know 100% if they checked in another hacker or not. lmk about this

Check in modal includes 1) check in successful confirmation alert and 2) exit warning.
1) checkInConfirmed boolean state was added so instead of immediately closing the modal on success, triggers success alert with the applicant's name
2) showExitWarning state was added to CheckInModal w/ onDismiss and Cancel handlers. User tries to close modal without completing the check in --> warning alert "You haven't checked in this user yet" + "Exit Anyway" button

Vid of check in flow

https://github.com/user-attachments/assets/85e47e9a-4885-4075-a493-15ca92120e74

note: should i change 'user' to participant or is it tooo specific? also the word is too long. not sure lmk

### DIFF
--- a/apps/site/src/app/admin/participants/Participants.tsx
+++ b/apps/site/src/app/admin/participants/Participants.tsx
@@ -19,6 +19,7 @@ function Participants() {
 	} = useParticipants();
 	const [checkinParticipant, setCheckinParticipant] =
 		useState<Participant | null>(null);
+	const [checkInConfirmed, setCheckInConfirmed] = useState(false);
 
 	const { setNotifications } = useContext(NotificationContext);
 
@@ -65,7 +66,7 @@ function Participants() {
 			} else if (type === "waitlisted") {
 				await queueParticipant(participant);
 			}
-			setCheckinParticipant(null);
+			setCheckInConfirmed(true);
 			if (setNotifications) {
 				const message =
 					type === "accepted"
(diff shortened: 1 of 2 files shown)


NoNathan17 (changes requested):
Great work Victoria! This will definitely help a lot in the upcoming hackathons. A couple things to note though, I feel it would be a bit more intuitive to ALSO have the x button in the top right of the modal be able to exit the user from the modal when the user has not been checked in yet. 

Also, there already is a toast in the top right when a user has checked in, however I do believe your add makes it more prominent to the person on check-in that the user has been checked in, so don't worry about changing that, just something to note. 

I would also change user to participant just for consistency with the rest of the modal.

victorialee13:
@NoNathan17 Did you mean that if we press X on the top right of the modal, then press it again for confirmation, a user will be able to exit the modal? Just for clarification.

NoNathan17:
Yes, that's the user experience I was going for.

victorialee13:
@NoNathan17 is it good to go ahead and merge?

victorialee13:
Also, just for clarification to exit the modal, the user simply needs to exit twice. It doesn't matter which way they do it whether they do it twice: click outside the modal, press esc, press exit anyway, press X.

NoNathan17 (approved):

PR #874, opened 2026-04-15
Update/void application status

### LINKED ISSUE #861: Add VOID status to an Application
### Summary and Context

Directors should be able to add the status of VOID to an application that is no longer under consideration. This is because we want to perserve the data, but the applicant may not be attending the hackathon anymore.

RichieRish05:
New Feature Added:
   - Add new "VOIDED" status to decision enums
   - Added "/void-applicant" director-only API endpoints with proper auth and validation to change the status to "VOIDED", and included tests
   - Updated review endpoints to block review submissions on "VOIDED" applicants
   - Updated applicant portal to show "VOIDED" status on applicant side
   - Showed "VOIDED" status in the application review portal
   - Added director only button to void applicants in the review portal
   
Screenshot 2026-04-29 at 6 33 02 PM

Screenshot 2026-04-29 at 6 34 37 PM

Screenshot 2026-04-29 at 6 33 41 PM

Closes #861


NoNathan17 (commented):
Hi Rishi! Thanks for your work on this issue. A couple things that I should have clarified earlier. 

We only want to add a VOID status to a participant, instead of adding it to their actual user data. Additionally, we want there to simply be a button on their application to void their application that ONLY directors can see. If you have any clarifying questions, please let me know. Also, I would remove that claude file from tracking 😭

NoNathan17 (dismissed):

connieyyy (changes requested):
Hi Rishi! Thank you for implementing the new changes! 

Please remove **voided applicants** from Hacker Applications page for **organizer view**. This is to prevent organizers from reviewing the application because there is no logic to notify them that the applicant is voided right now.

connieyyy (approved):

PR #887, opened 2026-04-30
Update Directors page to manage org roles

### LINKED ISSUE #885: feat: update directors page to manage org roles
- [ ] Allow director to edit organizers' roles because every hackathon's leads will vary
- [ ] Allow director to delete organizer from all perms
- [ ] Create option to assign check in leads role in Directors > Organizers

connieyyy:
- Implemented feature for director to add and remove specific admin roles for organizer
- Added delete option to remove user document entirely
- Added check in lead option to add for organizer

Screenshot 2026-04-30 at 2 35 08 PM

Closes #885

### DIFF
--- a/apps/api/src/routers/director.py
+++ b/apps/api/src/routers/director.py
@@ -153,6 +153,40 @@ async def add_organizer(
     )
 
 
+@router.post("/update-organizers")
+async def update_organizer(
+    user: Annotated[User, Depends(require_director)],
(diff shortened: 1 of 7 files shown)


NoNathan17 (approved):
the PR looks good but for the delete modal id probably change it to something like "Are you sure you want to delete this user?" instead of what it says now

PR #891, opened 2026-05-04
Fix/notification missing fields

### LINKED ISSUE #804: Add notification for missing required field
When submitting review application, if there is a missing field (e.g. Weak/Medium/Strong), there is no notification for what field is missing. Users may think that the review was submitted successfully.

We can probably add a notification/indicator that application was not submitted successfully because of unanswered question.

victorialee13:
When reviewing applications, if a reviewer tries to submit without filling in all required fields (previous experience, and the three FRQ scores), an error notification banner appears and auto-dismisses after 3 seconds. The success notification on a successful review submission also auto-dismisses after 3 seconds. 

Closes #804


NoNathan17 (approved):
W PR

PR #892, opened 2026-05-06
Max hacker check before RSVP

### LINKED ISSUE #833: Rsvp endpoint should have max hacker check
This year, we only allowed RSVPs up until a max amount. 

Thankfully, the SWR check prevented people from RSVPing by asking the backend for if the waitlist was open, but for waitlisted users, the RSVP endpoint itself doesn't check if the waitlist is open before allowing RSVP. 

Obviously, we can still see if we have max hackers by looking at the db and still audit and kick people off if they RSVPd after, but next time we should definitely check in the backend.

RichieRish05:
Used pre built _get_waitlist_status() func to check if the waitlist is open, enforcing a MAX_HACKER_CAP of 400. 

Closes #833


NoNathan17 (commented):
lgtm

PR #903, opened 2026-05-13
Order by Filter

### LINKED ISSUE #888: Add sort filter IH Hack Applications
Right now it's sorted by recently added, sort filter for alphabetical, oldest, newest, etc

victorialee13:
Added a Order by Filter on the IH Hacker Applications page (scoped to only that but either in this pr or another I can apply the same logic to the mentor, volunteer pages)

User can choose to sort Alphabetical (A-Z, Z-A) or by Registration (Latest, Oldest)
Defaulted to the "Latest Registered" Value

image
image
image

Mobile view also looks fine, but as a future issue I think we can update/reorganize the buttons for the Show top 400 and update thresholds

Closes #888


NoNathan17 (approved):
Overall, great work! I just made a quick change to the label to make it more understandable what the sort is doing. In the future, we can extend this functionality for my sorting.

PR #911, opened 2026-05-21
feat: auto-accept/reject applicants based on age and graduation status

### LINKED ISSUE #860: Autoaccept + Autoreject Feature
### Summary and Context

Instead of having to manually score people to accept them (such as ZotHacks mentors). We want there to be a feature where people who fit a certain criteria are either auto-accepted or rejected, such as if they are below 18, graduated, etc.

VishokL:
**Summary**
Adds hardcoded auto-accept/reject rules that run at admin read time (before score-based or review-based decisions). Auto-decided applicants show a badge on the admin applicant list pages so reviewers can skip them without manual scoring.

**Auto-decision Rules**
is_18_older === false (Applies to All Roles (Hacker, Mentor, Volunteer)) -- Reject
Graduated (education_level === "graduate") (Applies to Hackers) -- Reject

**Adding More Auto-decision Rules**

1. In `apps/api/src/admin/applicant_review_processor.py`, add a reason constant near the top (e.g. AUTO_REASON_MY_RULE = "MY_RULE")
2. Add the rule inside `_compute_auto_decision()`
3. Add a reason label in `apps/site/src/app/admin/applicants/components/AutoDecisionBadge.tsx`
4. Add a test in `apps/api/tests/test_applicant_review_processor.py`

**Other Notes**
Does not affect score normalization as long as reviewers do not review applications that have already been auto-decided.

LMK if there's any other rules that should be added.


apps/api/src/routers/user.py
connieyyy, line 377: I think this block of code is for the auto save drafts feature PR. We should keep it separate from this PR.

connieyyy (changes requested):
Hi Vishok! Thank you for the auto rejection features! Please also add auto accept feature (director perms only). I'm picturing a scenario where the director already has a list of previous ZH mentors. Director can go to portal can search up previous ZH mentors' names in the hacker applications page, and there's a button to auto accept them without affecting the score normalization process as well. Thanks!

apps/api/src/routers/user.py
VishokL: Honestly don't know how this got committed twice. Removed this and added the director-only auto-accept button to the hacker application detail page. Similarly to the auto-rules, it does not affect score normalization as long as reviewers do not review applications that have already been auto-decided. Could add stricter logic to prevent this if needed, would just require some rewriting of the existing logic surrounding `add_uids_to_exclude_from_hacker_normalization`.

connieyyy (changes requested):
Hello Vishok! Thank you for implementing the auto accept feature. 

- Can you disable review if applicant is auto accepted / rejected? I believe you should pull before implementing this feature because Nathan added a queue feature.
- I do not think the auto accept logic updates the admin summary counts in `HackerApplicantsList.tsx` because the summary is still computed from the score threshold. This makes the behavior inconsistent and can make auto-accepted applicants appear as not counted in the accepted totals.

connieyyy (dismissed):
lgtm, thanks!

NoNathan17 (approved):

